PR #27, opened 2020-04-06
Feature: Translations

PeteCoward:
This is a branch that has basic interface translation in place.

# Instructions
1. `brew install gettext` if you do not have get text tools ( if you have ever makemessages before in a django project you will have )
2. `yarn` to install new packages
3. Go add a call to `getText` in a riot script tag, or js file
4. Go add the attribute `translate` to an html element
5. `yarn translate`
6. go edit the `locale/tt-tl.po` file, and `yarn translate` again
7.  Run asteroid and go to settings and change language and test

# Explanation

1. Adds 3 npm packages
  - [gettext-extractor](https://www.npmjs.com/package/gettext-extractor) which scans the javascript and html looking for `getText` calls and elements with a `translate` attribute and creates `/locales/canoe.pot` - see `messages_scan.js` for calling point
  - [gettext-parser](https://www.npmjs.com/package/gettext-parser) which takes po files and makes json from them - see `messages_to_json.js` for calling point
  - [node-gettext](https://www.npmjs.com/package/node-gettext) which reads in the json and makes the calls to getText work
2. Adds a package.json npm script which 
  1. calls `messages_scan.js`
  2. uses gettext `msgmerge` to merge any newly scanned translations into the tetum po file
  3 calls `messages_to_json.js`
3. Adds `modules/riot-translation-plugin.js` which on riot update applies translations in html
4. Adds a translation module, making `getText` available in all imported modules through `webpack.ProvidePlugin`
5. Adds a Redux changeLanguage reducer that can be called from anywhere, and updates the App root on language change
6. Makes some test translations in the `Settings.riot.html` tag

# Questions and TODOS

1. In theory this pattern supports plurals an things, not tested, and not implemented, especially in the html elements
2. performance impact of the html translation pattern ( riot-translation-plugin )
3. should probably use `gettext` not `getText` :)
4. Does it work on anyone's machine but mine?
5. What happens if the html element with `translate` attribute has element content as well as text?
6. Should we split the commands up a little to make more sense?
7. Can this pattern be easily expanded to hold different languages?


PeteCoward:
merging on verbal approval from Kate ;)

PR #32, opened 2020-04-08
Enable ngettext and message format args

PeteCoward:
This removes node-gettext and gettext-parser in favour of gettext.js in order to support message format arguments.

Messages with arguments or plurality can be defined in the riot javascript like this

```javascript
return ngettext("%2 / %1 lesson complete",// this is the singular form
  "%2 / %1 lessons complete",             // this is the plural form
  course.data.lessons_count,              // the plurality of this number decides which form is used
  course.data.lessons_count,              // first subsequent argument replaces %1
  countComplete(course.data.slug),        // other arguments replace %2+ etc...    
);
```

Currently this functionality is not available in html elements, but it would not be too difficult to add support in the RiotTranslationPlugin for something like
```html
<p translate
  translate-plural="%2 / %1 lessons complete"
  translate-arg-1={ course.data.lessons_count }
  translate-arg-2={ countComplete(course.data.slug) }>
    %2 / %1 lesson complete
</p>
```

an example of a simple non plural string with a format is
```javascript
// message on completion of a lesson %1 is replaced by the lesson title
return gettext(
  "You just completed the lesson, \"%1\"", 
  props.title
 );
```

Note that the comment on the previous line is made available to the po file and will help translators understand context through our new Transifex intergration ( I have invited you both to the team )

Screenshot 2020-04-08 at 13 52 01

Transifex integration also just created [this](https://github.com/catalpainternational/canoe/pull/33/files)  👯


src/riot/Screens/Settings.riot.html
```diff
@@ -23,6 +23,7 @@ <h4 translate>Language</h4>
                 class="btn-secondary"
                 onclick="{promptUserToInstallApp}"
                 translate
+                translation-comment="This button adds an icon to the device homescreen"
```
katevoss, line 26: What does `translation-comment` do?

PeteCoward: It adds that line to the output .po file that comes through to transifex as developer notes

Screenshot 2020-04-08 at 13 52 01

PR #38, opened 2020-04-08
Every Four Hours, App Polls for Service-Worker updates.

### LINKED ISSUE #35: Serviceworker: periodic ping
We need to make sure that users out there who have asteroid serviceworker actually get a new serviceworker without us having to contact every user and tell them refresh your browser please.

A simple periodic ping setup in the serviceworker to re-call register might be appropriate ( period to be discussed, but not too frequent )

sandroroux:
Closes #35.

### DIFF
--- a/src/riot/Components/ServiceWorker.riot.html
+++ b/src/riot/Components/ServiceWorker.riot.html
@@ -2,17 +2,23 @@
     <script>
         import { ON_ADD_TO_HOME_SCREEN } from "js/Events";
 
+        const EVERY_FOUR_HOURS = 1000 * 60 * 60 * 4;
+
         export default {
+            state: {
+                swUpdatePoller: null,
+            },
+
             onMounted() {
                 if (!("serviceWorker" in navigator)) {
                     return;
                 }
 
-                navigator.serviceWorker.register("sw.js").catch(err => {
+                navigator.serviceWorker.register("sw.js").catch((err) => {
                     console.error(`Can't register Service Worker: ${err}`);
                 });
 
-                window.addEventListener("beforeinstallprompt", async e => {
+                window.addEventListener("beforeinstallprompt", async (e) => {
                     e.preventDefault();
                     const deferredPrompt = e;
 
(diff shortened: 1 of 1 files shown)


src/riot/Components/ServiceWorker.riot.html
```diff
@@ -2,17 +2,23 @@
     <script>
         import { ON_ADD_TO_HOME_SCREEN } from "js/Events";
 
+        const EVERY_FOUR_HOURS = 1000 * 60 * 60 * 4;
+
         export default {
+            state: {
+                swUpdatePoller: null,
```
PeteCoward, line 9: maybe as there is no need for the DOM to react to this variable it could be moved up outside of the export default.

src/riot/Components/ServiceWorker.riot.html
```diff
@@ -2,17 +2,23 @@
     <script>
         import { ON_ADD_TO_HOME_SCREEN } from "js/Events";
 
+        const EVERY_FOUR_HOURS = 1000 * 60 * 60 * 4;
```
PeteCoward, line 5:
```suggestion
        const SERVICE_WORKER_POLL_INTERVAL = 1000 * 60 * 60 * 4; // every four hours
```
otherwise we have to change the variable name, if we change this :)

src/riot/Components/ServiceWorker.riot.html
```diff
@@ -29,7 +35,32 @@
...
+            },
+
+            onUnmounted() {
+                this.clearIntervals();
+            },
+
+            pollForServiceWorkerUpdates() {
+                this.state.swUpdatePoller = window.setInterval(
+                    this.updateServiceWorker,
+                    EVERY_FOUR_HOURS
```
PeteCoward, line 49:
```suggestion
                    SERVICE_WORKER_POLL_INTERVAL
```

src/riot/Components/ServiceWorker.riot.html
sandroroux: Good idea. I named it `SW_UPDATE_INTERVAL`.

sandroroux:
Hey, @PeteCoward. Thanks for the review. For whatever reason, when I resolved your feedback, your approval disappeared. My mistake. Can you review this again?

PeteCoward:
Ahh, that was my settings on the repo, dismiss stale reviews.

PeteCoward:
We have a couple of things like this, Riot tags which have no actual interface.
I wonder if they are better off as pure JS modules to simplify the riot codebase a bit.
I guess the benefit the riot gives us is controlling when it is mounted maybe

PR #45, opened 2020-04-10
refactor serviceworker management from riot to js module

PeteCoward:
A little play...

I started trying to remove Events, and got sidetracked into serviceworker, and a little distracted by webpack dynamic loading.

The main canoe.js file is now just the serviceworker and the store and babel. On a prod build this is just 114 kb. It registers the sw ( I used workbox window but am not entirely convinced by it ), then dispatches redux events. Only when the sw is installed does it dynamically load the main app bundle and launch the riot.

Not sure this gives us much benefit, but I had fun clearing service workers and watching them refresh.

Still draft as I think I have cut some notifications subscription funcitonality ( commented out to remind me to put it back ) and I'm not entirely convinced it works ALL the time.

Now adds loading display and service worker unsupported message, deliberately crap pending project side input. I think Daniel designed a loading page though


src/index.html
```diff
@@ -33,8 +33,37 @@
...
+    <div id="canoe_loading">
+        <img inline src="src/img/site_logo_white.svg">
+        <div id="service_worker_loading">
+            Favor hein...<br />
+            Please wait...
+        </div>
+        <div id="service_worker_notsupported" hidden>
+            Your browser does not support serviceworkers
+        </div>
+    </div>
```
sandroroux, line 45: I think the loading screen should be in `App`. It's weird to have app state inside and outside the app tag.

I can understand the view that a service worker is separate from the App. I disagree: a service worker without an App is useless.

Also, HTML should use kebab case, not snake case.

src/index.js
```diff
@@ -1,22 +1,33 @@
...
+import { store } from "ReduxImpl/Store";
+import { initializeServiceWorker } from "js/ServiceWorkerManagement"
 
-import "./scss/olgeta.scss";
+let storeState = store.getState();
 
-changeLanguage('tet');
+store.subscribe( () => {
+    const newStoreState = store.getState();
+    if(newStoreState.serviceWorker !== storeState.serviceWorker) {
```
sandroroux, line 10: I'd flip the conditional and un-indent this method.

```javascript
if (newState.serviceWorker === oldState.serviceWorker) {
    return;
}
```
```

File: src/js/ServiceWorkerManagement.js [hunk 3]
```diff
@@ -0,0 +1,45 @@
...
+ *
+ */
+import { Workbox } from 'workbox-window';
+import { serviceWorkerEvent } from "ReduxImpl/Store";
+const SW_UPDATE_INTERVAL = 1000 * 10 * 60 * 4;
+import { ON_ADD_TO_HOME_SCREEN } from "js/Events";
+
+
+export async function initializeServiceWorker() {
+    if ('serviceWorker' in navigator) {
```
sandroroux, line 12:
```javascript
if (!navigator.serviceWorker) {
    return;
}
```

src/js/redux/Store.js
```diff
@@ -30,6 +31,10 @@ export const changeLanguage = language => {
     store.dispatch({ type: LANGUAGE_CHANGE, language: language });
 };
 
+export const serviceWorkerEvent = event_type => {
```
sandroroux, line 34: `dispatchServiceWorkerEvent`. Methods need verbs.

src/app.js
```diff
@@ -0,0 +1,18 @@
...
+changeLanguage('tet');
+
+riot.install(function (component) {
+    // all components will pass through here
+    installTranslationPlugin(component);
+    installReduxPlugin(component, store);
+});
+
+riot.register("app", App);
+riot.mount("app", { store });
```
sandroroux, line 18: `App` doesn't need `store` anymore (I think). I wrote access methods—`isBrowserSupported` and `isSiteDownloaded`—between Riot code and the Redux store.

We should remove the `store` parameter.

src/index.html
PeteCoward: The reason for this being here is to show something quickly, before the assets required for the app are downloaded, and if the app install fails in some way.

A service worker is useless without the app ✅ 
The question is whether we should try to run the app with AND without a serviceworker. If so we need to know this, and be a lot more careful with app code.

kebab-case fix on the way

src/js/redux/Store.js
PeteCoward: I too do not like this name. I'd like a naming review of all these methods when things settle down.

src/app.js
PeteCoward: will check

PeteCoward: well it does still use the props.store, I'll leave that refactor for another day

PR #46, opened 2020-04-11
Toasts - network and sitedownloader use redux

PeteCoward:
Starting point was to remove the OnlineOrOffline riot tag, that led to Toasts using redux, which led to SiteDownloader and ErrorChecks for failed network using redux.


src/riot/Components/Toast.riot.html
```diff
@@ -1,9 +1,17 @@
 <Toast>
     <p id="snackbar">{state.toastMessage}</p>
+    <p id="error"></p>
```
sandroroux, line 3: I'd rather the single toast be style-able than there be two toasts.

src/riot/Components/Toast.riot.html
```diff
@@ -1,9 +1,17 @@
...
+        import { setErrorShown } from "ReduxImpl/Store";
         const TOAST_TIMEOUT_MS = 2500;
+        const NETWORK_MESSAGES = {
+            "offline": () => gettext("You are offline."),
+            "online": () => gettext("You are online."),
+        };
+        const DOWNLOADER_MESSAGES = {
+            "downloading": () => gettext("Site is downloading."),
+            "downloaded": () => gettext("Site download is complete!"),
+        }
```
sandroroux, line 14: The client should specify the messages, not the `Toast`.

PeteCoward: I disagree. The client should specify the type of event, not the actual message.
It's the interface's job to decide how to communicate that to the user.
It's not a pattern I've followed uniformly but I would prefer to do it that way.

src/riot/Components/Toast.riot.html
PeteCoward: I toyed with making it red and align left rather than right but decided to not devote more time to it.
Very happy to think about a better pattern for this component.
I want to look at what happens when events come quickly though, at the moment the first timeout can hide the second message after not enough time to see it.

src/riot/Components/Toast.riot.html
sandroroux: Think about `window.alert` or how Bootstrap handles modals and toasts. They don't include canned messages. They are UI libraries, which was my intention with this component.

PR #51, opened 2020-04-14
Redux Store contains flattened courseware.

sandroroux:
This PR begins to reduce our courseware size. By flattening our state objects, we can reduce`LessonPage`, `CoursePage`, and `HomePage`'s data representation (i.e. the server's JSON representation).

Now, Redux State separates the HomePage (`home`), CoursePages (`courses`) and LessonPages (`lessons`):
image

`home`, `courses`, and `lessons` are page-ID indexed. Each flattened page refers to its parent and/or children by ID:
* `home` has a `parentId` and `courseIds`;
* `courses` have a `parentId` and `lessonIds`;
* `lessons` have a `parentId`.
You can find any parent or child page with a ID-based search of the state object.

The site still uses the Redux State's `pages`. This PR maintains it. This PR doesn't migrate anything to the flat courseware representation.


PeteCoward:
@sandroroux make it ready, and I'll review it :)

src/js/WagtailPagesAPI.js
```diff
@@ -76,3 +78,53 @@ export const getOrFetchWagtailPage = async path => {
...
+};
+
+export const getLessonById = async (lessonId) => {
+    await _getOrFetchWagtailPageById(lessonId);
+    // Until we can switch to a flatter data representation, this ensures the
+    // store has the lesson.
+
+    const lesson = getLesson(lessonId);
+    if (!lesson) {
+        throw new Error(`Lesson ${lessonId} doesn't exist.`);
```
PeteCoward, line 107: @sandroroux can we talk about this? I worry that we could easily end up throwing this error due to CMS content changes, and not sure what would result as a user experience

PR #52, opened 2020-04-15
Rogue getText and some unsubscribe fixes

PeteCoward:
- replaces a leftover getText with gettext
- adds the SITE_NAME to the DefinePlugin dict so that we can...
- replace 'Olgeta' with a placeholder SITE_NAME in unsubscribe messages
- renames some confusing method duplications, I'm not sure if they were causing errors, but they were sure causing me headaches ( Settings had a subscribe tag method and a imported subscribe method )

### DIFF
--- a/locale/canoe.pot
+++ b/locale/canoe.pot
@@ -5,22 +5,28 @@ msgstr ""
 "Report-Msgid-Bugs-To: engineering@catalpa.io\n"
 "Language: en\n"
 
+#: src/riot/WagtailPages/HomePage.riot.html:106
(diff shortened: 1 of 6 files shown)


sandroroux:
@PeteCoward The doubled method names are a Canoe pattern. Since tag HTML cannot use imported methods, we create tag-level wrapper methods. You'll see the same thing in `App`:
```javascript
isBrowserSupported() {
    return isBrowserSupported();
},

isSiteDownloaded() {
    return isSiteDownloaded();
},

isUserLoggedIn() {
    return isUserLoggedIn();
},
```

PR #53, opened 2020-04-15
Img compression manual

PeteCoward:
Svg compression by manually passing images through the https://vecta.io/nano tool recommended by @sergioestevess .

59kb production gain
canoe.js    565 KiB - pre PR
canoe.js    506 KiB - post PR

going to investigate using https://github.com/tcoopman/image-webpack-loader in a separate PR to compare gains


sergioestevess:
Have you tried reducing the precision to "1"? Usually "0" breaks the image...

image

sergioestevess:
It might save us couple more bytes.

PeteCoward:
I was using precision 3

boerwastaken:
There's also https://github.com/scour-project/scour, which I have used, and which is an actual computer program (and python library) that runs on your computer instead of a website you can't automate or integrate into your build process.

Might be worth doing a little comparison.

katevoss:
I looked through the images and everything looks good to me! Is there anything in specific I should look for @PeteCoward ?

PeteCoward:
Well neither is for merge yet, we need to evaluate whether it is better to do it manually like this, image by image, keeping our inline svg capability, or toolchain it like the other PR, which takes less thought but removes inlining and risks unintended quality deprecation.
Like wicher says, we need to compare a few approaches and evaluate relative winz.

PeteCoward:
I'm re-opening this as it's a non-harmful ( unless someone with better eyes can spot a quality issue ) reduction in size.
The alternative https://github.com/catalpainternational/canoe/pull/54 I don't like as it removes inlining increasing the number of requests and so potentially the chance of failure

sergioestevess:
The tool I just recommended has a [NPM package](https://www.npmjs.com/package/nanosvg) too. But it requires a license...

boerwastaken:
If it's about convenience, scour is in Ubuntu (python3-scour) and you can `pip install` it anywhere.

PR #54, opened 2020-04-15
Img optimize webpack loader

PeteCoward:
Using https://github.com/tcoopman/image-webpack-loader to optimize images in the build process

https://github.com/tcoopman/image-webpack-loader

This does not inline svgs like we had before so the number of files spat out by the build process increases, but the total size of files is a 165 KiB win.

canoe.js     564 KiB -> 287 KiB
dist/            981 KiB  -> 816 KB

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
         "gettext-extractor": "^3.5.2",
         "html-loader": "^0.5.5",
         "html-webpack-plugin": "^3.2.0",
+        "image-webpack-loader": "^6.0.0",
         "node-sass": "^4.13.1",
         "optimize-css-assets-webpack-plugin": "^5.0.3",
         "sass-loader": "^8.0.2",
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -91,18 +91,17 @@ module.exports = (env) => {
                     },
                 },
                 {
-                    test: /\.(png|jpg|gif)$/,
-                    use: ["file-loader"],
-                },
-                {
-                    test: /\.svg$/,
-                    use: {
-                        loader: "svg-url-loader",
-                        options: {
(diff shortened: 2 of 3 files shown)


PeteCoward:
I could not work out ho to configre the precision of the svg compression :)

sergioestevess:
The available options for SVG compression: https://github.com/svg/svgo#what-it-can-do

boerwastaken:
I've noticed that yes, one can remove redundancy and increase entropy in a preprocessing step, but then since the HTTP transport compression has more entropy to deal with, the gains are much less on the compressed size-on-wire than on the uncompressed size-on-disk. Of course a dedicated SVG minifier can do smarter optimizations, and a bunch of lossy transformations to boot, but `gzip -9` is pretty good as well (and you can also think of configuring Brotli compression, as most web browsers now support that too).

In short, if you're trying to minimize transfer size, don't compare uncompressed sizes, but `gzip -9` the artefacts, then compare, and weigh the (much smaller) savings against the baggage of more dependencies and more complicated build chains. Perhaps also do a brotli compression run and then we might want to make a case to enable brotli transport compression on our web servers.

Storage size (in SW cache) is another story of course, as I suspect the responses are stored uncompressed.

PeteCoward:
deployed here https://asteroid-dev.temp.build/

> The available options for SVG compression: https://github.com/svg/svgo#what-it-can-do

yeah, I tried but could not find the right options sauce to see any difference

katevoss:
How do you recommend testing and reviewing this PR @PeteCoward  ?

PeteCoward:
Closing this as I'm not happy that it removes inlining. If we can find a way to optimize and inline that would be best.

PR #61, opened 2020-04-20
Test question with multiple answers

katevoss:
Eventually this will close https://github.com/catalpainternational/olgeta/issues/222

The purpose of this draft PR is to explain the logic and design choices while I finish up the remaining to-do's for this issue. There are undoubtably things I've missed which is why I'm sharing now, so you can help me catch them 😄 

### `<LessonTest>` 
* includes three new tags: `TestSingleAnswer`, `TestMultipleAnswer`, `TestFeedbackModal`, and it includes the logic for which tag(s) to show.
* in `onMounted` it updates whether the question has multiple answers with `isSingleAnswer`. When I tried to put it into `beforeOnMounted` I got errors, but there may be a better way to deal with the errors.

### `<TestFeedbackModal>`
* the modal's new home is here, along with the translation messages.

### `<TestMultipleAnswer>`
* `hiddenAnswers` is an array of answers that the user has selected, but the user _can't_ see if the answer is correct
* `visibleAnswers` is the array of answers that the user _can_ see whether the answer is correct or incorrect
* In `showResults()` when the user clicks Submit, the answers from `hiddenAnswers` is added to the `visibleAnswers`. That way if a user selects an answer > click submit > select another answer > click submit > all selected answers are now visibly correct or incorrect.
* I styled the checkboxes from scratch because it was simpler than trying to overrule the default <input type="checkbox">` styling
* in `selectAnswer(answer)` if an answer is visible (in the `visibleAnswers` array) then it is deselected. If it is hidden, it is also deselected. Thus the very specific order of if's and else's.

## Missing things:
* ~~`TestFeedbackModal` needs support for multiple feedback messages, for when a user selects multiple answers (both correct and incorrect). This also needs to be styled. But the backend doesn't support custom messages yet.~~ 
  * **Fixed in** https://github.com/catalpainternational/canoe/pull/61/commits/680216240c555ed83c79d29b207d4a70bbe495d8 and  https://github.com/catalpainternational/canoe/pull/61/commits/fdfc6be73d3650bc97da9dbb4f20385499a69dc8
* `TestFeedbackModal` may need an additional modal to handle the case when a user is missing correct answers and has no incorrect answers. This would give them feedback that they are correct _and_ they're missing at least one answer.
*  `TestMultipleAnswer` needs support for Google Analytics via `dispatchTestAnswerEvent`. It may need some backend support first in order for GA to support logging multiple answers to a question.
* ~~Support for multiple `TestMultipleAnswer` questions in a row. Right now the props change but `TestMultipleAnswer` does not re-mount, which causes issues for the second MultipleAnswer question.~~ 
  * **Fixed in** https://github.com/catalpainternational/canoe/pull/61/commits/572251c5f2fd25d15ac166c94ffee9ba72677592
* **New** By fixing ^above^ the visible answers now clear when the feedback modal appears so they're no longer visible in the background of the modal. See image here: https://github.com/catalpainternational/canoe/pull/61#issuecomment-616885081


katevoss:
In https://github.com/catalpainternational/canoe/pull/61/commits/572251c5f2fd25d15ac166c94ffee9ba72677592 I fixed the issue where when there are two successive multiple-answer-questions, `visibleAnswers` was not clearing because `TestMultipleAnswers` is not re-mounted. Unfortunately it clears the answers from the background of the modal:
image

Any other ideas how to address this?

src/riot/Lesson/LessonTest.riot.html
```diff
@@ -2,44 +2,34 @@
...
-                        <span class="warning"></span>
-                        <p class="feedback-message text-orange-1" translate>Incorrect answer</p>
-                    </div>
-                </li>
-            </ol>
+            <TestSingleAnswer 
+                if="{state.isSingleAnswer}"
+                lessonName="{props.lesson.title}"
+                answers="{state.cards[state.cardIdx].answers}"
+                question="{state.cards[state.cardIdx].question}"
```
PeteCoward, line 9: why not just pass the whole card down?

src/riot/Lesson/TestMultipleAnswer.riot.html
```diff
@@ -0,0 +1,86 @@
...
+    >Submit</button>
+
+    <script>
+        import { dispatchTestAnswerEvent } from "js/Events";
+
+        export default {
+            state: {
+                hiddenAnswers: [],
+                visibleAnswers: [],
+                numberTotalCorrectAnswers: ''
```
PeteCoward, line 42: I think you might be better off just having one state of 'answers'. If you copy the answer array from props to state then `selectAnswer` can simply toggle a isChecked on the answer object.
I think that will make bot script and html sections much simpler to code.

katevoss: Using one state of answers, how would I differentiate between answers that are checked compared to answers that are marked as incorrect/correct?
image
The first two answers are "visible" answers, and the 3rd answer is "hidden" because the user hasn't submitted it yet.

src/riot/Lesson/LessonTest.riot.html
katevoss: Good suggestion, added in https://github.com/catalpainternational/canoe/pull/61/commits/92038b3c2f4278b965d6aff7c6378d14cd0fcfd0

src/riot/Lesson/TestMultipleAnswer.riot.html
PeteCoward: I don't like that naming. All answers are always visible! a showFeedback  attribute could be true or false?

src/riot/Lesson/TestFeedbackModal.riot.html
```diff
@@ -1,7 +1,9 @@
 <TestFeedbackModal>
     <div id="modal" class="feedback-modal">
-        <h2 if="{props.isQuestionComplete}">{ correctTitleMessage() }</h2>
-        <h2 if="{!props.isQuestionComplete}">{ incorrectTitleMessage() }</h2>
+        <div each="{answer in props.selectedAnswers}">
+            <h2 if="{props.isQuestionComplete}">{ correctTitleMessage(answer) }</h2>
+            <h2 if="{!props.isQuestionComplete}">{ incorrectTitleMessage(answer) }</h2>
```
PeteCoward, line 5:
```suggestion
            <h2>{ props.isQuestionComplete ? correctTitleMessage(answer) : incorrectTitleMessage(answer) }</h2>
```

src/riot/Lesson/TestMultipleAnswer.riot.html
katevoss: How about `showAccuracy`? Would that be an attribute that I add to the answer in state?

src/riot/Lesson/LessonTest.riot.html
```diff
@@ -2,44 +2,32 @@
...
-            <button class="btn-primary btn-wrong" onclick="{ () => toggleModal(false) }" translate>Try again</button>
-        </div>
-    </div>
+    <TestFeedbackModal
+        if="{state.showModal}"
+        isQuestionComplete="{state.isQuestionComplete}"
+        selectedAnswer="{state.selectedAnswer}"
+        isLastQuestion="{state.cardIdx === state.cards.length - 1}"
+        toggleModal="{toggleModal}"
+        course="{props.course}"
```
sandroroux, line 26: I wouldn't pass the whole course, only the information the modal requires, `courseSlug`.

src/riot/Lesson/LessonTest.riot.html
```diff
@@ -80,13 +75,23 @@ <h2>{ incorrectTitleMessage() }</h2>
...
+            },
+
             onBeforeUnmount(props, state) {
                 window.removeEventListener("hashchange", this.hashChanged);
             },
 
             hashChanged() {
+                const cardIdx = getLessonCardIdx();
+                const isSingleAnswer = this.state.cards[cardIdx] && this.state.cards[cardIdx].answers.filter(a => a.correct).length === 1;
+
```
sandroroux, line 91: This line is long. For legibility, I'd break it:
```javascript
const cardIdx = getLessonCardIdx();
const card = this.state.cards[cardIdx];
const correctAnswers = card.answers.filter(answer => answer.correct);
const isSingleAnswer = correctAnswers.length === 1;
```

src/riot/Lesson/LessonTest.riot.html
```diff
@@ -103,79 +108,21 @@ <h2>{ incorrectTitleMessage() }</h2>
...
-                    isAnswerCorrect: answer.correct
-                };
-                dispatchTestAnswerEvent(eventData);
-                this.toggleModal(true);
-            },
+            toggleModal(showModal, selectedAnswer, isQuestionComplete) {
+                this.update({
+                    selectedAnswer,
+                    isQuestionComplete 
+                });
```
sandroroux, line 115: You could use one update and be [DRYer](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)?

```javascript
toggleModal(showModal, selectedAnswer, isQuestionComplete) {
    const testElement = this.$("#test"); 
    
    this.update({
        showModal
        selectedAnswer,
        isQuestionComplete 
    });

    if (showModal) {
        testElement.classList.add("showElement");
    } else {
        testElement.classList.remove("showElement");
    }
},
```

src/riot/Lesson/LessonTest.riot.html
```diff
@@ -56,21 +44,28 @@ <h2>{ incorrectTitleMessage() }</h2>
     <script>
         import LessonComplete from "RiotTags/Lesson/LessonComplete.riot.html";
         import LessonFeedback from "RiotTags/Lesson/LessonFeedback.riot.html";
+        import TestSingleAnswer from "RiotTags/Lesson/TestSingleAnswer.riot.html";
+        import TestMultipleAnswer from "RiotTags/Lesson/TestMultipleAnswer.riot.html";
```
sandroroux, line 48: `Feedback` now has multiple meanings. We should rename `LessonFeedback` or `TestFeedbackModal`.

Either `TestFeedbackModal` can be `CheckAnswersModal`, or `LessonFeedback` can be `LessonSurvey`.

src/riot/Lesson/LessonTest.riot.html
```diff
@@ -2,44 +2,32 @@
...
+        isQuestionComplete="{state.isQuestionComplete}"
+        selectedAnswer="{state.selectedAnswer}"
+        isLastQuestion="{state.cardIdx === state.cards.length - 1}"
+        toggleModal="{toggleModal}"
+        course="{props.course}"
+        lesson="{props.lesson}"
+        module="{props.module}"
+        cardIdx="{state.cardIdx}"
+    ></TestFeedbackModal>
     <LessonFeedback
```
sandroroux, line 31: `Feedback` now has multiple meanings. We should rename `LessonFeedback` or `TestFeedbackModal`.

Either `TestFeedbackModal` can be `CheckAnswersModal`, or `LessonFeedback` can be `LessonSurvey`.

src/riot/Lesson/TestMultipleAnswer.riot.html
```diff
@@ -0,0 +1,78 @@
...
+            </li>
+        </ol>
+    </div>
+    <h4 if="{state.visibleAnswers.length > 0}" translate>Find all correct answers and uncheck incorrect answers</h4>
+    <button
+        class="btn-primary"
+        onclick="{showResults}"
+        disabled="{state.hiddenAnswers.length === 0 && state.visibleAnswers.length === 0}"
+        translate
+    >Submit</button>
```
sandroroux, line 25: `Submit` works, but maybe `Check Answers` is clearer?

src/riot/Lesson/TestMultipleAnswer.riot.html
sandroroux: @katevoss, to Peter's point, it might be easier to avoid the hidden/visible distinction. You can track whether a box is checked, and style the box according to its rightness:

(what follows is pseudocode)
```javascript
feedback = [];
submit() {
    const checkboxes = this.$$('li');
    for (const checkbox of checkboxes) {
        if (!isSelected(checkbox)) {
            continue;
        }
        else if (isWrong(checkbox)) {
            feedbackObj = {
                isCorrect: false,
                reason: checkbox.giveReason()
            };
            feedback.push(feedbackObj);
            highlightAnswerIsWrong(checkbox);
        }
       else {
            feedbackObj = {
                isCorrect: true,
                reason: checkbox.giveReason()
            };
            feedback.push(feedbackObj);
            highlightAnswerIsRight(checkbox);
       }
    }
}
```

src/riot/Lesson/LessonTest.riot.html
```diff
@@ -2,44 +2,32 @@
...
+                lessonName="{props.lesson.title}"
+                card="{state.cards[state.cardIdx]}"
+                toggleModal="{toggleModal}"
+            ></TestSingleAnswer>
+            <TestMultipleAnswer
+                if="{!state.isSingleAnswer}"
+                lessonName="{props.lesson.title}"
+                card="{state.cards[state.cardIdx]}"
+                toggleModal="{toggleModal}"
+            ></TestMultipleAnswer>
```
sandroroux, line 16: From a functionality standpoint (and with styling aside), you should be able to make a `SingleAnswer` from a `MultipleAnswer`. If you ignore `label`s, the back-end already doesn't distinguish `single` and `multi`.

I'm curious, what are your thoughts? Why split the two?

src/riot/Lesson/LessonTest.riot.html
```diff
@@ -80,13 +75,23 @@ <h2>{ incorrectTitleMessage() }</h2>
                 window.addEventListener("hashchange", this.hashChanged);
             },
 
+            onMounted(props, state) {
+                this.update({ 
+                    isSingleAnswer: this.state.cards[this.state.cardIdx] && this.state.cards[this.state.cardIdx].answers.filter(a => a.correct).length === 1 
+                });
```
sandroroux, line 81: Same as [this comment](https://github.com/catalpainternational/canoe/pull/61/files#r414116781). The line is long; I'd split it.

src/riot/Lesson/TestFeedbackModal.riot.html
```diff
@@ -0,0 +1,67 @@
...
+                    this.props.module,
+                    this.props.cardIdx
+                );
+            },
+
+            correctTitleMessage(selectedAnswer) {
+                // title for correct answer feedback
+                return selectedAnswer.feedback_title 
+                ? selectedAnswer.feedback_title
+                : gettext("Great");
```
sandroroux, line 42: Better to:
```javascript
return selectedAnswer.feedback_title || gettext("Great");
```

src/riot/Lesson/TestFeedbackModal.riot.html
```diff
@@ -0,0 +1,67 @@
...
+            },
+
+            correctTextMessage(selectedAnswer) {
+                // text for correct answer feedback
+                return selectedAnswer.feedback_text 
+                ? selectedAnswer.feedback_text 
+                : gettext("Well done, keep it up!");
```
sandroroux, line 49: Same as [this](https://github.com/catalpainternational/canoe/pull/61/files#r414148628).

src/riot/Lesson/TestFeedbackModal.riot.html
```diff
@@ -0,0 +1,67 @@
...
+            },
+
+            incorrectTitleMessage(selectedAnswer) {
+                // title for incorrect answer feedback
+                return selectedAnswer.feedback_title
+                ? selectedAnswer.feedback_title
+                : gettext("Oops");
```
sandroroux, line 56: Same as [this](https://github.com/catalpainternational/canoe/pull/61/files#r414148628).

src/riot/Lesson/TestFeedbackModal.riot.html
```diff
@@ -0,0 +1,67 @@
...
+            },
+
+            incorrectTextMessage(selectedAnswer) {
+                // text for incorrect answer feedback
+                return selectedAnswer.feedback_text 
+                ? selectedAnswer.feedback_text 
+                : gettext("Read the question more carefully!");
```
sandroroux, line 63: Same as [this](https://github.com/catalpainternational/canoe/pull/61/files#r414148628).

src/riot/Lesson/TestMultipleAnswer.riot.html
```diff
@@ -0,0 +1,78 @@
...
+                this.update({ 
+                    numberTotalCorrectAnswers, 
+                    hiddenAnswers: [] 
+                });
+            },
+
+            showResults() {
+                this.state.visibleAnswers = this.state.visibleAnswers.concat(this.state.hiddenAnswers);
+                this.state.hiddenAnswers = [];
+                this.update();
```
sandroroux, line 48: Use the update method to change `state`:
```javascript
const visibleAnswers = this.state.visibleAnswers.concat(this.state.hiddenAnswers);
const hiddenAnswers = [];
this.update({
    visibleAnswers,
    hiddenAnswers,
});
```

src/riot/Lesson/TestMultipleAnswer.riot.html
```diff
@@ -0,0 +1,78 @@
...
+
+                const numberIncorrectAnswersChosen = this.state.visibleAnswers.filter(a => !a.correct).length;
+                const numberCorrectAnswersChosen = this.state.visibleAnswers.filter(a => a.correct).length;
+                const numberCorrectAnswersMissing = this.state.numberTotalCorrectAnswers - numberCorrectAnswersChosen;
+                const isQuestionComplete = numberCorrectAnswersMissing === 0 && numberIncorrectAnswersChosen === 0
```
sandroroux, line 53: This my preference, so ignore it if you'd like:
```javascript
const isQuestionComplete = numberCorrectAnswersChosen === this.state.numberTotalCorrectAnswers;
```

src/riot/Lesson/LessonTest.riot.html
PeteCoward: With functionality coming with differences between feedback message behaviour in single and multi answer questions I think it makes sense to make a new tag. They are being split on the CMS side now to allow for validation of the number of answers marked as correct

src/riot/Lesson/LessonTest.riot.html
katevoss: The `.showElement` class darkens the background of the modal over the test question, whereas `showModal` triggers the modal to display. I guess they could be combined into the `.feedback-modal` class but it's semantically clearer to separate the modal styling and animation from deactivating the background.

src/riot/Lesson/TestMultipleAnswer.riot.html
katevoss: >Use the update method to change state:

Is there a benefit to doing it this way?

src/riot/Lesson/TestMultipleAnswer.riot.html
katevoss: We also have to check that there are no incorrect answers selected with `numberIncorrectAnswersChosen === 0`, so it could be 
```js
const isQuestionComplete = numberCorrectAnswersChosen === this.state.numberTotalCorrectAnswers && numberIncorrectAnswersChosen === 0;
```

PeteCoward:
@katevoss is it intentional that the right and wrong indicating checkbox higlights persist when you click try again?

captured (1)

katevoss:
@PeteCoward yes that is intentional that the checkbox persist. Do you think they should be unchecked when a user has to re-select answers?

src/riot/Lesson/TestMultipleAnswer.riot.html
sandroroux: It's Riot-4-suggested. Your `update` method was common in earlier versions of Riot.

I think the newer version follows a functional programming mindset: that state should not be directly modified, but replaced. This attitude motivates Redux reducers too, and React does something similar (I don't know about Vue).

In theory, if Riot ever uses a virtual DOM, the maintainers may optimize the newer way.

src/riot/Lesson/TestMultipleAnswer.riot.html
sandroroux: Yup, that makes sense.

sandroroux:
@katevoss, @PeteCoward, either way, a user can "figure out" the answers by checking all the boxes and hitting "Check answers".

PeteCoward:
Just checking. Let's get it in Rc-1.0 for review

PR #63, opened 2020-04-21
Card missing duration

katevoss:
Fixes https://github.com/catalpainternational/asteroid/issues/51

It appears that `ngettext` has a minimum number of characters that it will translate. "1 min" or "3 mins" is too short, but "1 minute" and "3 minutes" works.


PeteCoward:
It works if you actually add translations, the bug is a little more complex than just message is too short I think.
I'm closing this PR, as the fix will come in with translations populated from transifex

PR #65, opened 2020-04-21
Load in improvements

PeteCoward:
Following up @boerwastaken concerns about inability to retry on serviceworker failed install, this adds feedback for failed install, and a reload on tap ability.

Failure can be reproduced by running a canoe-build, serving the app through `python -m http.server 8080 --bind localhost` from inside your dist directory, then deleting one of the precached images.

Also updates workbox packages to 5.1.2 from 5.0.0


src/js/ServiceWorkerManagement.js
```diff
@@ -19,7 +19,9 @@ export async function initializeServiceWorker() {
     wb.addEventListener("externalactivated", () => {
         serviceWorkerEvent("externalactivated")
     });
-
+    wb.addEventListener("redundant", () => {
+        serviceWorkerEvent("install-failed")
+    });
```
RobertoMaurizzi, line 24: Assuming this isn't one of the many 'let's rename standard lifecycle events to confound the enemy' from Workbox, you should get a redundant when and old service worker dies after the new one takes control... why install failed?

PeteCoward: Good point I may have to look deeper into other cases redundant heppens

PeteCoward: I did have a more targetted approach, because we build workbox with webpack I can alias the WorkboxError class, extend it with my own and get access to all it's internal errors, but then Jeff Posnick from workbox pointed me towards redundant and made me feel rather stupid for missing that event.

RobertoMaurizzi: I've now seen 😅 that you also get a redundant when a SW fails to load due to some JS error (in my case I was asking it to precache a file I didn't publish) so for sure it works in that case.
I'm quite sure I've seen some old SW going redundant during skipwait/reload but I'm not using Workbox's skipWait because it didn't work without loading the SW from WB itself 🤪

boerwastaken:
I don't know the specifics of the canoe/workbox/serviceworker stack, I'm only looking at this patch.  Just mentioning that I would say a failed service worker install is only a problem if there's no previous service worker to fall back on. I can't see from the patch whether that case is distinguished.

boerwastaken:
Also, more general, if you ever make a mistake somewhere in this path causing the reload affordance to not come up while there is in fact real necessity for reload, non-browser clients without reload button (as the Asteroid Android app now is, because its menu bar has been removed with catalpainternational/asteroid#22)  still be dead in the water.

It's like when the power goes down in your house and you need to reset a fuse, but instead of a simple hardware mechanism for doing that to your frustration you find out that to reset the fuse you nowadays have to use some fancy touchscreen widget with colorful friendly UI, which is currently not working because there is no power.

RobertoMaurizzi:
Since I had this exact problem today 😅 as a general observation if you need a reload but the button doesn't come up and you have no interface to clear the data, you can still kill all tasks on the phone, wait a few seconds so that Chrome and friends will close properly, then reopen the app: the old service worker shouldn't be around anymore and the new one will start.

boerwastaken:
Yes, a techie can do that various things to get rid of state, but there's two problems that pertain to especially less techie crowds:

1. how to perform `$technical_seance` to resolve the issue
2. that `$technical_seance` is worth an attempt in the first place

There's value in a reload button that's always there for you. You need to be pretty furkin confident about your code and its environment, including any previous state, before taking away the user's escape hatch.
And if there's a reload button, one might as well add a nuke-all-local-state button (caches, SW's, cookies, IDB, localstorage, ...). If I'd get a support call about an app seeming to be stuck for unknown causes I'd be a very happy engineer if I could suggest tapping the nuke button to see if that helps. Or I might not even get called because the user has already figured out how to help him/herself (mash one of the two buttons presented, if it doesn't work, mash the other, hey I can log in again, hurray)

RobertoMaurizzi:
One of the things I hate most about how SW were implemented is that they broke reload/hard reload.
I do like the idea of a nuke button half hidden somewhere, I used that more than a few times (pre-web days)

src/js/ServiceWorkerManagement.js
PeteCoward: I can reproduce getting that second case redundant, and it does not actually have any efffect, the event fires, the retyr button is made visible, but is totally invisible due to the app being mounted on top :)
I still need to edit this though, as the redux state gets out of line

PeteCoward:
This raises the question of whether when the user closes and re-opens the APK, the GeckoView retains any state at all? Are the serviceworker, cache, localstorage still there when it comes back?

If not, then that is the easy support fix for a stuck app. but we've lost offline ability to close and reopen the app.
If it is retained then that is good for offline capability, but we'll open issues for improving the reset capability.

@boerwastaken can you confirm?

boerwastaken:
> This raises the question of whether when the user closes and re-opens the APK, the GeckoView retains any state at all? Are the serviceworker, cache, localstorage still there when it comes back?

An APK is a zipfile. You can close and re-open an file, and Geckoview will never know. You probably mean something like "the Android app".

With that out of the way, of course it retains state, it would be a bit useless for app purposes if you couldn't even persist a cookie or IDB stuff. The Eskola app, also based on Appelflap/Geckoview, then wouldn't be able to start up when offline.

PeteCoward:
> With that out of the way, of course it retains state,

Does it cause a refresh of the page when re-opened?

boerwastaken:
> Does it cause a refresh of the page when re-opened?

Not necessarily. It's something to go to lengths to to avoid, because a page reload every time you switch away from the app and back again sounds pretty annoying.

On a cold boot of the apparatus I think it needs to do a full page load though (and that's when the service worker is supposed to kick in for the offlineity)

boerwastaken:
I'd say give it a whirl of you want to experiment, you can grab it from the play store.

PeteCoward:
Well as long as its not a "hard" refresh, this PR still fails on that front :)

PeteCoward:
I can't decide if it's best to just bite the bullet and let the app run without a serviceworker, or keep fiddling around in serviceworker land to detect the hard reload situation reliably and deal with it.

src/index.js
```diff
@@ -12,23 +12,31 @@ store.subscribe( () => {
     }
 
     switch(newStoreState.serviceWorker) {
+        case "install-failed":
+            // show a retry message
+            document.querySelector("#service-worker-loading").hidden = true;
+            document.querySelector("#service-worker-notsupported").hidden = true;
```
sandroroux, line 18: A method should contain this. `toggleServiceWorkerNotSupported`.

sandroroux:
@PeteCoward, I get the feeling that service workers and websites should be independent. Though Canoe operates as an app, Canoe is, at core, a website and most accessible as a website.

I do think canoe's service worker should be optional. `SiteDownloader`/`DataDownload` should react to the service worker lifecycle, not the whole app.

Also, requiring service workers breaks `command+shift+R`.

PeteCoward:
No-one but us use hard refresh, not even sure how you do it on a mobile device :) 

If we do allow no serviceworker, we have to:
1. not pre-cache if there is no sw ( or manually interact with the cache bypassing the sw )
2. on first load decide between:
    1. waiting for sw failure/success to load in the site
    2. immediately load in the site, and accept that both site and sw will be downloading the same resources at the same time

It is less work to require a sw.

boerwastaken:
Not saying we SHOULD, but 

> immediately load in the site, and accept that both site and sw will be downloading the same resources at the same time

We COULD stick a mutex lock in there, on individual resources, and then both and main thread will be working together filling up that cache...

PeteCoward:
My new plan to move forward with this...

1. show please wait while trying serviceworker install
2. show "launch anyway" button if the serviceworker is not supported
  or just launch anyway
3. show "retry offline install" and "launch anyway" if serviceworker failed install
    or just launch anyway
4. Only pre-cache if sw controlling
    4b. refactor to add to cache and get from cache instead of using sw for content
5. optimizeoptimizeoptimize assets

boerwastaken:
Some thoughts:

### ad 1+2+3
- Provide visual feedback to the user for loading progress. Otherwise the "well, you can always tap 'launch anyway'" argument is a bit of a cop-out as the user has no information to decide whether that's a good approach or not. I ran into this quite early. If you're offering users actions, then you should also offer users actionable information (a looped animation is NOT it), otherwise it'll just cause anxiety, plus it leads to folkloristic phenomena such as users telling eachother "yeah always immediately tap 'launch anyway' otherwise it will become stuck" etc.

### ad 4
- You might still use `Cache.*()` to put in/retrieve from the same place as an eventual SW would. But I think a "no serviceworker, no glory" approach makes more sense than going to great lengths to emulate an SW. 

### ad 4b
If you're going to emulate an SW, I think you could load an SW as a webworker. Of course the SW hooks will not be called but you could use message passing to call into the same primitives as the SW-running-as-SW would. Precaching then also becomes an option again. Feasibility of "if no SW registration, then load the SW as a WebWorker" depends on whether your existing SW lends itself to such caprioles.


## But
But all in all I would say 

1. just don't launch without a serviceworker
2. don't boot into the app without a complete precache of the app necessities
3. separate the SW lifecycle from the data lifecycle
4. co-version app versions and data versions, so that your app doesn't get data it can't digest.

Basically build some strong guarantees first.

src/index.html
```diff
@@ -40,7 +40,11 @@
             Please wait...
         </div>
         <div id="service-worker-notsupported" hidden>
-            Your browser does not support serviceworkers
+            Please update your browser
+        </div>
+        <div id="service-worker-failed" onclick="location.reload()" hidden>
+            Sorry, please click to try again<br />
+            Deskulpa, favor koko fail
```
PeteCoward, line 47:
```suggestion
            Deskulpa, favor koko fali
```

PeteCoward:
> co-version app versions and data versions, so that your app doesn't get data it can't digest.

would love to get that in 

merging this as a step along the way, I hope to get support in for loading in anyway and not caching data if it's requested.

PR #68, opened 2020-04-27
Remove woff files

PeteCoward:
This gains us about 150kb of initial download size, in exchange for not supporting really old browsers

https://caniuse.com/#feat=woff2

double declaring woffs and wofff2s makes sense without a service worker as the downloads are optimized ( early browsers download and use woffs, recent ones only download woff2s ) But with pre-caching it makes everyone download both files. Our google analytics so far show no browsers that do not support woff2s accessing asteroid.tl


RobertoMaurizzi:
Do browsers that don't support woff2 support Service Workers? If not I'd say it's never worth it to precache .woff fonts...

PeteCoward:
I can't see a difference 
https://caniuse.com/#feat=serviceworkers 
https://caniuse.com/#feat=woff2

PR #70, opened 2020-04-29
Front-End Uses i18n Wagtail Content

sandroroux:


src/js/redux/Store.js
```diff
@@ -75,6 +75,11 @@ export const isBrowserSupported = () => {
     return store.getState().isBrowserSupported;
 };
 
+export const getHome = () => {
+    const home = store.getState().home.en;
```
PeteCoward, line 79: tet is the requested default language ( of the frontend ) by the way

src/riot/WagtailPages/HomePage.riot.html
```diff
@@ -18,131 +18,108 @@ <h6 if="{state.currentCourse && isCurrentCourseInProgress(state.currentCourse)}"
...
-                title="{state.nextLesson[course.data.id].title}"
-                description="{state.nextLesson[course.data.id].description}"
-                duration="{state.nextLesson[course.data.id].duration}"
-                link="{state.nextLesson[course.data.id].id}"
-                isCompleted="{isLessonComplete(course, state.nextLesson[course.data.id])}"
-                isComingSoon="{state.nextLesson[course.data.id].coming_soon}"
-            >
-            </Card>
-        </div>
+        <CoursePreview each="{course in state.courses}" course="{course}"></CoursePreview>
```
PeteCoward, line 21: I don't like using 'Preview' in two ways, can we find another name?

src/js/redux/Store.js
sandroroux: Done.

src/riot/WagtailPages/HomePage.riot.html
sandroroux: Good point. I'll think on it.

src/js/redux/Store.js
```diff
@@ -75,6 +78,17 @@ export const isBrowserSupported = () => {
...
+export const getHome = () => {
+    const language = getLanguage();
+    let home = null;
+    if (language === "tet") {
+        home = store.getState().home.en;
+    } else {
+        home = store.getState().home.en;
+    }
+    return home;
+};
```
sandroroux, line 90: We should avoid hard-coded language codes.

src/js/redux/_Reducers.js
```diff
@@ -43,16 +43,27 @@ const addHome = (state = {}, action) => {
     switch (action.type) {
         case ADDED_HOME_PAGE:
             const home = action.home;
+
+            let languageCode = null;
+            if (home.meta.slug.includes("tdt")) {
+                languageCode = "tet";
+            } else {
+                languageCode = "en";
+            }
```
sandroroux, line 52: More hard-coded language codes.

sandroroux:
@PeteCoward The front-end still must determine the language through slugs. You had said this was a fine option until we had one better.

As for that better option, the manifest or the page objects can carry the translation mapping. I propose that the page objects refer to their translation until we can refactor the manifest and maintain a site-tree data structure (with Redux, for example— its flatter page representation maintains a site tree).

For any of this to be possible, we should solve the canonical_ptr/page_ptr issue. Our mappings must be correct.

PR #71, opened 2020-04-30
Set far future expiry and secure on jwtToken cookie

PeteCoward:
This generalises the setCookie method to accept attributes and key only attributes allowing use to set an expiry and the secure attribute.

to address https://github.com/catalpainternational/asteroid/issues/68


boerwastaken:
The expiry date chosen here caused a new bug: catalpainternational/asteroid#73 .

PR #75, opened 2020-05-04
Lighthouse suggested feedback

PeteCoward:
- add theme_color
- define start_url
- use a 512px icon

bumps up our score on lighthouse a little, the only slightly controvesrial thing might be using only a 512 icon. 
I'd like to look into excluding some things from the precaching so we can provide the recommended range of icon sizes without making everyone download them all.


PeteCoward:
The performance metrics do change quite alot between runs. But this PR improves the PWA score ( which is a little irrelevant to us as we have Appelflap ) That section does not give a pct score like the rest, but just shows me different icons. this PR addresses some of the recommendations it made in that section

PR #81, opened 2020-05-06
Fix homepage "wave" image

katevoss:
It may be a bit controversial to set the `<body>` background color, but it does fix the issue of the top-menu's background.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -32,7 +32,7 @@
     <% } %>
 </head>
 
-<body>
+<body class="background-blue-1">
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -32,7 +32,7 @@
     <% } %>
 </head>
 
-<body>
+<body class="background-blue-1">
```
PeteCoward, line 35: Yes, controversial

PeteCoward:
merging but see #83

PR #82, opened 2020-05-06
Quickfix for tetum translation oddness

PeteCoward:
So the gettext library we are using has broken behaviour. This is a quick fix while I look into it further.

options:
1. use a different library
2. code it ourselves
3. keep this hacky fix
4. Actually get transifex to believe tetum is nplurals=2
5. investigate if it is po2json causing our problems


PeteCoward:
\o/ https://github.com/guillaumepotier/gettext.js/pull/34

PR #86, opened 2020-05-07
Safari uses jpegs

PeteCoward:
ReniditionSelector chooses jpeg rendition to cache and display if on Safari

### DIFF
--- a/src/js/RenditionSelector.js
+++ b/src/js/RenditionSelector.js
@@ -2,20 +2,18 @@ import { BACKEND_BASE_URL } from "js/urls"
 import { getManifestFromStore } from "ReduxImpl/Store"
 
 const FILTER_SPEC = 'width-800|format-webp';
+const SAFARI_FILTER_SPEC = 'width-600|format-jpeg';
 
 export function getImagePath(id) {
     const images = getManifestFromStore().images;
-    return _getRenditionUrl(images[id]);
+    return _getRenditionUrl(images[id], true);
 }
 
 export function getImageUrls(images) {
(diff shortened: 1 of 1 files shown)


src/js/RenditionSelector.js
```diff
@@ -18,17 +20,14 @@ export const getImagePath = async (imageId) => {
...
+    return Object.values(images).map((url) => _getRenditionUrl(url));
 };
 
-function _getRenditionUrl(renditions) {
-    return `${BACKEND_BASE_URL}/media/${renditions[RENDITION_FORMAT]}`;
-}
+function _getRenditionUrl(renditions, includeDomain=false) {
+    const filter_spec = navigator.userAgent.match(/Safari/g) ? RENDITION_FILTER_SPEC_SAFARI : RENDITION_FILTER_SPEC_DEFAULT;
+    return `${includeDomain ? BACKEND_BASE_URL : ""}/media/${renditions[filter_spec]}`;
+};
```
sandroroux, line 33: This looks good. Instead of `userAgent` matching, can we reuse this code in `Settings.riot.html`?
```javascript
shouldShowAddToHomeScreen() {
    const { name } = getBrowserVersion(navigator.userAgent);
    const inSafari = name === "Safari";
    const inPWAMode = window.matchMedia("(display-mode: standalone)").matches;
    return !(inSafari || inPWAMode || inAppelflap());
},
```
Call it something like `utilities.js::isAppInSafari()`.

PeteCoward: I missed that code, was copying the code from another place we use userAgent

PR #89, opened 2020-05-07
getMostRecentCompletion is language aware.

sandroroux:


src/js/actions/completion.js
```diff
@@ -125,12 +126,25 @@ export const isTheCourseComplete = (courseSlug, coursesLessons) => {
     return numberOfCoursesCompletions === numberOfCoursesLessons;
 };
 
+const isCourseInTheCurrentLanguage = (courseSlug) => {
+    const currentLanguage = getLanguage();
+    if (currentLanguage === "en") {
+        return !courseSlug.includes("tdt");
+    } else {
+        return courseSlug.includes("tdt");
+    }
```
sandroroux, line 135: Hard-coding!

PeteCoward:
OK, lets get this into one PR, I'm only looking at the diff between this branch and master :)

We can't rely on the slug to determine language, slugs can be edited by users!
We should use the TranslatablePage.language.code, but in order to get at that efficiently we need to change the query to:

`TranslatablePage.objects.filter(pk__in=site.root_page.get_descendants(inclusive=False))`
"Get all the translatable pages who are in the descendents of the site root ( not including the translatable root )"

This is because the wagtail get_descendents method returns a `Page` queryset, not a `TranslatablePage` queryset, so it does not understand the link to language.
You should be able to make that change in line 164 of elearning_content/views.py, then you can add 'language__code' to the values, just like you have added 'slug'

Then getting the language home pages is just find the first page for each language code in the result set, or all the direct children of the translatable root

PR #90, opened 2020-05-08
Refactor platform detection

PeteCoward:
In response to https://github.com/catalpainternational/canoe/pull/86#discussion_r421801728

- moves the platform detection out of the DjangoPushNotifications module into it's own module
- edits to Notifications to support this
- edits to Settings to support this


PeteCoward:
The Chrome on iOS is a candidate to move into this method, but I left it alone due to it being a bit special.

PR #91, opened 2020-05-10
use UA for appelflap detection; feasible since my workaround for UA/C…

boerwastaken:
use UA for appelflap detection; feasible since my workaround for UA/CORS issue — see https://bugzilla.mozilla.org/show_bug.cgi?id=1629921#c5 .

Correlated change in Appelflap is https://github.com/catalpainternational/appelflap/commit/30fc73f33c1b526f5d05c1cf2ffe6bea1170baf5

### DIFF
--- a/src/js/PlatformDetection.js
+++ b/src/js/PlatformDetection.js
@@ -25,8 +25,7 @@ export function getPlatform() {
 };
 
 function inAppelflap() {
-    const foundAppelflap = navigator.languages.find((word) => /^ep-[a-j]{4,5}$/.test(word));
-    return !!foundAppelflap;
+    return navigator.userAgent.startsWith("io.catalpa.canoe.");
 };
 
 function inPWAMode() {


PeteCoward:
Do you know what the behaviour of the browser name / version bit is with the overridden user agent?
Also whether the Google Analytics Browser and Browser Version dimensions are still available?

boerwastaken:
> Do you know what the behaviour of the browser name / version bit is with the overridden user agent?

Stays as it was (`appCodeName`, `appName`, `appVersion`).

> Also whether the Google Analytics Browser and Browser Version dimensions are still available?

That's a rather vague question even if I'd be familiar with and had access to your spyware suite, which I am not and which I don't.
This is the information available to it, if it doesn't do what you want with this info, go complain to your spyware manufacturer overlord :stuck_out_tongue_closed_eyes: 

navigator

PeteCoward:
I can't see in the linked pr exactly what the user agent is set to ( other than that it presumably starts with `io.catalpa.canoe` ).
What i meant but did not explain well is the `getPlatform` method https://github.com/catalpainternational/canoe/pull/91/files#diff-0e857b8852d6fcaf9306ba2eaf18e256L2 parses the user agent string to get the browser and version. How does that respond to what you have configured in your useragent override?

boerwastaken:
> I can't see in the linked pr exactly what the user agent is set to

then look at the posted picture just above your comment, it tells you exactly what the user agent is set to.

boerwastaken:
> the getPlatform method [...] parses the user agent string to get the browser and version. How does that respond to what you have configured in your useragent override?

Well, you're the "programmer" who copy-pasted [that code from StackOverflow](https://stackoverflow.com/a/16938481)\*. So if you have questions about whether that code is robust to appearance of an unexpected UA string, you should go and ask yourself whether it's a robust implementation and what the false positive / false negative rate might be.

Next time please add a comment in your code with reference to source when you copy-paste nontrivial chuncks-o-code. Not for the attribution aspect, that's a question of your personal honour, but just so that we know what kind of questions to ask about it.

In this case my question would be: wouldn't feature probing get you where you want to be?

\*) or you copied it from someone who copied it from there or somewhere else, or …

PeteCoward:
What is the value you pass as User Agent?

boerwastaken:
navigator

boerwastaken:
Sandro asked for a more direct way of determining whether we're in Appelflap than [this]((https://github.com/catalpainternational/canoe/pull/59/files). which I agree is desirable, using the User-Agent instead of some TCP port number encoded into `navigator.languages` satisfies that and this has now become possible thanks to a workaround for Mozilla bug 1629921.

PeteCoward:
I'm not complaining, or reviewing. I was just asking so I could understand what those values would be when running in appelflap. I missed your comment https://github.com/catalpainternational/canoe/pull/91#issuecomment-627189094. 

That method is only used in the Notifications subscription ( which we hide in appelflap for now ), and the detection of Chrome running on Safari, so it should not cause a problem.

Is there any difference between the user agents / browser versions sent by the different builds of a given appelflap release, and would that be interesting data for us to have?

boerwastaken:
All clear then!

> Is there any difference between the user agents / browser versions sent by the different builds of a given appelflap release, and would that be interesting data for us to have?

The way I've set it up now is `application id - version - (debug|release) / platform`. The application ID identifies the build product flavour; for instance there's an `io.catalpa.canoe.asteroid.release` (as opposed to `.preview`) too. In the past using the user agent was occasionally useful serverside, for instance in Eskola to show users a "please upgrade" page. For detecting whether youre "in the app" it's also handy, I expect a false negative rate of zero (because we're building the app) and a near-zero false positive rate (the rate would change when someone (accidentally? deliberately? maliciously?) starts distributing a browser with a UA starting with `io.catalpa.canoe.projectname` and our users start running the PWA in it).

We'll need to keep it a safe assumptions that the Appelflap UA starts with the application ID (and that the application ID does not change), as as of now certain functionality depends on it.

PeteCoward:
@sandroroux please have a look

boerwastaken:
There's another layer of copy/paste indirection: https://github.com/jazzband/django-push-notifications/commit/05c4290d26d76fd14db86dc3380d3a67fbdf15ff#diff-88b99bb28683bd5b7e3a204826ead112R174-R194

sandroroux:
Yup, @boerwastaken. We inherited the code from Django Push.

PR #92, opened 2020-05-11
Back button grabs page in correct language

sandroroux:
### NOTE
When you back-button to a page not in the current language, these changes grab the correct version of the page. This swap breaks the ability to go back.

You back-button to a URL, which grabs a new page and routes to the new page's URL. If you back-button again, you hit the old page's URL, which routes you to the new page's URL. You can’t escape this loop.

### DIFF
--- a/src/js/Routing.js
+++ b/src/js/Routing.js
@@ -3,6 +3,7 @@ import {
     getOrFetchWagtailPage,
     getOrFetchManifest,
     getHomePage,
+    getWagtailPage,
 } from "js/WagtailPagesAPI.js";
 
 const IS_SETTINGS_NOTIFICATIONS_OR_PROFILE = /#([A-Za-z]+)/;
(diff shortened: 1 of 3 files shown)


src/js/Routing.js
```diff
@@ -20,9 +21,39 @@ const getPreviewPageUrl = (queryString) => {
...
+export const getWagtailPageOrRouteToTranslation = async (pageId) => {
     const manifest = await getOrFetchManifest();
+    const pageUrl = manifest.pages[pageId];
+    let page = await getOrFetchWagtailPage(pageUrl);
+
+    const currentLanguage = getLanguage();
+    const translationInfo = page.data.translations;
 
+    // If Canoe ever adds three or more languages, the rest of the function must
+    // change.
```
PeteCoward, line 45: then we should remove 'pt' from the backend

src/js/Routing.js
```diff
@@ -49,6 +79,7 @@ export const getPage = async () => {
         page = await getHomePage();
     }
 
+    console.log("Got page.");
```
PeteCoward, line 82:
```suggestion
```

PR #93, opened 2020-05-12
i18n-and-Image Error Handling

sandroroux:
* Site works with one language and, without languages, fails gracefully.
* Site detects when the manifest lacks a rendition and fails gracefully.

### DIFF
--- a/src/js/RenditionSelector.js
+++ b/src/js/RenditionSelector.js
@@ -33,7 +33,7 @@ const getPlatformSpecificRendition = () => {
     if (browser.name === "Safari") {
         renditionType = JPEG_RENDITION;
     } else {
-        renditionType = JPEG_RENDITION;
+        renditionType = WEBP_RENDITION;
     }
     return renditionType;
 };
--- a/src/js/RenditionSelector.js
+++ b/src/js/RenditionSelector.js
@@ -30,4 +30,27 @@ export function getImageUrls(images) {
 function _getRenditionUrl(renditions, includeDomain=false) {
     const filter_spec = navigator.userAgent.match(/Safari/g) ? RENDITION_FILTER_SPEC_SAFARI : RENDITION_FILTER_SPEC_DEFAULT;
     return `${includeDomain ? BACKEND_BASE_URL : ""}/media/${renditions[filter_spec]}`;
+    return Object.values(images).map(_getRenditionUrlWithoutDomain);
+}
+
+export const _getRenditionUrlWithoutDomain = (renditions) => {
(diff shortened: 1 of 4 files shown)


PeteCoward:
I'm going to leave this open for now. I want to talk about error handling. We need to decide if we swallow errors or communicate them

src/js/RenditionSelector.js
```diff
@@ -1,9 +1,9 @@
 import { BACKEND_BASE_URL } from "js/urls";
 import { getOrFetchManifest } from "js/WagtailPagesAPI";
+import { getPlatform } from "js/PlatformDetection";
 
-const RENDITION_FILTER_SPEC_DEFAULT = 'width-800|format-webp';
-const RENDITION_FILTER_SPEC_SAFARI = 'width-600|format-jpeg';
-
+const WEBP_RENDITION = "width-800|format-webp";
+const JPEG_RENDITION = "width-600|format-jpeg";
```
sandroroux, line 6: @PeteCoward, I'm removing "filter_spec" because "filter_spec" is a back-end concept (and a name that returns vague Google results).

To a front-end developer, these are keys to the image type they want.

src/js/RenditionSelector.js
```diff
@@ -12,22 +12,48 @@ export class MissingImageError extends Error {
     }
 }
 
-export const getImagePath = async (imageId) => {
+export const getImageUrl = async (imageId) => {
```
sandroroux, line 15: @PeteCoward I standardized this file's naming convention. It differentiates [between a "path" and a "URL".](https://doepud.co.uk/blog/anatomy-of-a-url)

A URL includes the protocol and domain (the origin). A path doesn't.

src/js/RenditionSelector.js
```diff
@@ -12,22 +12,48 @@ export class MissingImageError extends Error {
...
     const manifest = await getOrFetchManifest();
     const images = manifest.images;
     const image = images[imageId];
 
     if (!image) {
         throw new MissingImageError(`Image with ID, ${imageId}, doesn't exist.`);
     }
-    return _getRenditionUrl(image, true);
```
sandroroux, line 23: @PeteCoward, in JS, I think we should avoid methods with [flag-switches](https://martinfowler.com/bliki/FlagArgument.html). I like single-responsibility methods.

src/js/RenditionSelector.js
PeteCoward: I added it precisely because it maps to the backend concept!

PR #95, opened 2020-05-13
73 other cookie issues

boerwastaken:
More cookie issues.

This fixes catalpainternational/asteroid#73 on Geckoview. Really now. Hopefully.

First issue (inconsequential, but I was there anyway) is an misformatted date string for the absolute expiry, at least if I follow the MDN documentation there should be a comma after the DOW. Then I thought, these expiry datetime strings look error-prone, let's use max-age instead so we can have a simple integer value.

To (finally?) solve my login problem I had to change `samesite=strict` to `samesite=lax`. Otherwise the `document.cookie` simply stays... empty, and it seems canoe doesn't like that. The `samesite=strict` cookie gets stored allright, it just seems to not be available to the JS context. I'm not sure why that is or whether it's a bug in Gecko. It's not recorded as `httponly` so that's not it.

Can we have `samesite=lax` instead?

### DIFF
--- a/src/js/AuthenticationUtilities.js
+++ b/src/js/AuthenticationUtilities.js
@@ -69,7 +69,7 @@ export const login = async usernameAndPassword => {
     // Browsers refuse to set secure cookies from non https locations
     setCookie(JWT_TOKEN_STORAGE_KEY, token,
         window.location.protocol === "https:" ? ["secure"] : [],
-        {"max-age": 3e8, samesite: "strict"}
+        {"max-age": 3e8, samesite: "lax"}
     );
     localStorage.setItem(USERNAME_STORAGE_KEY, username);
     localStorage.setItem(USER_ID_STORAGE_KEY, userId);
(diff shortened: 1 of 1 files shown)


boerwastaken:
> The danger I assumed was sending that cookie along with other requests made to origins we don;t control, off the top of my head the only time we do that currently is with Google Analytics

But that's not how cookies work? The domain of the cookie doesn't match the domain of google analytics. Or is some Asteroid code blindly attaching cookies to any request it can get its hands on, is that what you mean? I'm not sure a same-site policy would even help against that, as it's a policy for the browser, not for what you're doing in JS-land, right?

boerwastaken:
I think that too. What would happen in that scenario is that asteroid.tl would see the JWT cookie attached to the img request, and presumably we can trust asteroid.tl.

PeteCoward:
I believe it may also be somesubdomain.asteroid.tl, but likewise :)


Peter Coward
Senior Software Engineer
Catalpa International
m +67075448273
w http://www.catalpa.io

boerwastaken:
Well if that's undesirable, then set the host-only flag, it's what it's for.

PR #96, opened 2020-05-13
isTheCourseComplete uses new completion counting.

sandroroux:
Closes catalpainternational/Asteroid#62.

### DIFF
--- a/src/js/actions/completion.js
+++ b/src/js/actions/completion.js
@@ -87,22 +87,12 @@ export function isComplete(course, lesson, section) {
 }
 
 export const isTheCourseComplete = (courseSlug, lessonSlugs) => {
-    const numberOfCompleteLessons = countNumberOfCompleteLessons(courseSlug, lessonSlugs);
+    const numberOfCompleteLessons = countCompleteLessonsInCourse(courseSlug, lessonSlugs);
     const numberOfLessonsInCourse = lessonSlugs.length;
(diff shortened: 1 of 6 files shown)


src/js/actions/completion.js
```diff
@@ -113,19 +114,27 @@ export const countNumberOfCompleteLessons = (courseSlug, lessonSlugs) => {
...
-    const numberOfCoursesLessons = coursesLessons.size || coursesLessons.length;
-    return numberOfCoursesCompletions === numberOfCoursesLessons;
+const isTheCourseComplete = (courseSlug, lessonSlugs) => {
+    const numberOfCompleteLessons = countNumberOfCompleteLessons(courseSlug, lessonSlugs);
+    const numberOfLessonsInCourse = lessonSlugs.length;
+    return numberOfCompleteLessons === numberOfLessonsInCourse;
 };
 
-export function getMostRecentCompletion() {
+export const isCurrentCourseInProgress = (aWagtailCourse) => {
```
sandroroux, line 123: @PeteCoward What do you think of `completion.js` methods manipulating wagtail page JSON?

An alternative: I can create an canoe-side interface that calls `completion.js`. I would move `isCurrentCourseInProgress` and `getLatestCompletion` to that interface.

PeteCoward: I am not a fan of it at all.

sandroroux: Roger that. Interface it is.

PR #99, opened 2020-05-15
Audio/video content card style

katevoss:
Contributes to https://github.com/catalpainternational/olgeta/issues/214

Depends on this backend PR: https://github.com/catalpainternational/canoe-backend/pull/18


src/img/film.svg
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
```
sandroroux, line 1: @katevoss, did you pass these SVGs through [vecta.io](https://vecta.io/nano)?

See https://github.com/catalpainternational/canoe/pull/53 for context.

katevoss: I'll do that, thanks for the reminder!

src/riot/Components/Media.riot.html
```diff
@@ -1,32 +1,35 @@
...
-                src="{prefixUrlWithBaseUrl(source.src)}"
-                type="{source.type}"
+                src="{prefixUrlWithBaseUrl(props.media.sources[0].src)}"
+                type="{props.media.sources[0].type}"
             />
             <span translate>Your browser does not support the audio element.</span>
         </audio>
+        <p>{state.cached ? 'This audio will play offline.' : 'This audio is not yet available offline.'}</p>
+    </div>
+    <div if="{props.media.sources[0].type === 'video/mp4'}">
```
sandroroux, line 15: When course authors don't use mp4, we should error:
```html
<div if="{props.media.sources[0].type !== 'video/mp4'}">
<p class="error-message">Error: Asteroid only supports MP4. Please upload "{props.media.title}" as an MP4.<p>
</div>
```

sandroroux: We should do the same for audio.

src/riot/Components/Media.riot.html
```diff
@@ -1,32 +1,35 @@
-<media>
-    <div>
-        <p>{state.cached ? 'This video will play offline.' : 'This video is not yet available offline.'}</p>
-    </div>
-
-    <div>
-        <audio if="{props.media.media_type === 'audio'}" controls crossorigin="anonymous">
+<Media>
+    <div if="{props.media.sources[0].type === 'audio/mpeg'}">
```
sandroroux, line 2: We could split these into `Audio` and `Video` tags. It may improve readability.

src/riot/Components/Media.riot.html
katevoss: Should this validation be on the Media uploader in `canoe-backend`? Otherwise this message will show to end users, who can't change the type of video.

PeteCoward: This source type should not control if video or audio is shown. It is like a mimetype for the file, not the highter level 'media/video' swtich, which should still be in `props.media.medai_type`

PeteCoward: Except it is not because I accidentally removed it https://github.com/catalpainternational/canoe-backend/pull/18/files#diff-106d241f2da10a0dc93d80876891cb32R133

PeteCoward: see https://github.com/catalpainternational/canoe-backend/pull/18/files#r426990496

src/riot/Components/Media.riot.html
katevoss: I think I agree, though it would mean duplicating some of the functions like `prefixUrlWithBaseUrl`, which could be put into a shared utils like `js/utilities.js`, though it doesn't currently have many shared utilities like this...so I'll start adding them 😄

sandroroux: Yup. I think code sharing is great. Feel free to put `prefixUrlWithBaseUrl` in `utilities.js`, or to create a file called `URLAndPathTools.js`. I know some methods that could move to a `URLAndPathTools`.

My translation PR, #70, distinguishes between paths and URLs. `http://localhost:8000/api/v2/pages/1` is a URL. A URL includes a protocol (`http://`) and domain (`localhost:8000`). This example URL's path is `/api/v2/pages/1`.

If you want to follow my naming standard, you can rename `prefixUrlWithBaseUrl` to `getMediaUrl(mediaPath)`.

katevoss: Changes made, would you take another look @sandroroux ?

PR #105, opened 2020-05-20
Checks service worker for notification subscription.

sandroroux:
Closes #84.


sandroroux:
@boerwastaken Will do. Thanks for the review.

PR #113, opened 2020-05-26
Feedback completion improvements

katevoss:
Closes https://github.com/catalpainternational/asteroid/pull/89

### DIFF
--- a/src/scss/_lesson.scss
+++ b/src/scss/_lesson.scss
@@ -716,15 +716,11 @@ LessonFeedback {
     .thank-you-container {
         @include feedback-containers;
 
-        .smiling-smiley {
-            height: 4em;
-            width: 4em;
-            margin-bottom: 2em;
+        .correct-flag {
+            height: 6em;
+            width: 6em;
+            margin: auto;
         }
-    }
-
-    .thank-you-container.btn-primary {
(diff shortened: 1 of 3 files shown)


src/scss/_lesson.scss
```diff
@@ -716,17 +716,13 @@ LessonFeedback {
     .thank-you-container {
         @include feedback-containers;
 
-        .smiling-smiley {
-            height: 4em;
-            width: 4em;
-            margin-bottom: 2em;
```
sandroroux, line 722: The diff muddles whether you deleted `.smiling-smiley`. Keep in mind that `LessonFeedback.riot.html` line 15 needs `smiling-smiley`.

katevoss: I replaced `smiling-smiley` with `correct-flag` because the class in line 719 is a child of the `thank-you-container` parent class, which only applies to that specific "smiley" (now "flag").

The `smiling-smiley` in LessonFeedback.riot.html line 15 is styled by `button.feedback` on line 730 of _lesson.scss, which styles all the other "smileys". Does that clarify it?

sandroroux: Thanks for explaining. Sounds good.

Issue #103, opened 2020-05-18
Language setting is not persisted


boerwastaken:
Observed behaviour:

1. When I choose a site language, it switches to that language, as it should.

2. When I close the tab, and then navigate to the site (https://asteroid.temp.build) in a new tab, stuff appears in not-my-choice from step 1 and the language choice menu also shows not-my-choice from step 1 selected.

Expected behaviour:

- [ ]  Language choice is persisted

PeteCoward:
This was a stretch goal. Thanks for highlighting though. I too think it's worth prioritising

boerwastaken:
https://docs.djangoproject.com/en/3.0/ref/settings/#std:setting-LANGUAGE_COOKIE_NAME

katevoss:
@boerwastaken I don't believe that the Django backend cares what language the frontend is in. I believe that the purpose is for end-users to be able to set their language and for their selection to persist even if they reload or refresh the page. 

@PeteCoward If we use a cookie to set their language preference, would we be required to get the user to agree to cookies? I don't believe that Timor Leste is covered under the GDPR but it might be a good idea regardless.

boerwastaken:
> If we use a cookie to set their language preference, would we be required to get the user to agree to cookies? I don't believe that Timor Leste is covered under the GDPR but it might be a good idea regardless.

As we don't need to let any server know what language the user wants, we don't need to use a cookie, and if we'd use something local-only instead of a cookie (localstorage springs to mind) then we won't need to worry about GDPR at all :innocent: 

Btw so far afaik we haven't worried about GDPR, while at the same time in most projects we're pretty gung ho about flinging Google Analytics spyware at unsuspecting users. Hmmmm :thinking:

PR #114, opened 2020-05-26
Store language in localStorage

katevoss:
closes https://github.com/catalpainternational/canoe/issues/103

This PR does a few things:
1. When a user first loads the app, it looks for a `userLangage` in localStorage, which is where it is stored from their last visit. If they are a first-time user, then localStorage doesn't have the key.
2. Next it looks into their browser language. If the language is some variation of `en`, canoe changes the language to English.
3. Finally, if the browser's language is anything but English it sets the default language to Tetum.
4. When a user changes their language on the Settings page, it updates the language in Store and it stores the language in localStorage, so that it will persist in future sessions.

Feedback and suggestions welcome.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -6,17 +6,6 @@ import App from "RiotTags/App.riot.html";
 import { store, changeLanguage } from "ReduxImpl/Store";
 import "./scss/canoe.scss";
 
-const LANGUAGE_STORAGE_KEY = "userLanguage";
-const storedLanguage = localStorage.getItem(LANGUAGE_STORAGE_KEY);
-
-if (!!storedLanguage) {
-    changeLanguage(storedLanguage);
-} else if (navigator.language.includes("en")) {
-    changeLanguage("en");
-} else {
-    changeLanguage("tet");
-}
-
 riot.install(function (component) {
     // all components will pass through here
     installTranslationPlugin(component);
--- a/src/app.js
+++ b/src/app.js
@@ -6,7 +6,11 @@ import App from "RiotTags/App.riot.html";
 import { store, changeLanguage } from "ReduxImpl/Store";
 import "./scss/canoe.scss";
 
-changeLanguage('tet');
+if (navigator.language.includes("en")) {
+    changeLanguage("en");
+} else {
+    changeLanguage("tet");
+}
 
 riot.install(function (component) {
     // all components will pass through here
--- a/src/app.js
+++ b/src/app.js
@@ -3,10 +3,15 @@ import { installReduxPlugin } from "ReduxImpl/RiotReduxPlugin";
 import { installTranslationPlugin } from "RiotTranslationPlugin";
 
 import App from "RiotTags/App.riot.html";
-import { store, changeLanguage } from "ReduxImpl/Store";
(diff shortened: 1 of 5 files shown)


src/app.js
```diff
@@ -6,7 +6,16 @@ import App from "RiotTags/App.riot.html";
 import { store, changeLanguage } from "ReduxImpl/Store";
 import "./scss/canoe.scss";
 
-changeLanguage('tet');
```
sandroroux, line 9: The site language is a Redux value. There's a more complicated, but more architectural way to do this: we should [initialize the Redux state](https://redux.js.org/recipes/structuring-reducers/initializing-state) with the correct language. Your code can set the Redux store's default state.

Wrap the code in a function:
```js
const getInitialLanguage = () => {
    const storedLanguage = localStorage.getItem(LANGUAGE_STORAGE_KEY);
    if (!!storedLanguage) {
        return storedLanguage;
    } else if (navigator.language.includes("en")) {
        return "en";
    } else {
        return "tet";
    }
}
```

and pass it to `createStore` as the `language` initial value:
```js
// This method is at Store.js's line 16.
const initialStoreState = {
    language: getInitialLanguage()
};

export const store = createStore(
    reducers,
    initialStoreState,
    window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__()
);
```

Let me know if you'd like me to explain this approach more.

src/js/redux/Store.js
```diff
@@ -48,6 +48,9 @@ export const storeSiteDownloadedIs = (trueOrFalse) => {
 
 export const changeLanguage = (language) => {
     store.dispatch({ type: LANGUAGE_CHANGE, language: language });
+
+    const LANGUAGE_STORAGE_KEY = "userLanguage";
```
sandroroux, line 52: I'd suggest you define `LANGUAGE_STORAGE_KEY` once. If future developers must change the key, we want them to update one piece of code.

If you follow my other comment's approach, the key can stay in `Store.js`.

src/js/redux/Store.js
```diff
@@ -48,6 +69,8 @@ export const storeSiteDownloadedIs = (trueOrFalse) => {
 
 export const changeLanguage = (language) => {
     store.dispatch({ type: LANGUAGE_CHANGE, language: language });
+
+    localStorage.setItem(LANGUAGE_STORAGE_KEY, language);
```
PeteCoward, line 73: I'd put this in the storeDispatch method inside Tranlsations.js

PeteCoward: or the setLocale method

PeteCoward: but I can see why it is here I suppose

sandroroux: Yeah, @PeteCoward. I'd rather hide the store behind this interface. In fact, we could move these read/write methods into their own file.

katevoss: @sandroroux I'm planning to merge this PR, did you want to me to move the read/write methods? Or should that be a separate issue?

sandroroux: Definitely a separate task. Thanks for merging.

PR #116, opened 2020-05-28
Keep images from shrinking

katevoss:
Closes https://github.com/catalpainternational/asteroid/issues/92

Can "close" catalpainternational/olgeta#287.

### DIFF
--- a/src/scss/_login.scss
+++ b/src/scss/_login.scss
@@ -1,5 +1,5 @@
 Authentication > div {
-    min-height: calc(100vh - 20px - 20px);
+    min-height: 100vh;
     flex-direction: column;
     align-items: center;
     display: flex;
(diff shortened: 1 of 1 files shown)


src/scss/_login.scss
```diff
@@ -1,6 +1,5 @@
 Authentication > div {
-    height: calc(100vh - 20px - 20px);
-    justify-content: center;
+    min-height: calc(100vh - 20px - 20px);
```
sandroroux, line 2: @katevoss I think removing this `min-height` solves the issue. What do the other changes bring?

katevoss: I removed `justify-content: center` because it was causing the logos to scroll off the top of the page if the viewport was too short. And `flex: none` keeps the logos from being responsive and shrinking. Let me know if you need any more info.

PR #117, opened 2020-05-28
Project Overriding EOD wip

PeteCoward:
The key change here is in the webpack config resolve.modules, adding the `src/` folder as a module import location, and configuring the project config merge to prepend to that array not append.

This allows a project to override an individual js module, sass file, or image really easily.

The imports need to be changed to module imports though, relatice imports cannot be overridden. `url()` imports in css need to have the `~` prefix to  enable this.

Also removes the extra `modules/` folder we had configured, which tidies things up a little


src/scss/_lesson.scss
```diff
@@ -171,10 +169,6 @@ LessonPage h2 {
     border-bottom: none;
 }
 
-LessonPage .hide {
-    display: $hideInCanoe;
-}
```
PeteCoward, line 176: this will be replaced by a rule added to `scss/project_themables.scss` in asteroid

PeteCoward:
Examples of how to override things...

Implementing projects provide a path to a folder ( js module ) in the canoe-project.js config file
```
module.exports = {
    WEBPACK_CONFIG: {
        resolve: {
            modules: [path.resolve(__dirname, "canoe-overrides")],
        }
    },
...
```

The webpack config merge will prepend that to the default resolve.modules which includes `canoe/src` and `node_modules`

Webpack will then look for non-relatively imported assets in these places in this order:
1. any project configured override locations
2. `canoe/src`
3. node_modules
It will use the first one it finds

## images
for any image imported into css with `url(~img/img.svg)` the implementing project just needs to drop an identically named image in their configured `canoe-overrides/img` folder
NOTE: this only works if you prepend the path in url with `~`

## JS moduiles, Riot tags, scss files
Any source code module can be overriden just by dropping a new module in the canoe-overrides path
NOTE: this only works if the import is not relative
`import "js/mymodule";` can be overridden
`import "../js/mymodule"; cannot be overridden

We should change all imports to non-relative imports to allow easy overriding in future, and avoid problems with imports from inside overridden modules. I have not done that to keep this initial PR simple as possible

sandroroux:
For JS and Riot, we should hew towards npm modules.

Check out [this tutorial](https://medium.com/recraftrelic/building-a-react-component-as-a-npm-module-18308d4ccde9) for making a React component. We can apply the principles to our Riot tags and JavaScript modules.

PeteCoward:
> For JS and Riot, I think we should hew towards npm modules.
> 
> Check out [this tutorial](https://medium.com/recraftrelic/building-a-react-component-as-a-npm-module-18308d4ccde9) for making a React component. We can apply the principles to our Riot tags and JavaScript modules.

Effectively this does head in that direction, encouraging that pattern at least. I don't like the idea of actually installing components which are under active development in projects as npm modules ( the code ends up in`/node_modules` so is not obvious how to edit, or painful to change / publish / etc ).

PR #126, opened 2020-06-04
Card-specific alignments

katevoss:
Closes https://github.com/catalpainternational/asteroid/issues/97


src/riot/Components/QuoteCard.riot.html
```diff
@@ -1,4 +1,7 @@
 <QuoteCard>
-    <h3><span>"</span>{props.card.quote}<span>"</span></h3>
+    <span>"</span>
```
sandroroux, line 2:
```suggestion
    <span>“</span>
```

PR #127, opened 2020-06-05
Quick references page and navigation

katevoss:
Closes https://github.com/catalpainternational/asteroid/issues/102

Notes: 
* ~~The `active` state in `References.riot.html` for the icons isn't triggered by `this.update({ iconActive: e.srcElement.className });`, any idea why?~~ Fixed in another branch
* I ~~commented out~~ **removed** the Notifications page registration and the icon in the Bottom menu (TBA whether we're removing it completely)
* I added the back arrow in the top-left corner (TBA whether we're keeping it)
* ~~The search, favorite, references, and filter icons aren't correct but they'll go in a future PR~~ Removed since the scope has changed


src/riot/Components/BottomMenu.riot.html
```diff
@@ -1,8 +1,9 @@
 <BottomMenu>
     <footer if="{!state.isObjective && !state.isContent && !state.isTest}">
         <a class="courses-icon { props.contentType === 'HomePage' ? 'active' : ''}" href="#"></a>
-        <a class="notifications-icon { props.contentType === 'notifications' ? 'active' : ''}" href="#notifications"></a>
+        <!--<a class="notifications-icon { props.contentType === 'notifications' ? 'active' : ''}" href="#notifications"></a>-->
         <a class="profile-icon { props.contentType === 'profile' ? 'active' : ''}" href="#profile"></a>
+        <a class="references-icon { props.contentType === 'references' ? '' : ''}" href="#references"></a>
```
sandroroux, line 6: Is this supposed to be:
```html
<a class="references-icon { props.contentType === 'references' ? 'active' : ''}" href="#references"></a>
```

katevoss: Yes, well spotted! I will switch it to include the "active" state, thanks

src/scss/canoe.scss
```diff
@@ -43,7 +44,8 @@ div.content-wrapper {
 
 .notifications,
 .profile,
-.settings {
+.settings,
+.references {
     padding-top: 65px;
     margin-top: -65px;
```
sandroroux, line 50: We don't need this screen-sliding negative margin. In fact, we should avoid it. TopMenu sets its own `background-color`: for Profile, the TopMenu is green; for Notifications, it's orange; for Home and Cards, it's blue.

Let's delete negative margins.

src/scss/_references.scss
```diff
@@ -0,0 +1,55 @@
+References .content-wrapper {
+    border-radius: 1.875em 1.875em 0 0;
+    background-color: $gray-10;
+    padding-top: 1.875em;
```
sandroroux, line 4: Could you use `CardFrame`? It grants the rounded corners and sizes the screen predictably. You may need to left-align some things in the CardFrame because the CardFrame center-aligns by default.

src/riot/Components/BottomMenu.riot.html
```diff
@@ -1,8 +1,8 @@
 <BottomMenu>
     <footer if="{!state.isObjective && !state.isContent && !state.isTest}">
         <a class="courses-icon { props.contentType === 'HomePage' ? 'active' : ''}" href="#"></a>
-        <a class="notifications-icon { props.contentType === 'notifications' ? 'active' : ''}" href="#notifications"></a>
         <a class="profile-icon { props.contentType === 'profile' ? 'active' : ''}" href="#profile"></a>
+        <a class="references-icon { props.contentType === 'references' ? 'active' : ''}" href="#references"></a>
```
sandroroux, line 5: The `active` class should turn `references-icon` blue. Do we have the "active" asset yet?

katevoss: Not yet, I don't have the assets yet for either the inactive or active icons.

src/scss/canoe.scss
katevoss: You are correct that `.references` doesn't need this negative margin. 

However we do still need the `background-color: $tan` for Settings as it colors the background of the Asteroid logo. Without it it looks like this: 
image

And Profile needs `background-color: $tan` too, though not the negative margin/padding:
image

src/scss/_references.scss
katevoss: I tried using `.card-frame` instead of `.content-wrapper` but it broke the search icon and would need some adjustments, so I think I'm going to keep it as-is unless you object.

sandroroux: Yup. That's fine. We can always loop back.

src/scss/canoe.scss
sandroroux: Yeah, sorry. I should've been clearer.

The page's background can stay as is. If you look in `TopMenu.riot.html`, you'll see that TopMenu sets its color to the page that appears underneath it:

```js
updateState() {
                const [afterTheHash, lessonModule] = parseURLHash();
                ...
                // Bunch of code that figures out which page we're on.
                ...

                if (this.state.isProfile) {
                    this.state.color = "profile-green";
                } else if (this.state.isNotifications) {
                    this.state.color = "notifications-orange";
                } else if (this.state.isHome || this.state.isCard) {
                    this.state.color = "blue-1";
                } else {
                    this.state.color = "";
                }
```

I think you can add to the conditionals and set the TopMenu $tan:
```js
if (this.state.isProfile) {
    this.state.color = "profile-green";
} else if (this.state.isNotifications) {
    this.state.color = "notifications-orange";
} else if (this.state.isHome || this.state.isCard) {
    this.state.color = "blue-1";
} else if (this.state.isSettings) {
    this.state.color = "settings-tan"; // or, "tan-1", or whatever you want to call the background-color
} else {
    this.state.color = "";
}
```

katevoss: I started going down that path and found myself in a quagmire of css to make the background the correct color. Your suggestion changes the TopMenu to tan but other parts of the background stay blue in a wider screen. 

I might be missing something though, was `TopMenu.riot.html` the only place that needed a fix? Or was it in addition to removing the highlighted section from `canoe.scss`?

katevoss: To be addressed in this future issue: https://github.com/catalpainternational/asteroid/issues/110

PR #129, opened 2020-06-09
The front-end analog to Peter's removing of course.colour.

sandroroux:


src/js/redux/_Reducers.js
```diff
@@ -79,7 +79,6 @@ const addCourse = (state = {}, action) => {
                     type: course.meta.type,
                     title: course.title,
                     slug: course.data.slug,
-                    color: course.data.colour,
```
PeteCoward, line 82: I'd prefer the store to store the page data from wagtail as is to avoid this kind of chasing around schema changes

PR #130, opened 2020-06-10
Add cards to quick references

katevoss:
Closes https://github.com/catalpainternational/asteroid/issues/103

This branch is off of this branch/PR: https://github.com/catalpainternational/canoe/pull/127 That PR (#127) should be merged first, then this one.

There is the beginning of Search here but I've moved most of it to another branch to keep my PRs a reasonable length. Sorry for the leaking concepts.


katevoss:
@PeteCoward I changed the base branch for this PR so the full set of commits will now appear here, and I updated this PR with commits from https://github.com/catalpainternational/canoe/pull/127. I'd still like to merge https://github.com/catalpainternational/canoe/pull/127 first in case I missed something.

PR #132, opened 2020-06-12
get image on update as well as mount

PeteCoward:
Fixes a problem with the first image showing up on multiple cards in lesson content.

The problem was that the code to retrieve the image path only ran on mount, and switching between two cards does not re-mount the card.

The fix is to run the code on update as well as on mount.

Investigating this showed up an important thing about the way riot works. 
If you do things in onBeforeMount or onBeforeUpdate ( where i would ideally like to do this kind of stuff ) if those things are async at all, then the actual mount or updates will not wait for them to complete, which can result in infinite updating, hence the guard against re-doing this by checking the imageId state property.

### DIFF
--- a/src/riot/Components/ImageWrapper.riot.html
+++ b/src/riot/Components/ImageWrapper.riot.html
@@ -16,23 +16,28 @@
         export default {
             state: {
                 imagePath: "",
+                imageId: "",
+            },
+            onUpdated(props, state) {
+                this.setImagePath(props, state);
             },
 
-            async onMounted(props, state) {
+            onMounted(props, state) {
+                this.setImagePath(props, state);
+            },
+
+            setImagePath(props) {
                 const { imageId } = props;
-
-                let imagePath = "";
-
-                try {
-                    imagePath = await getImageUrl(imageId);
-                } catch (error) {
-                    if (!(error instanceof MissingImageError)) {
-                        throw error;
-                    }
+                if (this.state.imageId !== imageId) {
+                    getImageUrl(imageId).then(imagePath => {
+                        this.update({imagePath, imageId});
+                    }).catch(error => {
+                        if (!(error instanceof MissingImageError)) {
+                            throw error;
+                        }
+                    });
                 }
-
-                this.update({ imagePath });
-            },
+            }
         };
     </script>
 </ImageWrapper>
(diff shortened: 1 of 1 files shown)


src/riot/Components/ImageWrapper.riot.html
```diff
@@ -16,23 +16,28 @@
...
+                    }).catch(error => {
+                        if (!(error instanceof MissingImageError)) {
+                            throw error;
+                        }
+                    });
                 }
-
-                this.update({ imagePath });
-            },
+            }
```
sandroroux, line 40: **Optional:**
* Unpack the props outside of the method. Pass only what you need.
* Flip the conditional.

**Mandatory:**
* Use `async`.


```js
async setImagePath(imageId) {
    if (this.state.imageId === imageId) {
        return;
    }

    try {
        const imagePath = await getImageUrl(imageId);
        this.update({imagePath, imageId});
    } catch (error) {
        if (!(error instanceof MissingImageError)) {
            throw error;
        }
    }
}
```

sandroroux:
I don't quite understand this:
> If you do things in onBeforeMount or onBeforeUpdate ( where i would ideally like to do this kind of stuff ) if those things are async at all, then the actual mount or updates will not wait for them to complete, which can result in infinite updating, hence the guard against re-doing this by checking the imageId state property.

In general, we should mount a UI and update it as data becomes available. What causes infinite updating?

sandroroux:
Oh, I see. You shouldn't call `update` in `onUpdate` unless you can avoid the infinite loop.

src/riot/Components/ImageWrapper.riot.html
PeteCoward: It makes sense to pass the whole props due to the 2 calling points.

I was trying to avoid the problem I mentioned by removing the `async` keyword and never put it back.
I don't see the benefit of using it here other than for consistency.
It adds more cruft than it removes in this case where you need a try catch anyway and have to sprinkle asyncs and awaits in 2 other places

sandroroux: Consistency is important.

You only need `async` on the `setImagePath` signature. `onUpdated` and `onMounted` can call an `async` update without `await`ing it.

PeteCoward:
I'm not sure about absolutely ruling we always use `async await`. There are situations where promises make for easier reading. I'm personally happy with a mix, we all need to understand both :)

sandroroux:
@PeteCoward, fair enough. Thank you for making the change, in this case.

PR #141, opened 2020-06-17
#116: Front-end uses the Back-end's Resources

sandroroux:
Closes #116.


src/riot/App.riot.html
```diff
@@ -32,7 +31,6 @@
             <Settings if="{state.contentType === 'settings'}"></Settings>
             <Resources
                 if="{state.contentType === 'resources'}"
-                dummyArticles="{articleDummy}"
                 listOfTags="{listOfTags}"
```
katevoss, line 34: I think you want to remove the `listOfTags` attribute here too.

PR #143, opened 2020-06-19
Front-end removes QuickReferences.

sandroroux:
Also, `ResourcesRoot` routes to the `Resources` tag.

The front-end portion of catalpainternational/canoe-backend#45.


src/riot/App.riot.html
```diff
@@ -30,10 +30,10 @@
             ></LessonPage>
             <Settings if="{state.contentType === 'settings'}"></Settings>
             <Resources
-                if="{state.contentType === 'resources'}"
+                if="{state.contentType === 'resources' || state.contentType === 'ResourcesRoot'}"
```
katevoss, line 33: When would the contentType change from `resources` to `ResourcesRoot`?

sandroroux: The ResourcesRoot is a Wagtail Page and has an ID. This `||` handles when you route to the ResourceRoot's page ID. It lands you on the Resources list.

PR #144, opened 2020-06-19
Card back buttons routes to the previous card.

sandroroux:
Closes catalpainternational/asteroid#108.

### DIFF
--- a/src/riot/Lesson/LessonContent.riot.html
+++ b/src/riot/Lesson/LessonContent.riot.html
@@ -70,7 +70,7 @@
         import QuoteCard from "RiotTags/Components/QuoteCard.riot.html";
         import ListCard from "RiotTags/Lesson/ListCard.riot.html";
 
-        import { getLessonCardIdx, getLessonModuleHash } from "js/Routing";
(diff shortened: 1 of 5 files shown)


src/riot/Lesson/LessonContent.riot.html
```diff
@@ -91,7 +91,7 @@
             },
 
             onBeforeMount(props, state) {
-                state.cardIdx = props.isLessonPage ? getLessonCardIdx() : '0';
+                state.cardIdx = props.isLessonPage ? getLessonCardIdx() : "0";
```
PeteCoward, line 94: ugh.. we are storing `cardIdx as a string :)

sandroroux: Indeed. Would you like that to change in this PR?

sandroroux: @PeteCoward I redirected this fix into RC-1.1. This change doesn't exist anymore.

sandroroux:
@PeteCoward @katevoss I directed this bug fix into RC-1.1. Do y'all mind giving it another look?

PeteCoward:
Incidentally I recently learned that we can use history.pushState and the popstate event for forward back history tracked changes which don't affect the url. Might be useful at some point

PR #145, opened 2020-06-19
Adding queryString to URI

katevoss:
~~This draft is partially functional:~~
~~- if you go to a URI like http://localhost:8080/#resources?qs=test, the query "test" will be in the search bar and the resource articles are filtered by that string~~
~~- you can change the query in the search bar, and the resource articles will filter correctly.~~
~~- as you change the query in the search bar, the label and helper text ("enter at least 3 char") will appear/disappear correctly.~~
~~- as you change the query in the search bar, the query string `qs` in the URL automatically updates~~

~~It does not (yet):~~
~~- handle filter by tags~~

Ready for review!

I chose to make all tags lowercase because tags are automatically made lowercase in the URL. It makes for easier adding/removing of url params.

### DIFF
--- a/src/js/Routing.js
+++ b/src/js/Routing.js
@@ -102,7 +102,7 @@ export const parseURLHash = () => {
     return afterTheHash.split("/");
 };
 
-export const parseURLQuery = () => {
+export const getSearchQueryFromUrl = () => {
     const currentHash = parseURLHash();
     const queryString = currentHash[0].split("?")[1];
     return queryString;
--- a/src/js/Routing.js
+++ b/src/js/Routing.js
@@ -102,6 +102,12 @@ export const parseURLHash = () => {
     return afterTheHash.split("/");
 };
 
+export const parseURLQuery = () => {
+    const currentHash = parseURLHash();
+    const queryString = currentHash[0].split("?")[1];
+    return queryString;
+}
+
 export function getLessonCardIdx() {
     return parseInt(location.hash.split("/")[2]) - 1;
 }
--- a/src/js/Routing.js
+++ b/src/js/Routing.js
@@ -93,8 +93,8 @@ export const getPage = async () => {
 };
 
 export const getPreviewQueryFromURL = () => {
-    const querystring = window.location.search.replace(/^\?/, "");
-    return querystring;
+    const queryString = window.location.search.replace(/^\?/, "");
+    return queryString;
 };
 
 export const parseURLHash = () => {
--- a/src/js/SearchAndFiltering.js
+++ b/src/js/SearchAndFiltering.js
@@ -18,7 +18,7 @@ export const doPageTagsMatchSelections = (pageTags, selectedTags) => {
 };
 
 export const doesQueryMatchSearchContent = (query, ...searchableStrings) => {
-    if (!query || query.length < 3) {
+    if (!query) {
         return true;
     }
     const lowercaseQuery = query.toLowerCase();
--- a/src/js/SearchAndFiltering.js
+++ b/src/js/SearchAndFiltering.js
@@ -3,7 +3,7 @@ import { intersection } from "js/SetMethods";
 export const getTagsFromPages = (pagesWithTags) => {
     const eachArticlesTags = pagesWithTags.map((page) => page.tags);
     const setOfTags = new Set(eachArticlesTags.flat());
-    return Array.from(setOfTags);
+    return Array.from(setOfTags).map(tag => tag.toLowerCase());
 };
 
 export const doPageTagsMatchSelections = (pageTags, selectedTags) => {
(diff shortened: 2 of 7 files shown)


src/js/SearchAndFiltering.js
```diff
@@ -18,7 +18,7 @@ export const doPageTagsMatchSelections = (pageTags, selectedTags) => {
 };
 
 export const doesQueryMatchSearchContent = (query, ...searchableStrings) => {
-    if (!query) {
+    if (!query || query.length < 3) {
         return true;
     }
```
sandroroux, line 23: This should move to before (or after) we call `doesQueryMatchSearchContent`. Query length decisions are independent of whether a query matches certain searchable content.

src/riot/Components/SearchBar.riot.html
```diff
@@ -1,33 +1,16 @@
 <SearchBar>
     <div class="form-group">
-        <p class="search-text" if="{state.hasSearchQuery && !state.query}" translate>Enter at least 3 characters</p>
-        <input id="search" type="text" oninput="{onSearchInput}"/>
-        <label class="{state.hasSearchQuery ? 'hide' : ''}" for="search" translate></label>
+        <p class="search-text" if="{props.query.length < 3 && props.query.length > 0}" translate>Enter at least 3 characters</p>
```
sandroroux, line 3: It might make sense to do this `if` check in `onSearchInput`.

When we receive search input, `SearchBar` checks that the query meets certain criteria (longer than three characters) and `update`s its UI. When the query is "good", we submit it to `Resources`.

src/js/Routing.js
```diff
@@ -102,6 +102,12 @@ export const parseURLHash = () => {
     return afterTheHash.split("/");
 };
 
+export const parseURLQuery = () => {
```
sandroroux, line 105: `getSearchQueryFromUrl` might be more direct.

src/riot/Screens/Resources.riot.html
```diff
@@ -51,25 +52,46 @@ <h3 translate>Resources</h3>
...
+                    // and add tag to uri
                 }
                 this.update();
             },
 
             submitQuery(query) {
+                // add/update query in uri
+                const queryStringFromURI = parseURLQuery();
+                const urlParams = new URLSearchParams(queryStringFromURI);
+                const encodedQuery = encodeURIComponent(query);
```
sandroroux, line 86: `submitQuery` receives the `query` from `SearchBar`. Why does `submitQuery` read the URL?

src/riot/Components/SearchBar.riot.html
katevoss: What's the benefit for moving the `if` check? Do you mean that `onSearchInput` would update state (e.g. `isShortQuery`), which would trigger the `Enter at least 3 characters` message?

src/riot/Screens/Resources.riot.html
katevoss: `submitQuery` reads the URL to get the `urlParams`, which are then used to `set` the `query` in the URL with ```urlParams.set("qs", encodedQuery);
                window.history.replaceState({}, '', `${location.pathname}#resources?${urlParams}`);```
Earlier I tried to read the `urlParams` once and set them into state, but that didn't work. Is there another way to do this?

sandroroux:
@katevoss I'm sorry for the delay. To understand the design, I had to play with the code. Tags don't work as I expect:
* I select multiple filters, click into one of the articles that displays, and click the browser back button. In fact, the back button/browser history and filters misbehave.
* Search works great— with the back button too.
* Search, tags, and the back button malfunction.

I'll let you fix tags before I comment too much. Let me know if I can help.

katevoss:
@sandroroux the back button and filters are fixed in https://github.com/catalpainternational/canoe/pull/145/commits/ff52f0e819e3e853bd7e883b3893dad74630c7f0

sandroroux:
@katevoss my filter behavior varies wildly, but it might be the fault of my devstack. While I try to figure out what's going on, here are a couple things I noticed. Can you tell me if you get the same behavior?

## Filters disappear from the URL
### Setup:
I create filters A, B, C, and D so that the order of the filters on the Resources page is “a, b, c, d”.

### Reproduction Steps
Select A, B, C, D, in that order. Click into an article, hit the browser's back button. Deselect B.

The URL changes from `#resources?filter=a&filter=b&filter=c&filter=d` to `http://localhost:8080/#resources?filter=d`.

From here, I deselect D, the URL goes from `#resources?filter=d` to `#resources?filter=c`, though the UI shows that A and C are selected.

If I refresh the page, the UI shows C is selected (meaning the code pulls from the URL correctly).

katevoss:
Ready for review again @sandroroux @PeteCoward

PR #148, opened 2020-06-23
Quote alignment fix

katevoss:
Closes https://github.com/catalpainternational/asteroid/issues/97


katevoss:
@sandroroux @pedro16v The quotation marks have to be styled separately to change their size and appearance as-is. Barring spending more hours on that final quotation mark, it's not feasible to move it to after the end of the quote.

pedro16v:
Yes, I did hack a bit of CSS to make it look like that 😁 If it's not a task that takes minutes then simply leave it as is.

PR #149, opened 2020-06-24
Help button on home page

katevoss:
Closes https://github.com/catalpainternational/asteroid/issues/130


katevoss:
Converted to draft while I wait for the finalized design.

src/riot/Components/ContactInformation.riot.html
```diff
@@ -0,0 +1,28 @@
+<ContactInformation>
+    <div class="custom-contact-container">
+        <p>Send us a message and we'll help you log in</p>
+        <button class="btn-primary" onclick="{sendWhatsAppMessage}">Send WhatsApp</button>
+        <p class="sm-note">Don't have WhatsApp?</p>
+        <button class="btn-secondary" onclick="{sendSmsMessage}">Send SMS</button>
+    </div>
+
+    <script>
+        const phoneNumber = "19787602338";
```
sandroroux, line 10: If you're interested, let's configure the phone number in `canoe-environment.js`. It's a quick change; let me know if you want to talk about it.

sandroroux: Seeing as I will layer in my guest login changes, let me know if I should tackle this.

sandroroux:
@katevoss Just needs a file conflict resolution.

src/riot/Components/ContactInformation.riot.html
katevoss: I put the phone number and the button actions into a separate riot tag `ContactInformation` so that it can be overwritten differently by apps.

katevoss: I added the Asteroid-specific tag here https://github.com/catalpainternational/asteroid/pull/153

sandroroux: `canoe-environment.js` 's advantage is that we can configure the phone number without needing a PR. I like the idea of swapping front-end components, but imagine that swapping a component to set the phone number is more than we need.

Unless you and @PeteCoward already agreed to this, or I missed the point.

katevoss: In Peter's recently-merged branch [`better_project_overriding`](https://github.com/catalpainternational/canoe/pull/156), we can now override whole riot tags. The benefit of swapping the whole tag rather than just the phone number is that it gives the app (Olgeta, Asteroid/Haroman) the flexibility to decide what the CTA buttons do (call/text/email), and their destination (phone number, email). Asteroid/Haroman can send a WhatsApp message or an SMS, and Olgeta can have a "click to email" help button.

PR #152, opened 2020-06-30
Header changes

sergioestevess:
- Consistent _Home_, _Notification_ and _Profile_ pages header.

### DIFF
--- a/src/scss/_mixins.scss
+++ b/src/scss/_mixins.scss
@@ -1,8 +1,7 @@
 @mixin screen-banner-styles {
     margin-top: -65px;
-    padding-top: 65px;
     height: 30vh;
-    padding: 0 1em;
+    padding: 65px 1em 0 1em;
 
     .header-box {
         display: flex;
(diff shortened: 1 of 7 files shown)


src/scss/_mixins.scss
```diff
@@ -1,13 +1,31 @@
 @mixin screen-banner-styles {
-    flex-direction: row;
-    justify-content: space-evenly;
-    align-items: center;
-    display: flex;
+    margin-top: -65px;
```
sandroroux, line 2: @sergioestevess We got away from negative top margins. It affects the styles of everything under the top menu. `TopMenu` is aware of what appears below it. `TopMenu` can set its own color, removing the need for pulling the whole screen underneath.

sergioestevess: I'm happy to remove them, but we have to fix the TopMenu on wider screens:

image

sandroroux: Oops, I had submitted a fix for this. Looks like I might have missed something.

I like your changes though. Let me think on how we can combine approaches.

PR #161, opened 2020-07-08
#156: Connects Exam front- and back-end

sandroroux:
Closes https://github.com/catalpainternational/asteroid/issues/156.

Runs against https://github.com/catalpainternational/canoe-backend/pull/49 and https://github.com/catalpainternational/asteroid/pull/147.

### DIFF
--- a/src/riot/WagtailPages/CoursePage.riot.html
+++ b/src/riot/WagtailPages/CoursePage.riot.html
@@ -26,37 +26,41 @@
                 isComingSoon="{lesson.coming_soon}"
                 showFooter="{true}"
             ></Card>
-            <h5 class="section-title" translate>Course exam</h5>
-            <Card
-                title="{'Exam for Current Course'}"
-                description="{'this is the course exam description'}"
-                duration="{15}"
-                link="{props.wagtailPage.id}/exam/1"
-                isCompleted="{false}"
-                isComingSoon="{false}"
-                showFooter="{true}"
-                isExam="{true}"
-            ></Card>
+            <template if="{state.examCards.length > 0}">
+                <h5 class="section-title" translate>Course exam</h5>
+                <Card
+                    title="{`${props.wagtailPage.title}'s Course Exam`}"
+                    description="{'this is the course exam description'}"
+                    duration="{15}"
+                    link="{props.wagtailPage.id}/exam/1"
+                    isCompleted="{false}"
+                    isComingSoon="{false}"
+                    showFooter="{true}"
+                    isExam="{true}"
+                ></Card>
+            </template>
         </div>
     </div>
     <CourseExam
-        if="{state.hash === examModule}"
+        if="{state.hash === EXAM_MODULE && state.examCards.length > 0}"
         wagtailPage="{props.wagtailPage}"
         module="exam"
-        cards="{examCards}"
+        cards="{state.examCards}"
     ></CourseExam>
 
     <script>
+        import Card from "RiotTags/Components/Card.riot.html";
+        import CourseExam from "RiotTags/Lesson/CourseExam.riot.html";
+
         import { countCompleteLessonsInCourse } from "js/LearningStatistics";
         import { isComplete } from "Actions/completion";
         import { getCourseAndLessonSlugs } from "js/utilities";
-        import { getLessonModuleHash, getHashPieces } from "js/Routing";
-        import Card from "RiotTags/Components/Card.riot.html";
-        import CourseExam from "RiotTags/Lesson/CourseExam.riot.html";
+        import { getHashPieces } from "js/Routing";
 
         export default {
             state: {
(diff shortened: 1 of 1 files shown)


src/riot/WagtailPages/CoursePage.riot.html
```diff
@@ -26,84 +26,63 @@ <h5 class="section-title" translate>Your lessons</h5>
...
+                    isComingSoon="{false}"
+                    showFooter="{true}"
+                    isExam="{true}"
+                ></Card>
+            </template>
         </div>
     </div>
     <CourseExam
-        if="{state.hash === examModule}"
+        if="{state.hash === EXAM_MODULE && state.examCards.length > 0}"
```
katevoss, line 45: Why did you add the conditional `state.examCards.length > 0`?

You might want to add this conditional to the "Take exam" button on `<LessonComplete>` when a user is directed to take the Course Exam. That way the button only appears if there is an exam.

src/riot/WagtailPages/CoursePage.riot.html
```diff
@@ -26,84 +26,63 @@ <h5 class="section-title" translate>Your lessons</h5>
...
-                    answers: [
-                        { correct: false, text: "ba dum dum dum" },
-                        { correct: true, text: "dee ba da" },
-                        { correct: false, text: "ba dum dum dum" },
-                        { correct: true, text: "dee ba da" },
-                    ],
-                    question: "What does Karen sing?",
-                },
-            ],
+            EXAM_MODULE: "exam",
```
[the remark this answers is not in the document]
PeteCoward: @sandroroux loves a nitpick :)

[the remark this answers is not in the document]
sandroroux: @katevoss, that's a good point. I think I messed around with a conditional, which is why I changed the variable name.

As we discussed, the codebase lacks an agreed-upon style guide. In general, I try to write variables with unchanging values in uppercase.

Uppercase constants are [typical](https://stackoverflow.com/questions/34732630/all-caps-constants-in-javascript-and-requireds-and-imports) and, I think, add meaning. I like [this guy's take](https://www.freecodecamp.org/news/when-to-capitalize-your-javascript-constants-4fabc0a4a4c4/) too because he suggests that uppercase constants are optional, like all practices.

I'd rather the team agree on an approach. I'm for consistency between `CourseExam` and `LessonPage`, and will revert if you think the other approach is better.

src/riot/WagtailPages/CoursePage.riot.html
sandroroux: Someone could navigate to the URL when there are no exam cards, and errors would ensue. I should rethink this. The problem probably has a better solution.

Thanks for the heads-up on `LessonComplete`. I'll take a look.

src/riot/WagtailPages/CoursePage.riot.html
```diff
@@ -26,84 +26,63 @@ <h5 class="section-title" translate>Your lessons</h5>
...
-                isCompleted="{false}"
-                isComingSoon="{false}"
-                showFooter="{true}"
-                isExam="{true}"
-            ></Card>
+            <template if="{state.examCards.length > 0}">
+                <h5 class="section-title" translate>Course exam</h5>
+                <Card
+                    title="{`${props.wagtailPage.title}'s Course Exam`}"
+                    description="{'this is the course exam description'}"
```
katevoss, line 33:
```suggestion
```
There isn't (currently) a description of the course exam in the backend so let's remove this altogether.

src/riot/WagtailPages/CoursePage.riot.html
```diff
@@ -26,84 +26,63 @@ <h5 class="section-title" translate>Your lessons</h5>
...
-                link="{props.wagtailPage.id}/exam/1"
-                isCompleted="{false}"
-                isComingSoon="{false}"
-                showFooter="{true}"
-                isExam="{true}"
-            ></Card>
+            <template if="{state.examCards.length > 0}">
+                <h5 class="section-title" translate>Course exam</h5>
+                <Card
+                    title="{`${props.wagtailPage.title}'s Course Exam`}"
```
katevoss, line 32: Adding the "[CourseTitle]'s Course Exam" feels a little redundant as there is also a section title "Course Exam" right above the card. 
image

I would recommend removing it from the title, but I'm not too bothered so it's not a required change. Do you have any thoughts?

src/riot/WagtailPages/CoursePage.riot.html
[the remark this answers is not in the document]
katevoss: Consistency is my number 1 priority and second to that I think writing variables that represent constants as UPPERCASE is a good approach. If you could make the relevant changes in `LessonPage` too then I'm on board!

Issue #160, opened 2020-07-08
Home Screen Courses and tagging


PeteCoward:
From the design here:
https://projects.invisionapp.com/share/HTXX33SWNQD#/screens/424363278

- [x] course card simplified
<img width="235" alt="Screenshot 2020-07-08 at 11 53 52" src="https://user-images.githubusercontent.com/1092029/86869559-bf852400-c111-11ea-8166-188f3127972c.png">

- [x] completed courses are always at the bottom if the list, and have completion green tick ui
<img width="263" alt="Screenshot 2020-07-08 at 11 54 46" src="https://user-images.githubusercontent.com/1092029/86869616-d7f53e80-c111-11ea-94bb-010c16b22b9b.png">

- [x] tags occupy a single line, and are scrollable with a swipe
<img width="283" alt="Screenshot 2020-07-08 at 11 55 49" src="https://user-images.githubusercontent.com/1092029/86869702-fce9b180-c111-11ea-872a-b77662c2ce77.png">

PR #163, opened 2020-07-09
Home screen courses and tagging restyle

katevoss:
Closes https://github.com/catalpainternational/canoe/issues/160

### DIFF
--- a/src/riot/WagtailPages/HomePage.riot.html
+++ b/src/riot/WagtailPages/HomePage.riot.html
@@ -25,32 +25,17 @@
         ></TagList>
         <div
             class="course"
-            each="{course in props.wagtailPage.courses}"
+            each="{course in state.sortedListOfCourses}"
             if="{isCourseVisible(course.tags)}"
         >
-            <div class="course-header">
-                <div>
-                    <a href="#{course.data.id}">
-                        <h3>{course.data.title}</h3>
-                    </a>
-                    <div>
-                        <p>{ courseProgressMessage(course) }</p>
-                        <a href="#{course.data.id}">
-                            <p class="text-purple" translate>view all</p>
-                            <span class="arrow background-purple"></span>
-                        </a>
-                    </div>
-                </div>
-            </div>
             <Card
-                if="{state.nextLesson[course.data.id]}"
-                title="{state.nextLesson[course.data.id].title}"
-                description="{state.nextLesson[course.data.id].description}"
-                duration="{state.nextLesson[course.data.id].duration}"
-                link="{state.nextLesson[course.data.id].id}"
-                isCompleted="{isLessonComplete(course, state.nextLesson[course.data.id])}"
+                title="{course.data.title}"
+                link="{course.data.id}"
+                description="{courseProgressMessage(course)}"
+                isCompleted="{isCourseComplete(course)}"
                 isComingSoon="{state.nextLesson[course.data.id].coming_soon}"
                 showFooter="{true}"
+                type="course"
             ></Card>
         </div>
     </div>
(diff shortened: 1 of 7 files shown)


src/riot/WagtailPages/HomePage.riot.html
```diff
@@ -127,6 +134,18 @@ <h3>{course.data.title}</h3>
                 );
             },
 
+            isCourseComplete(course) {
```
katevoss, line 137: I noticed that there is a `isCourseComplete` function in `LessonComplete` and I started to factor it out into the `completion.js` but I got into a big spaghetti mess when I found that there's two definitions of `countCompleteLessonsInCourse`, one in `LearningStatistics.js` and one in `completion.js`. 

Which is all to say, if you have any insight into why we have two of the same function that would be very helpful!

sandroroux: Yup. `LearningStatistics` is meant to be an interface between the completion module, which is stand alone, and `canoe` (`LearningStatistics` is an attempt at dependency inversion). I wanted to limit `completion.js` hard dependencies between `canoe` and `completions.js`.

I meant to move towards more coherence between `completions.js` and `canoe`, but, in this case, I'm afraid I did the opposite. I'm sorry about that.

Can we discuss your refactor? I think I could help untangle the weirdness between completions and canoe.

src/riot/WagtailPages/HomePage.riot.html
```diff
@@ -96,6 +82,27 @@ <h3>{course.data.title}</h3>
...
+                courses.map((course) => {
+                    if (this.isCourseComplete(course)) {
+                        completedCourses.unshift(course);
+                    } else {
+                        incompleteCourses.unshift(course);
+                    }
+                });
+
+                incompleteCourses.reverse();
+                completedCourses.reverse();
```
sandroroux, line 102: If, instead of `unshift`ing and `reversing` the courses, you just `push`ed them, would you get the same result?

src/riot/WagtailPages/HomePage.riot.html
```diff
@@ -127,6 +134,18 @@ <h3>{course.data.title}</h3>
...
+                const { courseSlug, lessonSlugs } = getCourseAndLessonSlugs(course);
+                const numberOfCompleteLessons = countCompleteLessonsInCourse(
+                    courseSlug,
+                    lessonSlugs
+                );
+
+                return (
+                    countCompleteLessonsInCourse(courseSlug, lessonSlugs) === course.lessons.length
+                );
+            },
```
sandroroux, line 147: `LearningStatistics.js` has a `isTheCourseInProgress` method, which this `isTheCourseComplete` can use.`isTheCourseComplete` can be:
```js
isTheCourseComplete(course) {
    return !isTheCourseInProgress(course);
}
```

src/scss/_resources.scss
```diff
@@ -57,8 +61,22 @@ taglist {
...
 
     .tag-container {
-        width: 100%;
-        margin-bottom: 1.1em;
+        overflow: auto;
+        white-space: nowrap;
+        margin-bottom: 0.5em;
+    }
+
+    .tag-container::after {
```
sandroroux, line 69: You may want to repeat this for `::before`. Otherwise, you get a blurry edge on the right:
image

and a straight edge on the left:
image

src/scss/_resources.scss
```diff
@@ -106,19 +124,19 @@ searchbar {
...
+        content: "";
     }
 
     label {
         flex: none;
         font-size: 0.8em;
 
         &:before {
-            content: 'Search';
+            content: "Search";
```
sandroroux, line 135: I just realized, if this text is visible, it won't be translated. We can address this separately.

PR #165, opened 2020-07-13
#129, Part 1: In-Memory Grading

sandroroux:
Contributes to catalpainternational/asteroid#129.

### Overview
This branch implements in-memory grading, the first step in the process of storing exam responses as `UserAction`s:

_**Step 1:** Course Exams compute a grade; finishing the exam completes the exam card. The exam answers exist for as long as the page doesn't refresh._
**Step 2:** Course Exams store answers in IndexedDB, persisting answers across refreshes.
**Step 3:** Course Exams store answers on the server, as `UserAction`s. This synchs answers across a user's multiple devices.

### To Do
* Update completion methods so that a course is complete only after its exam is complete. (@PeteCoward, I'd like to handle this in another PR)

### DIFF
--- a/src/riot/Lesson/ExamResults.riot.html
+++ b/src/riot/Lesson/ExamResults.riot.html
@@ -1,0 +1,96 @@
+<ExamResults>
+    <template if="{state.passedExam !== null}">
+        <LessonFrame>
+            <div class="{!state.passedExam ? '' : 'confetti background'}"></div>
+            <div class="congratulations-container">
+                <div class="{!state.passedExam ? 'incorrect-flag' : 'report-card overlay'}"></div>
+                <template if="{state.passedExam}">
+                    <div>
+                        <h2 translate>Amazing work!</h2>
+                        <h3 translate>You've completed the course</h3>
+                        <p>{ passedExamMessage() }</p>
+                    </div>
+                    <button class="btn-primary" onclick="{goHome}" translate>
+                        Back to Courses
+                    </button>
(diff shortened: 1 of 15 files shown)


src/riot/Lesson/ExamResults.riot.html
```diff
@@ -0,0 +1,96 @@
+<ExamResults>
```
sandroroux, line 1: @PeteCoward, If we decide every course must have an exam, we can delete much code in `LessonComplete.riot.html`.

src/riot/WagtailPages/LessonPage.riot.html
```diff
@@ -118,17 +118,18 @@ <h5 class="section-title" translate>Lesson Activities</h5>
...
             components: {
                 lessonobjective: LessonObjective,
                 lessoncontent: LessonContent,
                 lessontest: LessonTest,
                 card: Card,
             },
 
+            OBJECTIVES: "objectives",
+            CONTENT: "content",
+            TEST: "test",
```
sandroroux, line 130: @katevoss asked me to do this [here](https://github.com/catalpainternational/canoe/pull/161#discussion_r452449381).

sandroroux:
@PeteCoward, I'd like to more clearly define the boundary between Canoe and Actions.

Right now, `LearningStatistics` is that boundary. It's supposed to be a dependency-inverting interface, but its name and the grab-bag-ness of its functions muddy that impression.

PR #181, opened 2020-07-17
Lesson follows Course's screen structure; the structure fixes their sticky header.

sandroroux:


PeteCoward:
This seems a bit big and risky. I'm not sure exactly what to look for ( what does it do badly before, and better after? )

PR #189, opened 2020-07-20
#129, Step 2: IDB Saves Exam Answers

sandroroux:
Contributes to https://github.com/catalpainternational/asteroid/issues/129. Closes https://github.com/catalpainternational/asteroid/issues/167.

### Overview
This branch wires grading and the local database.  Answers and grades survive refreshes.

**Step 1:** Course Exams compute a grade; finishing the exam completes the exam card. The exam answers exist for as long as the page doesn't refresh.
_**Step 2:** Course Exams store answers in IndexedDB, persisting answers across refreshes._
**Step 3:** Course Exams store answers on the server, as `UserAction`s. This synchs answers across a user's multiple devices.


src/js/actions/actions_store.js
```diff
@@ -14,30 +14,49 @@ import { postAction, getActions } from "./actions_api";
...
-        });
+    try {
+        const isActionSynched = await postAction(action);
+        // if response is ok set the record to known in idb
+        if (isActionSynched) {
+            markActionAsSynced(action);
+        }
+    } catch (err) {
+        console.error(err);
+    }
```
PeteCoward, line 42: do these errors not get logged to the console anyway?

src/js/actions/Exams.js
```diff
@@ -23,12 +34,37 @@ const assertAnswerShape = (answer) => {
...
+export const pullExamAnswersIntoMemory = async () => {
+    try {
+        const idbAnswers = await getExamAnswersFromIdb();
+        for (const idbAnswer of idbAnswers) {
+            const { questionId, answers, isCorrect } = idbAnswer;
+            storeExamAnswerInMemory(questionId, { answers, isCorrect });
+        }
+    } catch (err) {
+        console.error(`Couldn't load exam response ${questionid}`);
+    }
```
PeteCoward, line 54: Why the catch if its only purpose is to console log?

src/riot/Lesson/TestSingleAnswer.riot.html
```diff
@@ -1,8 +1,8 @@
 <TestSingleAnswer class="answer">
     <ol>
         <li
-            class="{answer === state.selectedAnswer && howToStyleThe(answer)}"
             each="{answer in props.card.answers}"
+            class="{answer === state.selectedAnswer && howToStyleThe(answer)}"
```
PeteCoward, line 5: I am not a fan of `howToStyleThe(answer)` coding style. I cannot sensibly justify my objection, it just offends me.

PeteCoward:
Looks like going in the right direction, does the sync to the server "just work" or do we need to generalise that part too?

sandroroux:
```python
In [7]: UserAction.objects.filter(user=a, data__course='haroman-app-launch-module', data__section="exam
   ...: ")                                                                                             
Out[7]: <QuerySet [<UserAction: sandro: 6aea467f-1fe7-4a08-8acd-071ef7b4d3da on 2020-07-22T15:56:07.982000+00:00>]>

In [8]: b = _                                                                                          

In [9]: b                                                                                              
Out[9]: <QuerySet [<UserAction: sandro: 6aea467f-1fe7-4a08-8acd-071ef7b4d3da on 2020-07-22T15:56:07.982000+00:00>]>

In [10]: b[0].__dict__                                                                                 
Out[10]: 
{'_state': <django.db.models.base.ModelState at 0x1059d3cc0>,
 'id': 12351,
 'user_id': 1,
 'action_uuid': UUID('6aea467f-1fe7-4a08-8acd-071ef7b4d3da'),
 'created': datetime.datetime(2020, 7, 22, 15, 56, 7, 982000, tzinfo=<UTC>),
 'data': {'type': 'completion',
  'course': 'haroman-app-launch-module',
  'lesson': 'exam',
  'synced': 0,
  'section': 'exam',
  'finalScore': 80}}
```

src/js/actions/actions_store.js
sandroroux: They do. I didn't think and preserved the original code's logic. I'll let the errors surface normally.

src/js/actions/Exams.js
sandroroux: I'll fix this.

src/riot/Lesson/TestSingleAnswer.riot.html
```diff
@@ -26,13 +26,37 @@
...
+                });
+            },
+
+            handleRiotMemoryLeaks() {
+                // This method fixes a riot bug(?) that occurs when two
+                // TestSingleAnswers are back to back. If question one
+                // has a selected answer that matches an answer in question two,
+                // question two will automatically select that answer. We don't
+                // want that, so this method clears the selected answer between
+                // questions.
```
PeteCoward, line 54: I think this 'leak' is because the state is not re-initialized on update.
When moving between two TestSingleAnswers the question changes, but the state is not reset.
I'd call it a feature not a bug :)

sandroroux: That's right. I'm unsure how to do what I want. The TestSingleAnswer `props` change during the question change. If new parameters are being passed into a tag, shouldn't the tag re-initilialize? It's like passing something into a function and the function running with old values.

src/js/actions/exam.js
```diff
@@ -0,0 +1,136 @@
...
+};
+
+export const clearInMemoryExamScores = () => {
+    finalScores.clear();
+};
+
+export const pullExamAnswersIntoMemory = async () => {
+    const idbAnswers = await getExamAnswersFromIdb();
+    for (const idbAnswer of idbAnswers) {
+        const { questionId, answers, isCorrect } = idbAnswer;
```
PeteCoward, line 63: I'm very confused by an answer having answers!

src/riot/Lesson/TestSingleAnswer.riot.html
PeteCoward: a tag is not a function, if we change a property on it then it is our responsibility to update it's state accordingly

src/js/actions/exam.js
PeteCoward: ahh, i see checkboxes answers can have multiple answers!

sandroroux: That's right. The IDB object's `answers` is an array of answers; each answer, an object. I think I `assert` this array requirement, so that all question types return answers in the correct form.

PR #191, opened 2020-07-21
Resources use new i18n approach.

sandroroux:


sandroroux:
@PeteCoward Should we do the `TRANSLATIONS` pattern everywhere?

PeteCoward:
Well, all translations MUST work.
Ideally yes, we should probably lose the html `translate` pattern as it's been buggy, and I think the fact is does not support substitution just overcomplicates things. But making that change is not a priority.

PR #196, opened 2020-07-22
Add highest score to exam card

katevoss:
Closes https://github.com/catalpainternational/asteroid/issues/185


src/riot/WagtailPages/CoursePage.riot.html
```diff
@@ -53,15 +54,17 @@ <h5 class="section-title">{ TRANSLATIONS.course_exam() }</h5>
...
+                your_lessons: () => gettext("Your lessons"),
+                course: () => gettext("Course"),
+                course_exam: () => gettext("Course exam"),
             },
             components: {
                 card: Card,
                 courseexam: CourseExam,
             },
 
+            highestScore: "85%",
```
PeteCoward, line 66: Will leave this one open and we'll chat to @sandroroux about getting this value from the stored data

sandroroux:
@PeteCoward `CoursePage` now uses IDB data to calculate the highest score.

Some notes about my changes:

* I store exam scores in a separate `Map`, exactly how I store exam answers. Like answers, these in-memory scores don't have a back-end endpoint. That said, for now, the back-end receives final scores from our finished-exam completion. I didn't use completion data to calculate the high score because I didn't want to mangle `completion.js`.
* Eventually, I think we should upload the in-memory scores and remove the completion-packaged score.
* I split `LearningStatistics` into `CompletionInterface` and `ExamInterface`. The `LearningStatistics` name didn't make sense and concealed its purpose: to invert the dependency between Canoe and Actions. This inversion's original reasoning (I did it way back) was to increase Canoe's modularity. "Don't depend on implementations, depend on interfaces"— that kind of thing. Either way, we must decide whether this inversion is too much trouble. For now, the names state the code situation better than `LearningStatistics`.

PR #200, opened 2020-07-27
Add guest banner to home

katevoss:
Closes https://github.com/catalpainternational/asteroid/issues/174


src/riot/App.riot.html
```diff
@@ -136,6 +136,10 @@
                     // If user returns to site and is logged in.
                     this.route();
                 }
+
+                if (isGuestUser()) {
+                    toggleGuestBanner(true);
+                }
```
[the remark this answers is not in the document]
katevoss: I forgot to move this back to HomePage, thanks for catching it

src/riot/WagtailPages/HomePage.riot.html
```diff
@@ -7,44 +8,28 @@ <h3>{ greetingMessage() }</h3>
...
-                description="{courseProgressMessage(course)}"
-                isCompleted="{!isCourseInProgress(course)}"
-                showFooter="{true}"
-                type="course"
-            ></Card>
+        <TagList if="{state.listOfTags.length > 0}" tags="{state.listOfTags}" onTagClick="{onTagClick}"
+            selectedTags="{state.selectedTags}"></TagList>
+        <div class="course" each="{course in state.sortedListOfCourses}" if="{isCourseVisible(course.tags)}">
+            <Card title="{course.data.title}" link="{course.data.id}" description="{courseProgressMessage(course)}"
+                isCompleted="{!isCourseInProgress(course)}" showFooter="{true}" type="course"></Card>
```
sandroroux, line 22: I don't see any code changes here. Did Prettier reformat the file?

My Prettier breaks lines at 100 characters and reverts this change to the earlier formatting.

src/js/redux/Store.js
```diff
@@ -31,6 +32,7 @@ const getInitialLanguage = () => {
 
 const initialStoreState = {
     language: getInitialLanguage(),
+    isGuestBannerVisible: true,
```
sandroroux, line 35: Rather than setting an initial value here, I'd set the default value in the Reducer:
```js
const toggleGuestBannerVisibility = (state = true, action) => {
```

While you can initialize state here, I think people set initial state with values you can't or shouldn't set as the Reducer default.

sandroroux: Just spitballing here: I noticed if you close the guest banner, the banner reappears as soon as you refresh the page.

You could store `isGuestBannerVisible` in local storage and hydrate this state with local storage's value. We did similarly with the language toggle.

src/riot/WagtailPages/HomePage.riot.html
```diff
@@ -106,7 +95,7 @@ <h6 if="{state.currentCourse && isCourseInProgress(state.currentCourse)}">
                     "You only have %1 lesson left to finish %2.",
                     "You only have %1 lessons left to finish %2.",
                     this.state.currentCourse.data.lessons_count -
-                        countCompleteLessonsInCourse(courseSlug, lessonSlugs),
+                    countCompleteLessonsInCourse(courseSlug, lessonSlugs),
```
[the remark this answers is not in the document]
katevoss: Yes, I'm not sure why it did though.

src/riot/WagtailPages/HomePage.riot.html
katevoss: You're correct, just my prettier settings deciding to reformat. I'll switch my code break point which should fix this.

PR #201, opened 2020-07-27
Add translation for Search bar

katevoss:
Closes https://github.com/catalpainternational/asteroid/issues/159

I'm not in favor of putting translations into CSS but this was a quick fix barring a redesign. Let me know if you have other ideas on how to fix it.

### DIFF
--- a/src/riot/Components/SearchBar.riot.html
+++ b/src/riot/Components/SearchBar.riot.html
@@ -1,6 +1,6 @@
 <SearchBar>
     <div class="form-group">
-        <input id="search" type="text" value="{props.searchString}"
+        <input id="search" type="text" value="{props.searchString}" placeholder="{ TRANSLATIONS.search() }"
             oninput="{debounce(onSearchInput, SUBMIT_AFTER_THIS_MANY_MS)}" />
         <label class="{getLanguage()} {props.searchString ? 'hide' : ''}" for="search"></label>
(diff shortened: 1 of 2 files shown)


PeteCoward:
I too am not a fan, let me sleep on it

src/riot/Components/SearchBar.riot.html
```diff
@@ -1,17 +1,14 @@
 <SearchBar>
     <div class="form-group">
-        <input
-            id="search"
-            type="text"
-            value="{props.searchString}"
-            oninput="{debounce(onSearchInput, SUBMIT_AFTER_THIS_MANY_MS)}"
-        />
-        <label class="{props.searchString ? 'hide' : ''}" for="search"></label>
+        <input id="search" type="text" value="{props.searchString}"
```
sandroroux, line 3: @katevoss I think you can use `placeholder` to plug in the empty-state text:
```html
<input id="search" type="text" placeholder="Search...">
```

katevoss: `placeholder` looks good, the only issue is that the text doesn't disappear until I start typing, not on focus.
image

sandroroux: I think it's fine because we're relying on HTML's built-in and (in theory) widespread behavior. Let me know if you disagree.

PR #208, opened 2020-08-04
Add loading animation when waiting for results

katevoss:
Closes https://github.com/catalpainternational/asteroid/issues/140

### DIFF
--- a/src/riot/App.riot.html
+++ b/src/riot/App.riot.html
@@ -16,6 +16,7 @@
                 contentType="{state.contentType}"
                 wagtailPage="{state.contentType === 'LessonPage' ? state.wagtailPage : ''}"
             ></TopMenu>
+            <LoadingDots if="{state.isLoading}" extrastyleclasses="light"></LoadingDots>
             <HomePage
                 if="{state.contentType === 'HomePage'}"
                 wagtailPage="{state.wagtailPage}"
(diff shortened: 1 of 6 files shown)


src/riot/App.riot.html
```diff
@@ -88,6 +90,7 @@
                 wagtailPage: {},
                 homePageURL: "",
                 appTitle: "",
+                isLoading: false,
```
sandroroux, line 93: Instead of using a flag to show the loading dots, you can display them using existing state. On route, set `wagtailPage` null or as an empty object. Mount `LoadingDots` when `wagtailPage` is in that null/empty state.

src/riot/Screens/Resources.riot.html
```diff
@@ -24,8 +24,7 @@ <h3>{ TRANSLATIONS.resources() }</h3>
                 showFooter="{false}"
             ></Card>
         </div>
-        <p if="{state.isLoading}">{ TRANSLATIONS.loading() }</p>
-        <p></p>
+        <LoadingDots if="{state.isLoading}" extrastyleclasses="dark"></LoadingDots>
```
sandroroux, line 27: We should consider a similar approach here—using the data state to mount `LoadingDots` rather than using an `isLoading` flag. Is it possible?

katevoss: This case is a bit different from loading the app. When state has mounted, there may not be any articles so I think it is still necessary to have a `isLoading` flag to distinguish between when the page is loading and when it is loaded but has no resource articles.

PR #153, opened 2020-06-30
Yarn vs node-gyp

sergioestevess:
- `yarn` successfully runs on Mac and `node-gyp` is not complaining anymore.
- Side effect - updates a bunch of other dependencies.

PR #210, opened 2020-08-05
yarn upgrade

PeteCoward:
This is the result of a `yarn upgrade-interactive`, tested under 
- node v12.18.3
- yarn 1.22.4

should replace #153, #203 

Will be tested locally and on the server over the next week


PeteCoward:
closing this in favour of #213

PR #214, opened 2020-08-06
Exams API adds hasUserTriedExam.

sandroroux:
Closes catalpainternational/asteroid#200.

### DIFF
--- a/src/js/ExamInterface.js
+++ b/src/js/ExamInterface.js
@@ -8,6 +8,7 @@ import {
     tallyExamScore as tallyScore,
     saveExamScore as saveScore,
     getExamHighScore as getHighScore,
+    hasUserTriedExam as hasTriedExam,
 } from "Actions/exam";
 
 export const saveExamAnswer = (questionId, answer) => {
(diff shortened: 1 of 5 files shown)


src/js/ExamInterface.js
```diff
@@ -8,6 +8,7 @@ import {
     tallyExamScore as tallyScore,
     saveExamScore as saveScore,
     getExamHighScore as getHighScore,
+    hasUserTriedExam as hasTriedExam,
```
katevoss, line 11: Why do these imports have aliases?

PeteCoward:
I can't repro the bug on this branch, which is good, but when a guest I get no highest score shown after completion, am investigating whether this is the case in RC-1.2

PeteCoward:
Problems are all with guest users

On RC-1.2 
As a user:
complete an exam with X%, see you have X% score, refresh, you see X% highest score GOOD
As a guest:
complete an exam with 100%, see you have 100% score, refresh, you see 0% highest score BAD

On this branch:
As a user:
complete an exam with X%, see you have X% score, refresh, you see X% highest score GOOD
As a guest:
complete an exam with 100%, see you have 100% score, refresh, you see no highest score at all BAD

src/js/ExamInterface.js
sandroroux: Good question. The short answer is because `ExamInterface.js` redeclares the same names. To avoid a clash between `Exams.js`'s methods and this interface's methods, I alias `Exams.js`'s methods.

The long answer is:

sandroroux:
Thanks for the repro steps, @PeteCoward. I'll look into it.

Issue #74, opened 2020-05-01
Rejecting Notification Permission causes console error


PeteCoward:
Low priority, but rejecting notifications permission throws a console error, might be nice to catch and suppress that

katevoss:
There shouldn't be any visible change so should this issue be moved to Done @pedro16v ?

PR #215, opened 2020-08-06
Catch notification error and surpress it

katevoss:
Closes https://github.com/catalpainternational/canoe/issues/74

### DIFF
--- a/src/js/Notifications.js
+++ b/src/js/Notifications.js
@@ -62,11 +62,10 @@ const postNotificationSubscription = async (subscriptionData) => {
 };
 
 const pushManagerSubscribesToNotifications = async (registration) => {
-    notificationSubscription = await registration.pushManager.subscribe({
+    const notificationSubscription = await registration.pushManager.subscribe({
         userVisibleOnly: true,
         applicationServerKey: urlBase64ToUint8Array(APPLICATION_SERVER_KEY),
     });
-
     return notificationSubscription;
 };
 
@@ -128,7 +127,7 @@ export const subscribeToNotifications = async () => {
     const swRegistration = await navigator.serviceWorker.ready;
     const subscription = await turnOnNotifications(swRegistration);
     if (!subscription) {
-        return false;
+        return;
     }
     storeRegistrationId(subscription.registration_id);
 };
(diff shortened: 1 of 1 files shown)


src/js/Notifications.js
```diff
@@ -62,10 +62,16 @@ const postNotificationSubscription = async (subscriptionData) => {
...
-        applicationServerKey: urlBase64ToUint8Array(APPLICATION_SERVER_KEY),
-    });
+    let notificationSubscription = null;
+    try {
+        notificationSubscription = await registration.pushManager.subscribe({
+            userVisibleOnly: true,
+            applicationServerKey: urlBase64ToUint8Array(APPLICATION_SERVER_KEY),
+        });
+    } catch (error) {
+        return false;
```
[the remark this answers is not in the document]
sandroroux: Also, I think the exception should make it out of `pushManagerSubscribesToNotifications`. Exceptions improve code legibility by saying "this function can't do what it should". A caller wants to know when the function misfires, and the best way to do that is to check for the exceptions.

I'd put the `catch` in turnOnNotifications:
```js
    const turnOnNotifications = async (swRegistration) => {
    try {
        const notificationSubscription = await pushManagerSubscribesToNotifications(swRegistration);
    } catch (err) {
        if (err instanceof DOMException) {
            return null;
        }
    }
```

You could even wrap the DOMException in your own Error class, to increase legibility.

src/js/Notifications.js
```diff
@@ -118,6 +127,9 @@ export const supportsPushManager = async () => {
 export const subscribeToNotifications = async () => {
     const swRegistration = await navigator.serviceWorker.ready;
     const subscription = await turnOnNotifications(swRegistration);
+    if (!subscription) {
+        return false;
+    }
```
sandroroux, line 132: If you don't receive a subscription, this function should only `return`. In typescript terms, it's [a `void` function](https://www.typescriptlang.org/docs/handbook/basic-types.html#void), meaning it has no return type.

sandroroux: If you think a caller should know whether it succeeds, I'd argue that you bubble up an exception on failure.

PR #217, opened 2020-08-10
Fixes a JS bug that prevents users from browser-backing across translations

sandroroux:


katevoss:
How do I reproduce the bug? @sandroroux

PR #204, opened 2020-07-30
Truncate long course and lesson titles

katevoss:
Closes https://github.com/catalpainternational/asteroid/issues/189

PR #218, opened 2020-08-11
#204: Profile page shows guests the right data.

sandroroux:
Closes catalpainternational/asteroid#204.

### DIFF
--- a/src/js/WagtailPagesAPI.js
+++ b/src/js/WagtailPagesAPI.js
@@ -1,5 +1,5 @@
 import { BACKEND_BASE_URL, WAGTAIL_MANIFEST_URL } from "js/urls.js";
-import { getAuthenticationToken } from "js/AuthenticationUtilities.js";
+import { getAuthenticationToken, isGuestUser } from "js/AuthenticationUtilities.js";
 import {
     storeWagtailPage,
     getWagtailPageFromStore,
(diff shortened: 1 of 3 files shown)


src/riot/Screens/Profile.riot.html
```diff
@@ -126,7 +124,11 @@ <h5>{ TRANSLATIONS.lessonsLeft() }</h5>
 
             async renderProfilePage() {
                 const homePage = await getHomePage();
-                const courses = homePage.courses;
+                let courses = homePage.courses;
+
+                if (isGuestUser()) {
```
katevoss, line 129: If no courses are visible should we show a profile page at all? If there is a profile page then maybe there should be a message that says something like "You can see your progress here once courses are available"?
Right now it's an empty page:
image

src/js/WagtailPagesAPI.js
```diff
@@ -147,6 +147,9 @@ export const getResources = async () => {
     const resources = [];
     for (const childPageId of resourcesRoot.data.children) {
         const childPage = await _getOrFetchWagtailPageById(childPageId);
+        if (!childPage.is_visible_to_guests && isGuestUser()) {
```
katevoss, line 150: Should we do this same check in Courses? Even though a course is hidden from a guest the data (content, objective, test, exam) is still accessible in the JS console:
image

PR #219, opened 2020-08-11
Show message when there are no visible courses or resources

katevoss:
Closes https://github.com/catalpainternational/asteroid/issues/205


src/riot/WagtailPages/HomePage.riot.html
```diff
@@ -80,13 +88,23 @@ <h6 if="{state.currentCourse && isCourseInProgress(state.currentCourse)}">
...
                 });
             },
 
+            checkForVisibleCourses(courses) {
+                for (const course of courses) {
+                    if (course.is_visible_to_guests) {
+                        return true;
+                    }
+                }
+            },
```
PeteCoward, line 106: IF you are not a guest, and all course are not visible to guests, this will return false
isAnyCourseVisible will then be false, will that not cause a problem?

PeteCoward: also some people would say that you should try to make sure a function returns a consistent return type, by returning false at the bottom ( other wise it will return true or null which have differetn types )

katevoss: I changed where it checks for visible courses and resources, which I think negates the consistent return type.

src/riot/Screens/Resources.riot.html
```diff
@@ -95,6 +101,18 @@ <h3>{ TRANSLATIONS.resources() }</h3>
...
+            checkForVisibleArticles(articles) {
+                if (isGuestUser()) {
+                    for (const article of articles) {
+                        return article.is_visible_to_guests
+                    }
+                } else if (articles.length > 0) {
+                    return true;
+                } else {
+                    return false;
+                }
```
sandroroux, line 113: If you look at `getResources`, you'll see that it only returns articles guests can see. If all articles are invisible to guests, `getResources` returns `[]`, which means `state.articles` will be an empty array.

I don't think you should special-case for whether articles are visible.

src/riot/WagtailPages/HomePage.riot.html
```diff
@@ -72,21 +77,41 @@ <h6 if="{state.currentCourse && isCourseInProgress(state.currentCourse)}">
...
+            async onMounted(props, state) {
+                this.update({ isLoading: true });
                 const unsortedCourses = props.wagtailPage.courses.filter(
                     (course) => course.lessons.length
                 );
 
                 const currentCourse = this.getCurrentCourse(unsortedCourses);
                 const listOfTags = getTagsFromPages(unsortedCourses);
                 const sortedListOfCourses = this.sortCourses(unsortedCourses);
+                const isAnyCourseVisible = this.checkForVisibleCourses(unsortedCourses);
```
sandroroux, line 93: Looking at this again, I think the HTML should receive the filtered list of courses. Before you send `sortedListOfCourses` to the `update`, `sortedListOfCourses` should filter out courses that guests can't see:
```js
sortedListOfCourses = sortedListOfCourses.filter(this.isCourseVisible);
```

With that, you can remove `isAnyCourseVisible` and the `if={isCourseVisible(course)}`, trimming some code and complexity.

sandroroux: `Resources` can mirror the approach with articles, but the refactor is slightly different. You don't have to do it in this PR, but I can describe it to you if you're interested.

PeteCoward:
I'll leave this open today to allow you to respond to Alessandro. I don't think this is release critical

src/riot/WagtailPages/HomePage.riot.html
```diff
@@ -72,21 +76,40 @@ <h6 if="{state.currentCourse && isCourseInProgress(state.currentCourse)}">
...
                 });
             },
 
+            removeHiddenCoursesFrom(allCourses) {
+                let coursesVisibleToGuests = [];
+                allCourses.map((course) => {
+                    if (course.is_visible_to_guests) {
+                        coursesVisibleToGuests.push(course);
+                    }
+                });
```
sandroroux, line 109: You could write this as a `filter`:
```js
return allCourses.filter((aCourse) => aCourse.is_visible_to_guests);
```

Do you prefer `map`?

PR #234, opened 2020-08-21
#222: Reduces Time to Offline

sandroroux:
Closes catalpainternational/asteroid#222.

Timings are approximate, measured from when the login page loads. I used Chrome's "fast 3G".

## On RC-1.2
Time to `HomePage`: **9.5s**
Time to Offline: **1.4 min**
885 requests

## On this branch
Time to `HomePage`: **7s**
Time to Offline: **22.66s**
881 requests


src/riot/App.riot.html
```diff
@@ -9,14 +9,16 @@
 
         <Login if="{!isUserLoggedIn()}"></Login>
 
-        <DataDownload if="{isUserLoggedIn()}"></DataDownload>
+        <LoadCompletions if="{isUserLoggedIn()}"></LoadCompletions>
```
PeteCoward, line 12: I'm really not happy about spreading this pattern of controlling the initiation of non-interface components in the interface markup.
I would much prefer both completions and download components to be triggered by javascript (redux state changes) than in this indirect way

sandroroux: I wouldn't think of this as spreading a pattern. I agree, we should come up with a better way to transition screens and state.

For now, using the paradigm we have, this is the best way to guarantee performance and introduce a bit more clarity:
* no more nebulous "DataDownload" tag;
*the code shows you don't need to download the site before using the app; you must download completions.

sandroroux:
I've redirected this at RC-1.3.

src/riot/App.riot.html
sandroroux: Also, the split is a big contribution to this PR's performance. Prepping all the data up front was a bottleneck. This way, the home page is already interactive, and the site, navigable, before the app fetches pages and images.

PeteCoward: I totally agree with the optimisations part of it.
But it adds another non-ui component whose lifecycle is controlled by html markup, and riot mounting and unmounting. HTML and riot should not be involved in the lifecycle of those components. An interface developer should be free to mess with riot tags ( or even replace it with vue ;) ) without changing the behaviour of these components. I know we are not there yet, but I don't want to keep moving in the wrong direction.

Both those components should initiate on login, and stop on logout, irrespective of what the interface does.

PeteCoward (approved):
Let's get it in, there are conflicts that need resolving in the page classes PR against this

PR #237, opened 2020-08-25
Page Classes Replace Page JSON.

sandroroux:


sandroroux:
@katevoss, I couldn't reproduce the bug: the Continue Button does what I expect. Do you mind sending some reproduction steps?

By the way, I had to rebase my old branch. If you test my branch again, delete your old copy and `checkout` the rebased one.

katevoss:
I retested on your rebased branch and I'm still not seeing the Continue Button on home. 

1. Clear the site data 
2. Log in as a user (not guest). 
3. View home page
4. There is no Continue learning button or message:

Screen Shot 2020-09-02 at 10 56 51

5. I complete a lesson activity
6. Back to Home page
7. Still no Continue learning:

Screen Shot 2020-09-02 at 10 59 48


### However, if I 
2. log in as a guest
3. View the home page
4. Complete a lesson activity
5. Go to Home page
6. I see the Continue Learning button:

Screen Shot 2020-09-02 at 11 00 10


I don't see any console errors or anything else that looks wonky.

sandroroux:
@katevoss, I locally cloned staging's database, changed your password, and logged into your account. Here's what I see with my latest changes:
Screen Shot 2020-09-03 at 15 01 12

I checked the code paths, but cannot tell why the continue button differs for guests and non-. Here's the only completions-and-guests code (in `DataDownload.html.riot`):

```js
async prepDataForAppLaunch() {
    if (!isGuestUser()) {
        await updateIdb();
        this.state.completionPoller = window.setInterval(updateApi, EVERY_FIVE_MINUTES);
    }

    await this.pullIdbDataIntoMemory();

    new SiteDownloader().requestTheSitesPagesAndImages();
    this.state.pagesPoller = window.setInterval(
        this.pollForWebpageChanges,
        EVERY_FIVE_MINUTES
    );

    dispatchLoggedInEvent();
},
```
`updateIdb` `fetch`es a user's completion data and stuffs that data into the local completion database.

`pullIdbDataIntoMemory` pulls completions from the local database and stores them in a JavaScript `Map`. That way, we don't have to query a database whenever we need completion info (each db query would be asynchronous; better that UI completion code avoid `await`s).

The app might be mangling your local data, or the way I compute the most-recently worked-on course could be faulty. 

This is me brainstorming. I'll keep thinking.

sandroroux:
@katevoss, @PeteCoward, the bug existed outside of this PR, but I think this PR fixes it. `getLatestCompletion` is now aware of which lessons are published and un-.

The diff is rather large. Unless there are structural issues or bugs, maybe we should merge this and consider follow-up fixes.

src/js/CoursePage.js
```diff
@@ -0,0 +1,109 @@
...
+    return {
+        courseSlug,
+        lessonSlugs,
+    };
+};
+
+const countCompleteLessons = (wagtailCourse) => {
+    const { courseSlug, lessonSlugs } = getCourseAndLessonSlugs(wagtailCourse);
+    return getFinishedLessonSlugs(courseSlug, lessonSlugs).length;
+};
```
sandroroux, line 28: Nice-to-have: I can inline this. No need for a separate method.

sandroroux: Done.

katevoss (approved):
I agree that we should merge and fix any issues in subsequent PRs.

I am still not seeing the "Continue learning" message but I agree that it is likely due to the state of my user, which can be investigated in a separate issue. 👍🏻

PR #241, opened 2020-09-10
Jest tests

sandroroux:
Front-end files with tests:
* `js/HomePage.js`;
* `js/CoursePage.js`;
* `js/LessonPage.js`;
* `js/ResourceArticle.js`;
* `Actions/completion.js`.

To run the tests, `cd canoe/` and run `yarn run jest`.


PeteCoward:
I'm marking this as draft, we're working towards using Ava which Lee found easier to work with, @sandroroux let me know if there is anything from here you want to keep

PR #271, opened 2020-10-30
270 incorrect cookie parsing

boerwastaken:
e6d9290 fixes an incorrect reference.

22e6721 fixes incorrect cookie parsing and closes #270.

Before this fix:
```
> document.cookie="huey=louie=dewey"
< "huey=louie=dewey"
> getCookie("huey")
< "louie"
```
Boooo!

After this fix:
```
> document.cookie="huey=louie=dewey"
< "huey=louie=dewey"
> getCookie("huey")
< "louie=dewey"
```

Hurray!

PR #274, opened 2020-10-30
Minimal mode app booting

### LINKED ISSUE #269: RFC: Lean boot
## Feature goals:

- determine minimal state set (IndexedDB, Cache, Serviceworker, Cookies, LocalStorage, ...) sufficient for booting Canoe
- adjust project code to minimize the amount of state needing special project-specific treatment from an Appelflap point of view

## Rationale

For PeP, there will need to be a way to transport state from a connected-and-updated Appeflap instance, to an Appelflap instance that hasn't ever seen the Internet (and might never). 

Condensed user story:

1. Teacher walks to the 3G tamarind tree, webapp pulls in updated lesson content & updated service worker & "app code"
2. Teacher walks to the school where there is no internet whatsoever
3. Teacher's device shares updated artefacts with student's devices
4. Student's devices, **some of which may be a blank slate, some of which may never ever see internet**, incorporate the updates

There are many pieces to that puzzle, but let's focus for a bit on the ones that concern Canoe and any Appelflapped project that needs browser state to be propagated in a P2P fashion.

### On selection
We don't necessarily always want transported state to be an exact clone of the originator state. For instance, a student likely has no business with a session cookie of the teacher, or with another student's lesson progress. So clearly, the to-be-transported state will need to be a **selection** of state, not bluntly "all state".

There's two places where such a selection could be encoded.

1. in Appelflap. Appelflap would need to know, for every Appelflapped project, exactly what things need transporting, and what things are secret/private stuff (login tokens, lesson progress). And when this selection changes - say the frontend adds a new LocalStorage item that should or should not be transported - then that would need to be reflected in the Appelflap code, which is not that easily accessible to a frontender who just wants to make a little change, and getting such a change out to devices is going to be a pain. That's clearly undesirable. 
Also — **when** is state finalized? It's clearly no good to transport half-constructed Caches. But how does Appelflap know when a Cache build has been completed? That's a decision which is up to the webapp, how would Appelflap know about that decision? :sob: 

2. in the webapp. The webapp knows when its serviceworker is updated and when its caches are complete. The webapp can also have knowledge of which things are private (session cookies, lesson progress) and which things are not. It can even decide on having separate Caches - for public stuff, and for private stuff, which opens up possibilities for easy-to-understand granularity of packing up & restoring, which we'll get to.
Clearly, this is the right place to encode selections. The webapp knows best. :raised_hands: 

### On transportable browser state item types

Every storage state item type (cookies, Caches, serviceworkers, IDB, LocalStorage, ...) needs handling for extraction and injection. Of course we want to write as little code as we can get away with, so we'd implement as little of these as possible in Appelflap. What is the minimum set?

1. ServiceWorker - the serviceworker itself, plus imported scripts. Without that, there'll be no offline functionality.
2. CacheStorage (so not the transparent browser cache). In the Canoe case, that's where lesson content is stored.

These concern almost always verbatim copies of server-originating data, so obviously these are necessary to make offline state offline-transportable.

As for the other types, for simplicity it would be good to keep them out of the essentials for a minimal-state anonymous webapp boot. Remember we're not trying to create exact clones of a webapp environment (as said, why would student A need student B's lesson progress state?), but rather, we're trying to get to just the point where have supplied enough state to the webapp so it doesn't **need** an internet connection to get on with things. The rest of the state (IDB state of lesson progress, LocalStorage state of username etc) should be reconstructed by the webapp, accommodating for the possibility it might never see a live internet connection (so don't let the user be stranded on a login screen, for instance).

### On granularity
On the target device, originator state will need to be merged in with originator state. What's the granularity for that? Individual cookies? Individual entries in individual IDB databases? Whole IDB databases? Clean-slate approach for IDB databases (remove all, add only what's defined in the incoming package)? What about caches?

For simplicity and predictability, I propose granularity on a by-name scale, coupled with replace-all.
Thus a webapp will be able to express:
- "the cache named `A` will now need to be packed up"
- "the serviceworker with origin `B` and scope `C` will now need to be packed up"

and then, on the receiving side, the strategy is append-or-replace. That is to say, an existing Cache with the same name as an incoming updated Cache will be wholly replaced, but any other Caches that exist on the target device will be preserved. Basically it's a lot like filesystem copying semantics that way! This also leaves the door open for webapps to decide on naming schemes that facilitate garbage collection — embed version info in the cache name, and have the service worker clean out previous incarnations. For instance that's what Gather does to enable atomic webapp upgrades. But for simple use cases it'll be sufficient to use the replacement mechanism and not worry about garbage collection.

### On protocol Webapp <-> Appelflap <-> peer Appelflaps+Webapps

It's too early to set anything in stone. I'm currently thinking along these lines:

1) a HTTP API (through Appelflap's built-in web server) in favour of a direct JS API (using a WebExtension bridge into native Android code). Mostly because I'm not sure I can make extension scripts available to the serviceworker context, and I would like to keep the option open for the orchestration to be done from a serviceworker context.

2) namespaces for pubsub of artefacts. In the Canoe context, this is to enable behaviour where a student app subscribes to just a subset of the artefacts published by peers. For instance, a teacher app connected to the internet might pull in Chemistry and Mathematics content. The webapp will store these in separate Caches, to be published in separate namespaces ("chemistry" and "mathematics" spring to mind as namespace name candidates). Now, say a student webapp is only subscribed to the "mathematics" namespace and not to the "chemistry" namespace (because they're not enrolled in chemistry classes). Then Appelflap will not download and incorporate any artefacts in the "chemistry" namespace. We need a subscription mechanism like that, because the teacher device doesn't know the student identities a priori, and the webapp is not involved in who-is-who-and-who-gets-what. We're not going to turn a webapp into a webserver with dynamic personalized content. 
There will be a default namespace which always gets pulled in. So for the simplest use cases one can pretend that the namespaces don't exist.

3) It may be possible to poke the serviceworker with a special push message to inform that an update has taken place. Though if the webapp is smart and uses versioning in its artefact naming schemes, the injectee webapp should be able to determine this independently with some work.

boerwastaken:
Closes #269 

### Related
Comes on top of #268 and #271 . Those can be examined and merged first, or disregarded and treated as part of this PR. But it makes more sense to examine & merge those separately & first,  because they address separate issues (bugs) that have no direct relation to the #269 Lean booting goal (but I needed to fix them, because crash.). They're bugfixes, good to have regardless of this here PR.

***Note that there are no user-visible changes for existing use cases***. I'm trying to do the minimal, and the minimal can be improved upon. Especially keep in mind that yes, while `U+1f638` is debatable, no user and probably no dev will see it in the user interface, at least not until they mess about a bunch with their devtools. Hum-ing and hehm-ing about `U+1f638` is not what I want this PR and its reviews to be about.

### What does it do

It makes it so that Canoe can boot with just a `Serviceworker` and some `Cache`s installed. This makes it easier for an eventual Appelflap state transfer implementation to take care of, well, transferring state, as it doesn't have to set project-specific cookies/localStorage/IDB stuff in project-specific ways just to poke things into a state that the webapp doesn't crash with or does something weird with. It should be the webapp's responsibility to be able to bootstrap itself from a minimal state, reinitializing things where necessary. And now Canoe can! Hurray again!

In the process, out of necessity, I also made the "are we authenticated" evidence-based rather than belief-based.
It used to be based solely on presence or absence of a cookie with name "Token", disregarding the payload.
And even if it would be verifying the payload (which it can't, because currently Canoe's tokens are HMAC based rather than signature based!), it wouldn't know whether the server hasn't rotated keys or disabled the user. Evidence-based is just better.

In a nutshell:
- Don't assume presence of a cookie with name "Token" means we're authenticated (and will always be authenticated)
- Don't set the user stuck on a login screen when no such "Token" cookie is available — the user might be offline, unable to login, but might want to interact with the learning content that's available from the cache.
- Track authentication status. We're authenticated until proven otherwise, and once proven deauthenticated, we're deauthenticated until we've logged in again.

### Testing

1. Load the webapp on a clean profile as usual
2. Log in to fill your caches with e-learning goodness
3. Set yourself to offline (via network tab of devtools)
4. Play around with deleting any/some/all of Cookies, IDB, LocalStorage items, ... But not Caches, or the Serviceworker, duh.
5. For any of the operations of step 4, witness that the webapp still boots (ctrl-L, enter)
6. Hop online, load the webapp (ctrl-L, enter)
7. If you had deleted the cookie, you should now see a login screen - that you didn't see when you were offline, which is good, because it was no good presenting a login screen to a user who's offline!

#### Deauthentication detection

Currently the deauthentication-detection piggybacks on the requests to various backend-pages ("actions", lesson pages. I think). 

Nothing wrong with that, but I'm quite unfamiliar with this project, there may be better candidates that are more regularly polled. I don't know, is there anything that is ever polled again after boot? What if there are content updates? Is the user expected to refresh the page regularly or something?

In Gather I poll an endpoint on an interval, and this *invites* the user to re-login if it turns out that the user isn't anymore —
rather than dumping them to a login screen that can't be bypassed. Especially with intermittent connectivity, an invitation is the better approach — while there is a network available at the time of unauthenticatedness detection, there might not be one by the time the user walks back to their browser/tablet! So the poll-and-invite approach is clearly superior in that respect, especially if the invitation to re-login is automatically retracted when the network goes away (it's a stretch, but it'd be really good UX). In Gather, the re-login invitation is canceled if by some magic the authentication succeeds again, which is also nice. But in this app, with this PR, once a deauthentication is detected, you'll be dumped to the login screen (on reload anyway) with no way to restore access to the learning materials (that may actually be right there, on the device! how frustrating! locked out by your own devices! At least good oldfashioned books don't pull that trick on you!) other than having a) internet b) a valid login . For remote no-internet scenarios that's clearly undesirable. To fix that we'll need a wider discussion. I'd suggest a local-only-mode one-button on the login screen that just lets you get on with what you have on the device.

#### Auto-guest-login

Currently, as I'm just trying to do minimally invasive stuff to get to the desired goal of lean boots, no super smart behaviour is set up. But I saw that there was something like a "guest mode" on haroman.tl. On a cold boot of the webapp, the user could be offered to use the guest account instead of the rather minimal anonymous mode that's currently set up with this PR.


src/js/AuthenticationUtilities.js
```diff
@@ -22,50 +25,35 @@ const setCookie = (name, value, keyOnlyAttributes = [], attributes = {}) => {
...
+    })
 };
 
-export const login = async (usernameAndPassword) => {
-    let response = null;
-
-    try {
-        response = await fetchAuthToken(usernameAndPassword);
-    } catch (error) {
-        throw error;
```
boerwastaken, line 65: catch and throw it again! and again! and then some more! :smile_cat:

src/js/AuthenticationUtilities.js
```diff
@@ -22,50 +25,35 @@ const setCookie = (name, value, keyOnlyAttributes = [], attributes = {}) => {
...
         const [key] = tokenString.split("=");
         return key === name;
     });
 
-    if (!cookieValue) {
-        throw new Error(`Can't find "${cookieValue}" in cookie.`);
-    }
-
-    return cookieValue.split("=")[1];
+    return found_cookie? found_cookie.substring(found_cookie.indexOf('=') + 1): null
```
PeteCoward, line 33:
```suggestion
    return found_cookie ? found_cookie.substring(found_cookie.indexOf('=') + 1) : null
```

src/js/AuthenticationUtilities.js
```diff
@@ -120,3 +107,7 @@ export const getUserId = () => {
 export const getUserGroups = () => {
     return localStorage.getItem(USER_GROUPS_STORAGE_KEY);
 };
+
+export const setIsAuthed = (somebool) => {
```
PeteCoward, line 111: camelCase
```suggestion
export const setIsAuthed = (someBool) => {
```

src/js/AuthenticationUtilities.js
```diff
@@ -120,3 +107,7 @@ export const getUserId = () => {
...
+    return localStorage.setItem(USER_IS_AUTHED_STORAGE_KEY, Boolean(somebool));
```
PeteCoward, line 112: camelCase
```suggestion
    return localStorage.setItem(USER_IS_AUTHED_STORAGE_KEY, Boolean(someBool));
```

src/js/AuthenticationUtilities.js
```diff
@@ -1,11 +1,14 @@
 import { BACKEND_BASE_URL } from "js/urls";
 import { dispatchLoggedOutEvent, dispatchSiteDownloadEvent } from "js/Events";
 import { unsubscribeFromNotifications } from "js/Notifications";
+import { fetch_and_denote_unauthenticatedness as fetch } from "./Fetch";
 
 const USERNAME_STORAGE_KEY = "username";
+const ANONYMOUS_USER = "😸";  // U+1f638 Grinning cat face with smiling eyes
```
PeteCoward, line 7: I think for safety we should avoid the risk of showing people grinning cat emoji's by either setting it to something more descriptive, or not returning it from the getUsername function

PeteCoward: I know, should never see it, but developers do funny things, and if they do I'd rather it was descriptive

boerwastaken: I first tried returning an empty string but that makes Canoe crash.
Ideally it'd return `null` when there isn't one, it's the most appropriate response, but the callers need to be able to handle that.

boerwastaken: Now, absense of any stored username makes you a guest, coalescing the no-user case with the guest-user case. No more cats.

src/js/AuthenticationUtilities.js
```diff
@@ -85,6 +73,7 @@ export const logout = async () => {
     deleteCookie(JWT_TOKEN_STORAGE_KEY);
     localStorage.clear();
     unsubscribeFromNotifications();
+    setIsAuthed(false);
```
sandroroux, line 76: Shouldn't this be unnecessary? If the user logged out, there will be no `USER_IS_AUTHED_STORAGE_KEY` in local storage. `localStorage.getItem(USER_IS_AUTHED_STORAGE_KEY)` would return `null`.

src/js/AuthenticationUtilities.js
```diff
@@ -93,15 +82,12 @@ export const getAuthenticationToken = () => {
 };
 
 export const isUserLoggedIn = () => {
-    try {
-        return !!getAuthenticationToken();
-    } catch (e) {
-        return false;
-    }
+    const auth_status_denoted = localStorage.getItem(USER_IS_AUTHED_STORAGE_KEY);
+    return (auth_status_denoted === null || auth_status_denoted === "true");
```
sandroroux, line 86: Why does the lack of a `USER_IS_AUTHED_STORAGE_KEY` imply that the user is logged in? Am I misunderstanding this conditional?

sandroroux (changes requested):
@boerwastaken, I don't quite follow this PR's goal. If the user lacks indexedDB and `localStorage` data, we can't guarantee the app will work correctly, so we want the app to crash gracefully (which it doesn't right now, I see).

Some new bugs arise if we change this assumption. Let's say I delete `username` from local storage: with your change, I'm switched to "guest" mode. Even though I'm a guest, some of my user account's information carries forward to guest mode:
* my username's `userId` and `userGroups` still submits with analytics calls.
* my user name's completion data carries into the guest mode's course materials.

These aren't new bugs, but different bugs, so it's not a fault of your fix. I still think we should ask whether this approach is the best. Can we discuss why Appelflap needs our app's assumptions to change?

src/js/AuthenticationUtilities.js
boerwastaken: Because logged out state is innocent until proven guilty, so to speak.
If it has not been witnessed that a user is logged out, they're not.
Flipping this around (guilty until proven innocent, so to speak), Canoe would dump the user to a login screen on a lean offline boot, which is undesirable. Hence, innocent until proven guilty.

There's certainly smarter things that Canoe could do in this case, some of which I hint at in the PR under *Possible further improvements*. But I don't want to change too much of Canoe right now.

src/js/AuthenticationUtilities.js
boerwastaken: > If the user logged out, there will be no USER_IS_AUTHED_STORAGE_KEY in local storage.

Is that so? Where is the code which removes this localstorage item when the user logs out?

If the user explicitly logs out then we don't want to second-guess that. As explained below, unless unauthedness is proven (which can't happen offline), the user is presumed logged in, so removing explicitly setting `USER_IS_AUTHED_STORAGE_KEY` to false to reflect explicit user intent is the right thing to do currently. There's certainly improvements possible, as I write in the PR under *Possible further improvements*. But that's outside of the scope of what I'm trying to get working here.

boerwastaken:
> I don't quite follow this PR's goal.

Aight, I've expanded #269 to give more context, you may have been out of the loop. It's an RFC now, so if you need more clarification or have alternate ideas, or comments, please put them in there!


> If the user lacks indexedDB and localStorage data, we can't guarantee the app will work correctly, so we want the app to crash gracefully (which it doesn't right now, I see).

I don't think anyone's arguing that Canoe should be able to survive and repair arbitrary damage to IDB and localStorage data.

What I would like to get to is at least a well-defined minimal bundle of offline state necessary for Canoe to a) not crash and hopefully b) to do some more interesting things than just dump the user to a login screen. This PR does the minimal to accomplish a) and b).
As for "well-defined minimal bundle of offline state", it turns out that with little poking Canoe can reach the ideal minimal state of just-a-serviceworker-with-just-some-caches, as outlined in #269. If more state transfer, beyond Caches and Serviceworkers, is required, then that can be argued for in #269.

> Some new bugs arise if we change this assumption. Let's say I delete username from local storage: with your change, I'm switched to "guest" mode. Even though I'm a guest, some of my user account's information carries forward to guest mode:
>
>    my username's userId and userGroups still submits with analytics calls.
>    my user name's completion data carries into the guest mode's course materials.

Though it's fun seeing what happens when individual items are removed (applying arbitrary damage), and encouring to see that no permutation of removals make Canoe crash anymore now, robustness in the face of arbitrary damage is not the goal; "arbitrary damage can happen" has never been the assumption, and I don't think it'll ever need to be.

The much less wild assumption that should be worked with from now on is that any serviceworkers or caches are intact, and localStorage/IDB/cookies are either all missing, or intact. See #269 for the considerations from which this follows, and discuss there.

This PR is mostly bug fixes, and introduces no new behaviour for all existing deployment scenarios. It's a starting point so we have Canoe in a state where we can transplant its innards without the receiving vessel crashing.
More thoughtful handling of guest mode is certainly something worthwhile to have in Canoe. Leaving the user stranded on a login screen and locking them out from their study materials until they've found internet and succeed to login is not good style. But those are not issues that need to be addressed *right now*.

sandroroux (approved):

Issue #254, opened 2020-10-27
Show a course is in-progress, not started, or complete


katevoss:
CoS
- [x] course card can have these states:
   - [x] Active: no icon, no progress bar (not started)
   - [x] In-progress: progress bar, blue play icon
   - [x] Complete: progress bar 100% complete, green check icon

andrewwapling:
Hi @katevoss, 

The key states of Course Cards are:

1. Active – No Icon, No Progress Bar
2. In Progress – Blue Arrow Button Flag, Progress Bar displaying current progress through course via Lesson Card progression
3. Complete – Green Check Flag, Progress Bar 100% complete 
4. New – New Flag sitting on edge of card

Lesson cards follow the same structure as above except there is also a state for lessons not yet ready to be completed:

1. Locked – Locked Flag, Greyed out tone and image desaturation

Hope this helps and sing out with any further queries at all.

Thanks again, Andy

katevoss:
@andrewwapling can you export the icons for the blue "play" arrow, green "complete" check, and grey "locked" padlock? In InVision is fine https://invis.io/52Y02L9YTRF#/432778428_2C-_Course_Listing_-_Short_Version thanks!

andrewwapling:
Hi @katevoss , No worries at all. I've marked these items as exportable and re-synced to Invision so that you can access and download these assets: https://invis.io/52Y02L9YTRF

Let me know if any queries issues at all.

Thanks again, Andy

katevoss:
@andrewwapling Unfortunately the icons didn't export correctly, they show the words ("lock") rather than the visual (padlock)
![image](https://user-images.githubusercontent.com/2395211/98136435-ee789000-1e8e-11eb-9366-c977aa358510.png)

andrewwapling:
Hi @katevoss , That's odd. I've just re-synced again and appears ok at this end now. Can you please check? I've also exported manually here to be doubly sure. Hope this helps, Andy
[Flags.zip](https://github.com/catalpainternational/canoe/files/5490999/Flags.zip)

katevoss:
@andrewwapling The icons look correct on InVision now and thanks for also including them in a zip 👍🏻

andrewwapling:
@katevoss Excellent! Thanks for confirming with me 👍

PR #275, opened 2020-11-02
Add card status

katevoss:
Closes https://github.com/catalpainternational/canoe/issues/254

I made the cards on Resources, Profile, Course and Lesson page look good and did minimal support for showing their status and cleanup. I'm sure there will be more to clean up and improve when we get around to actually implementing those pages, but I wanted to leave it in a place where it looks okay.


src/riot/Components/Card.riot.html
```diff
@@ -1,8 +1,17 @@
 <Card>
-    <a class="card {props.isCompleted ? 'done' : ''}" href="#{props.link}">
-        <div class="card-image"></div>
+    <a class="card {props.status}" href="#{props.course ? props.course.id : props.link}">
```
katevoss, line 2: I added the ternary to the `href` so that the card can accept either when a course is passed down or when the link is passed down. 

As we go through the rest of the redesign, we'll switch over to one way of passing down the `course`.

sandroroux: As we discussed, consider pulling out `course` references from `Card`. Treat `Card` like a display component. Its `props` could be:
* `link`
* `status`
* `title`
* `progressValue`
* `progressMax`

src/riot/Screens/Profile.riot.html
```diff
@@ -147,6 +130,16 @@ <h5>{ TRANSLATIONS.lessonsLeft() }</h5>
                     this.getPercentOfLessonsCompleted()
                 );
             },
+
+            courseStatus(course) {
+                if (course.numberOfFinishedLessons == 0) {
```
sandroroux, line 135: Reminder to decide what to do about `==` vs. `===`.

sandroroux (commented):
Let me know what you'd like to do about my comments.

src/riot/Components/Card.riot.html
```diff
@@ -1,8 +1,22 @@
...
-        <div class="card-image"></div>
+    <a class="card {props.status}" href="#{props.link}">
+        <div class="card-image">
+            <span class="card-status">
+                <span class="status-icon"></span>
+            </span>
+        </div>
         <div class="card-title">
             <h5>{props.title}</h5>
+            <progress
```
katevoss, line 10: I added a separate bar for when courses are in-progress as they are the only object that can be "in-progress". Therefore they will pass in the props `progressValue` and `progressMax` but no other objects (lessons, references, etc) will.

sandroroux (approved):

PR #281, opened 2020-11-05
elaborate on conditions requiring logging in; fixes #280

### LINKED ISSUE #280: Crash on first load, login screen on second load
For me (Chromium 87) the first load attempt of Canoe (in dev-mode, ce283c12) fails. A subsequent load succeeds.

This is fallout from #274, specifically, if Riot thinks it's authed it will try to read all sorts of things, but is not prepared for any failures. But the failures **do** result in proof that the login screen needs to be shown, which is why a second load has the desired effect (of loading the login screen).

boerwastaken:
Fixes #280 

When there's no manifest & home page etc in the Cache, most Riot components will crash because they (or the stack below) don't handle the HTTP 403s arising from trying to fetch the manifest.
However, since #274, the 403s result in recorded evidence that the user is logged out, so on the next refresh the login screen is shown (as it should be).

This PR attempts to remedy the situation by presenting the login screen when the manifest is empty. Thus, a first load with empty caches will take users to the login screen, but a load with caches injected from outer space (offline Appelflap magic) will not strand users on the login screen, as in that case there'll be a manifest in the Cache, deo volente.

### DIFF
--- a/src/riot/App.riot.html
+++ b/src/riot/App.riot.html
@@ -7,16 +7,16 @@
         <OnlineOrOffline></OnlineOrOffline>
         <Toast></Toast>
 
-        <Login if="{!isUserLoggedIn()}"></Login>
+        <Login if="{userShouldLogin()}"></Login>
 
-        <LoadCompletions if="{isUserLoggedIn()}"></LoadCompletions>
+        <LoadCompletions if="{!userShouldLogin()}"></LoadCompletions>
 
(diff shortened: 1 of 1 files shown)


sandroroux (approved):

boerwastaken:
See also https://github.com/catalpainternational/canoe/pull/287#issuecomment-734235171

PR #284, opened 2020-11-09
CanoeTube frontend

boerwastaken:
Companion PR in canoe-backend: catalpainternational/canoe-backend#76

This lays some groundwork for "CanoeTube" self-hosted videos, with indicated entry points for doing Appelflap/Eekhoorn-related off-web in-Android-app storage.

There's some improvements to be made, I'll add inline comments to draw attention to those areas.

PR #285, opened 2020-11-13
Canoetube frontend v2

boerwastaken:
Companion to catalpainternational/canoe-backend#77
Supersedes #284

Adds resolving of media via media manifest.

### DIFF
--- a/src/js/utilities.js
+++ b/src/js/utilities.js
@@ -1,6 +1,6 @@
 import { dispatchToastEvent } from "js/Events";
 import { getLanguage } from "ReduxImpl/Interface";
-import { BACKEND_BASE_URL } from "js/urls";
+import { BACKEND_BASE_URL, MEDIA_PATH } from "js/urls";
 
 export const alertAppIsOffline = () => {
     dispatchToastEvent("You are offline.");
@@ -23,7 +23,7 @@ export const isCourseInTheCurrentLanguage = (courseSlug) => {
 };
 
 export const getMediaUrl = (mediaPath) => {
-    return `${BACKEND_BASE_URL}${mediaPath}`;
+    return `${BACKEND_BASE_URL}${MEDIA_PATH}/${mediaPath}`;
 };
 
 export const doesTheArrayContainTheObject = (theArray, theObject, matchingFunction) => {
--- a/src/js/utilities.js
+++ b/src/js/utilities.js
@@ -1,6 +1,7 @@
 import { dispatchToastEvent } from "js/Events";
 import { getLanguage } from "ReduxImpl/Interface";
 import { BACKEND_BASE_URL, MEDIA_PATH } from "js/urls";
+import { getOrFetchManifest } from "js/WagtailPagesAPI";
 
 export const alertAppIsOffline = () => {
     dispatchToastEvent("You are offline.");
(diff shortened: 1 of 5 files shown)


src/riot/Lesson/VideoCard.riot.html
```diff
@@ -1,18 +1,14 @@
 <VideoCard>
     <div>
         <div class="film"></div>
-        <video
+        <video id="thevideo"
```
PeteCoward, line 4: We should avoid using ids. If we drop more than one of these tags in a page it would fail. See comment below

src/riot/Lesson/VideoCard.riot.html
```diff
@@ -46,10 +43,13 @@ <h3 class="media-title">{props.media.title}</h3>
                 if (!this.state.cached) {
                     this.toggleAvailableOffline();
                 }
+                document.getElementById("thevideo").src = await this.getVideoMediaUrl(props.media.source);
```
PeteCoward, line 46: I think something like `this.$('video')` is the best way to do this in riot

PeteCoward (commented):
We should look at the AudioCard too ( transcodes includes audio and video )

src/riot/Lesson/VideoCard.riot.html
boerwastaken: Thanks. Ideally the media source would be templated in right from the Riot template, but that doesn't fly with async.

src/riot/Lesson/VideoCard.riot.html
boerwastaken: Applies

boerwastaken:
> We should look at the AudioCard too ( transcodes includes audio and video )

Looked and applied same mechanism.

src/js/utilities.js
```diff
@@ -23,7 +24,15 @@ export const isCourseInTheCurrentLanguage = (courseSlug) => {
 };
 
 export const getMediaUrl = (mediaPath) => {
-    return `${BACKEND_BASE_URL}${mediaPath}`;
+    return `${BACKEND_BASE_URL}${MEDIA_PATH}/${mediaPath}`;
+};
+
+export const resolveMedia = (mediaID) => {
+    return getOrFetchManifest()
+        // Choose smallest media item. Much more elaborate strategies are possible, but they need coordination with the backend
```
boerwastaken, line 32: I wonder whether we'll ever have any projects using multiple `TranscodeDefinition`s per media type, so selection among renditions could be a moot point, in practice.
I even wonder whether anyone will tailor transcode definitions to the project's use cases rather than use the [example fixture](https://github.com/catalpainternational/canoe-backend/blob/a8b0ac4a701d730977699cee20b67465de3716b4/wagtailmedia_transcode/fixtures/TranscodeDefinition.json) (but it seems ogg/vp8/vp9 are currently not supported on Safari :thinking: ).

src/riot/Lesson/AudioCard.riot.html
```diff
@@ -4,15 +4,14 @@
         <h3 class="media-title">{props.media.title}</h3>
         <p class="media-description">{props.media.description}</p>
         <audio controls crossorigin="anonymous" onplay="{ (event) => storeAnalytics(event) }">
-            <source src="{getMediaUrl(props.media.sources[0].src)}" type="{props.media.sources[0].type}" />
```
katevoss, line 7: Oh good, you removed this. I admit to adding it and I never liked it, good riddance to bad code!

src/riot/Lesson/VideoCard.riot.html
```diff
@@ -9,10 +9,6 @@
             onplay="{ (event) => storeAnalytics(event) }"
         >
             <!-- if preload set to none the video will not actually be requested, if set to auto it is requested -->
-            <source
-                src="{getMediaUrl(props.media.sources[0].src)}"
```
katevoss, line 13: Ditto here, good riddance!

src/riot/Lesson/AudioCard.riot.html
boerwastaken: I removed it out of necessity — as the media URL determination now depends on an async chain of events, it can't be templated in this way, so instead we spawn a coroutine in onMount (which you won't know from looking at the template) that pokes the element at some indeterminate moment in the future... I think I'd actually prefer an explicit call in the template! But AFAIK Riot can't do that, I think it would end up throwing the string representation of a Promise in there rather than the Promise-boxed result, utterly useless :rofl:
Is there a better way to use Riot templating with Promises/async functions?

PeteCoward:
This is being deployed to https://olgeta-staging-2004.temp.build via the `staging` branch on the olgeta repository
( ansible branch `update_canoe_hosts` from focal )
Frontend available on https://staging.olgeta.org.sb/ 
( `./deploy-canoe-staging-2004.sh` in olgeta repo `staging` )

PeteCoward (approved):

PR #288, opened 2020-11-26
Feature: Offlining Support

PeteCoward:
Just an experiment with writing a function to decouple our UI from the data fetching / storing
And to learn a bit, and maybe make @md8n comfortable introducing a scoped "You may typescript here" area

Entry point is `getDataAndRender` which is pretty duplicative of what we have but the CanoeManifest interface can be designed to help us understand the requirements of the new manifest structure, so I ended up trying to simplify the load story to help get that flow readable.

If I can understand what kind of json structure would support our needs then we can spec it and split the frontend and backend work, and even take @sandroroux 's jest implementation and verify that components behave as they should in isolation.

Not for merge, just for thinkx


PeteCoward:
The entry points to examine offline availability, trigger network request, background update content would be added, but use the same Store / Cache / Fetch / Manifest interfaces

@boerwastaken the stuff we were talking about with choosing which caches things go in would all be in the implementation of the CanoeCache ( once expanded to handle more than API JSON data )

PeteCoward:
pending a full test of this against current functionality still, would be good to merge if nothing is broken

md8n (dismissed):

PR #289, opened 2020-11-26
add preboot stage for app, determine blank/preseeded state before booting

### LINKED ISSUE #269: RFC: Lean boot

boerwastaken:
Fixes #281 
Fixes #269

Brings when the app bundle boots under our control (always nice to have the option of exercising control.). In this case it's used to determine "blank slate" vs "preseeded slate" state, and only when that has been determined, boot the app, which can then access that information in a synchronous fashion, so that the Riot template logic works without modification.

This fixes shortcomings (well, outright brokenness) of the previous implementation. As I've mentioned elsewhere, `ReduxImpl/Interface.getManifestFromStore()` doesn't actually always give you the manifest (I was enticed by the name, but was also wondering how it would do that in synchronous code, since the manifest is in a Cache... I shouldn't have been so trusty). While a probe for a manifest is an appropriate method to determine "is this a from scratch boot, or are we running on teleported caches", the answer can only come in an async way (because that's how the Cache interface works), while the code that needs to know — Riot template conditionals — is synchronous.

### DIFF
--- a/src/index.hbs
+++ b/src/index.hbs
@@ -86,6 +86,49 @@
             max-height: 100px;
         }
     </style>
+        <script>
+        const MANIFEST_CACHE_NAME = "manifest-cache";
+        const MANIFEST_URL = "{{{ htmlWebpackPlugin.options.api_base_url }}}/manifest";
+        const EMPTY_SLATE_BOOT_KEY = "empty_slate_boot";
+        var webpack_script_sources = [
+            {{# each htmlWebpackPlugin.tags.bodyTags }}
+            new DOMParser().parseFromString(new DOMParser().parseFromString("{{this}}", "text/html").documentElement.innerText, "text/html").scripts[0].src,
+            {{/each}}
+        ];
+        window.onload = (event) => {
+            // First deduce whether we're in blank slate state vs preseeded state, by checking for a cached manifest.
+            // Only then boot the app, so that Riot can synchronously access this fact.
+            caches.keys()
+                .then(cachenames => {
+                    if (cachenames.indexOf(MANIFEST_CACHE_NAME) < 0) {
+                        sessionStorage.setItem(EMPTY_SLATE_BOOT_KEY, true);
+                        return true;
+                    }
+                    else {
+                        return caches.open(MANIFEST_CACHE_NAME)
+                        .then(thecache => {
+                            return thecache.match(MANIFEST_URL);
+                        })
+                        .then(manifest => {
+                            let is_empty_slate_boot = (manifest === undefined);
+                            sessionStorage.setItem(EMPTY_SLATE_BOOT_KEY, is_empty_slate_boot);
+                            return is_empty_slate_boot;
+                        })
+                    }
+                })
+                .catch(err => console.error)
+                .then(_whevs => {
+                    webpack_script_sources.forEach(src => {
+                        script = document.createElement("script");
+                        script.src = src;
+                        script.type = 'application/javascript';
+                        script.async = false;
+                        script.defer = true;
+                        document.head.appendChild(script);
+                    });
+                });
+        };
+    </script>
 </body>
 
 </html>
--- a/src/riot/App.riot.html
+++ b/src/riot/App.riot.html
@@ -50,7 +50,7 @@
     </template>
 
     <script>
-        import { isUserLoggedIn } from "js/AuthenticationUtilities";
+        import { isUserLoggedIn, USER_IS_AUTHED_STORAGE_KEY } from "js/AuthenticationUtilities";
         import { getPage } from "js/Routing";
         import { ON_LOG_IN, ON_LOG_OUT, ON_REQUEST_SITE_DOWNLOAD } from "js/Events";
         import { isBrowserSupported, areCompletionsReady, getManifestFromStore } from "ReduxImpl/Interface";
@@ -182,10 +182,10 @@
             },
 
             userShouldLogin() {
-                return !(
-                    localStorage.getItem("fetch_result_indicates_authed") === "true"
-                    || Object.keys(getManifestFromStore()).length && isUserLoggedIn()
-                );
+                const is_deauthed = localStorage.getItem(USER_IS_AUTHED_STORAGE_KEY) === "false";
+                const is_firstboot = sessionStorage.getItem("empty_slate_boot") === "true";
+                const is_user_logged_in = isUserLoggedIn();
+                return is_deauthed || is_firstboot || !is_user_logged_in;
             },
 
             createHomePage(aWagtailPage) {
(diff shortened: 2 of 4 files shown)


src/index.hbs
```diff
@@ -86,6 +86,49 @@
             max-height: 100px;
         }
     </style>
+        <script>
```
PeteCoward, line 89: this hbs file is designed to be interface only. The preferred place for this kind of stuff is in `src/index.js`

src/riot/App.riot.html
```diff
@@ -182,10 +182,10 @@
             },
 
             userShouldLogin() {
-                return !(
-                    localStorage.getItem("fetch_result_indicates_authed") === "true"
-                    || Object.keys(getManifestFromStore()).length && isUserLoggedIn()
-                );
+                const is_deauthed = localStorage.getItem(USER_IS_AUTHED_STORAGE_KEY) === "false";
+                const is_firstboot = sessionStorage.getItem("empty_slate_boot") === "true";
+                const is_user_logged_in = isUserLoggedIn();
```
PeteCoward, line 187: Logic like this should be kept out of riot files. I want to remove this isLoggedIn / userShouldLogin from the riot.

boerwastaken: :+1: if you find a good home for it.

src/index.hbs
boerwastaken: Aaaah, well, putting it in `index.js` would also probably obviate the need for these delayed loading acrobatics, even better.

boerwastaken:
Superseded by #290

PR #292, opened 2020-12-01
Ava inclusion

md8n:
Instead of Jest for testing this includes Ava (this branch did start with trying to include jest).

Why? Because although Jest is under active development it still includes significant numbers of deprecated packages - this is not a good sign ...

Currently the 'best' testing set up is Cypress. Cypress is an awesome full testing suite - but it's also a FULL testing suite, and it struck me as overkill for our needs, and it had the wrong 'focus'.

Mocha + Jasmine, etc I also rejected based on past experience - it's painful to setup.
QUnit is also rejected.  It's reliable, but it's also a very old approach to JS testing.

Ava seemed to be the best compromise in terms of its focus, simplicity of setting up vs. complexities of figuring out the correct config for our needs, and its speed. 

Other changes introduced:
- A 'real' test file, plus some associated 'real' test data files.  These will need to be deleted once we have some proper tests in place.


src/ts/Interfaces/StoreCacheFetch.ts
```diff
@@ -0,0 +1,10 @@
+import { CanoeStore } from "./Store";
+import { CanoeCache } from "./Cache";
+import { CanoeFetch } from "./Fetch";
+
+/** CanoeStoreCacheFetch - These three are commonly used together */
+export interface CanoeStoreCacheFetch {
```
PeteCoward, line 6: this is unrelated to the task you were working on

webpack.config.js
```diff
@@ -34,12 +36,19 @@ module.exports = (env) => {
     );
 
     // merge environment and project configurations for use in webpack compilation
-    // webpack.DefinePlugin will replace process.env.CONFIG_KEY with configured valuea
-    const processEnvironment = Object.keys(environmentConfiguration).reduce((prev, next) => {
-        prev[`process.env.${next}`] = JSON.stringify(environmentConfiguration[next]);
-        return prev;
-    }, {});
-    processEnvironment["process.env.SITE_NAME"] = JSON.stringify(projectConfiguration.SITE_NAME);
+    // webpack.DefinePlugin will replace process.env.CONFIG_KEY with configured values
```
PeteCoward, line 39: please turn off format whole file in your editor

src/tests/resource.test.js
```diff
@@ -0,0 +1,40 @@
+import test from "ava";
```
PeteCoward, line 1: no need for this test file, please remove

src/ts/Interfaces/StoreCacheFetch.ts
md8n: Only somewhat - with eslint injected into the testing process it would complain about the number of parameters, making this combined object got past all of that,

messages_scan.js
```diff
@@ -22,7 +26,7 @@ const jsParser = extractor
             comments: jsParserCommentOptions,
         }),
     ])
-    .parseFilesGlob("./src/**/*.@(ts|js|tsx|jsx)");
+    .parseFilesGlob("./src/**/*.@(ts|js|mjs|cjs)");
```
md8n, line 29: Removed react specific glob patterns and replaced them with ones that are more relevant.

webpack.config.js
md8n: I formatted the whole file deliberately - It's a one-time payment off technical debt, it only requires review once.

PeteCoward: It takes 2 minutes to do that in a single PR and 1 minute to review and merge.
This helps prevent conflicts, and helps people reading this PR to understand the changes made

md8n: done

messages_scan.js
md8n: now in a separate pr

PeteCoward (approved):

PR #304, opened 2020-12-10
Sync tag placeholders

md8n:
This adds an extra button to the bottom menu of Canoe, which opens up a sync page.

The presence of the button is appelflap-aware, i.e. Canoe has to be running within appelflap for the sync button to be shown.

Click the button and a dummy sync page is shown.

The bottom menu and the sync page changes have not been styled properly.  The sync page itself will have significant functionality associated with it so it will have a very different appearance once that's down.  This :pr: is simply about the basic connection between Canoe and Appelflap.


src/ts/AppelflapConnect.ts
```diff
@@ -0,0 +1,46 @@
...
+export const appelflapFetch = async (
+    request_or_url: RequestInfo,
+    maybe_fetchopts: RequestInit | undefined
+): Promise<Response> => {
+    return await fetch(request_or_url, maybe_fetchopts);
+};
+
+export const appelflapMeta = async (): Promise<string> => {
+    const portNo = getPortNo();
+    const statusRequestURI = `http://127.0.0.1:${portNo}/meta/status`;
```
PeteCoward, line 35: I'd move the port and host into the `appelflapFetch` function above ( maybe later )

md8n: Done

PR #305, opened 2020-12-14
Appelflap cache api

md8n:
Implements the low level calls to Appelflap, as per Wicher's documentation.

Also adds a proper test implementation for checking the `getPortNo` functionality.


src/ts/AppelflapConnect.ts
```diff
@@ -11,55 +83,223 @@ export class AppelflapConnect {
...
-        const response = await this.appelflapFetch("meta/status");
+    private performCommand = async (
+        commandPath: string,
+        commandInit?: RequestInit,
+        returnType: "json" | "text" = "json"
+    ): Promise<any> => {
+        const response = await this.appelflapFetch(commandPath, commandInit);
 
         if (!response.ok) {
             return "Nope, that's not OK";
```
boerwastaken, line 143: if you're going to return a string (is that better than raising an exception?), wouldn't it be more helpful to template in the response code? with a reference to the API docs perhaps, so that it's more clear what's awry.

I'm not sure that choosing an error path (whatever form it takes) for `!response.ok` is the right thing to do. Perhaps the code wants to probe for a certain cache publication, in which case a 404 should be usefully distinguishable from a 500.

md8n: @boerwastaken - this is just placeholder code until we figure out what's really required for this interaction.

In general for JS it's preferable to not throw an exception, unless something 'exceptional' actually took place.  However, because we're still figuring things out I expect that some cases may warrant an exception.

src/ts/AppelflapConnect.ts
```diff
@@ -11,55 +83,221 @@ export class AppelflapConnect {
...
-            this._localHostURI
+        const requestInfo = `${
+            this.#localHostURI
         }:${this.getPortNo()}/${commandPath}`;
-        if (requestPath) {
-            requestInfo = `${requestInfo}/${requestPath}`;
-        }
+
+        /** When we finally have authorization via Appelflap, we'll change this */
+        const weHazAuthorization = false;
```
PeteCoward, line 117: pleaz no text speak
```suggestion
        const weHaveAuthorization = false;
```

src/ts/AppelflapConnect.ts
```diff
@@ -11,55 +83,221 @@ export class AppelflapConnect {
...
         if (!response.ok) {
             return "Nope, that's not OK";
         }
 
-        const pagesResponseJSON = await response.json();
+        switch (returnType) {
+            case "json":
+                return await response.json();
+            case "text":
+                return await response.text();
```
PeteCoward, line 150: which of the appelflap methods requires text return?

PeteCoward: oh i see from below. Not sure why though, could we say appelflap always returns json, and any text body that is not json is swallowed or errored

src/ts/AppelflapConnect.ts
```diff
@@ -11,55 +83,221 @@ export class AppelflapConnect {
...
 
         const requestInit =
-            method !== "GET" ? ({ method: method } as RequestInit) : undefined;
+            !weHazAuthorization ||
+            !commandInit ||
+            JSON.stringify(commandInit) == '{"method":"GET"}'
+                ? undefined
+                : commandInit
+                ? commandInit
+                : ({} as RequestInit);
```
PeteCoward, line 126: if won don't need auth, or we don't have a `commandInit` passed, or our init just specifies GET
then use `undefined`
else use `commandInit` if it exists otherwise an empty RequestInit

Took me a while to work that out :) not a big fan of chained ternary ops

1. why not just use the init if it's passed, I don't see the need to check if just a GET.
2. if not `weHazAuthorization` this would use `undefined`, totally ignoring `commandInit` so unauthed POSTs would not work as expected

src/ts/Types/CacheTypes.ts
```diff
@@ -0,0 +1,33 @@
+export type TPublicationTarget = {
+    webOrigin: string;
```
PeteCoward, line 2: what is webOrigin for?

src/ts/CachePublish.ts
```diff
@@ -0,0 +1,26 @@
...
+    TPublication,
+} from "./Types/CacheTypes";
+
+export class CachePublish {
+    #afc = new AppelflapConnect();
+
+    /** Get a list of published items from Appelflap */
+    publications = async (): Promise<TPublications> => {
+        const publicationList = await this.#afc.getPublications();
+        return JSON.parse(publicationList) as TPublications;
```
PeteCoward, line 14: `afc.getPublications()` does not specify type so defaults to 'json' type, so `afc.perform..` parses the return already, then you parse again here

src/ts/Types/CacheTypes.ts
PeteCoward: oh it was required in Wicher's api, i get it. Might be best to make clear it is a domain if so, not a full Uri path

src/ts/AppelflapConnect.ts
```diff
@@ -11,55 +83,221 @@ export class AppelflapConnect {
...
+    };
+
+    public getSubscriptions = async (): Promise<any> => {
+        const { commandPath } = this._commands.getSubscriptions;
+
+        return this.performCommand(commandPath);
+    };
+
+    /** Set up the Version Min and Max headers */
+    private setVersionHeaders = (
```
PeteCoward, line 221: this method actually gets headers not sets them

PeteCoward: it also totally hides and invisibly "corrects" anyone calling it wrongly

PeteCoward (changes requested):
A bit complex for my taste in places, but basically ok, but please check my reasoning for the comments and respond.
If I am correct changes are needed

src/ts/AppelflapConnect.ts
PeteCoward: `requiresAuthentication` might be better or just don't code it until you need it ;)

md8n: I put that in there precisely as a flag that we 'need' authorization.

src/ts/AppelflapConnect.ts
md8n: Some of the Appelflap returns are just simply human readable text response.  This just ensures that the text gets passed back in such cases.

src/ts/AppelflapConnect.ts
md8n: I have a fix for this, that I've been testing with in the new unit tests.

src/ts/Types/CacheTypes.ts
md8n: I stuck with Wicher's terms for the moment.  But getting these into definitions like this makes it really easy to change them later.

src/ts/CachePublish.ts
md8n: I think you may be right - in which case I have to type the return better.
I'm still working towards some of these via the unit tests

src/ts/AppelflapConnect.ts
md8n: It sets the headers for the fetch call based on what's passed in.
It is very deliberate that it takes total responsibility for these values once they've been passed to it.  I'll add more info to its JSDocs header to make that absolutely clear

PeteCoward: it returns an object that you use to set headers in a different place, but this method is badly named.
My comment about hiding bad values, was not to say that that behaviour was not documented, but that that behaviour is a bad idea.
Better to fail, so we see it.

src/ts/AppelflapConnect.ts
```diff
@@ -11,55 +84,228 @@ export class AppelflapConnect {
...
-        }
 
-        const requestInit =
-            method !== "GET" ? ({ method: method } as RequestInit) : undefined;
+        /** When we finally have authorization via Appelflap, we'll change this */
+        const weHaveAuthorization = false;
+
+        const requestInitRequired =
+            weHaveAuthorization ||
+            (commandInit && JSON.stringify(commandInit) !== '{"method":"GET"}');
```
PeteCoward, line 122: I mentioned before but the comment appear to have disappeared. What is the motivation behind trying to avoid using a RequestInit?
Why not just create an empty one if it's not passed and carry on?

md8n: because `fetch` wants the `RequestInit` as `undefined` if the anticipated set of values should be what's considered as the defaults.
So I test if there's anything that is not considered a default value and only set up a `RequestInit` in those circumstances.

src/ts/AppelflapConnect.ts
md8n: > Better to fail, so we see it.
Not in every case - I think as software developers we focus too much on failing things that we could simply and reasonably just 'fix'.

- Where we can't make reasonable assumptions - fail
- Where we can make reasonable assumptions - fix

md8n: I've changed this to throw an exception.  I don't like it, but I've done it.

src/ts/AppelflapConnect.ts
md8n: I've already made changes for this that will be in the first significant unit testing :pr:

src/ts/AppelflapConnect.ts
md8n: OK - I tested it, and fetch (at least node-fetch) doesn't care.  So I simplified the requestInitRequired test.

PR #309, opened 2020-12-18
Appelflap canoe lock

md8n:
The very first thing that Canoe needs to do if it is running under Appelflap, is tell Appelflap to not reboot it until its actually ready.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -2,6 +2,9 @@ import * as riot from "riot";
 import { installReduxPlugin } from "ReduxImpl/RiotReduxPlugin";
 import { installTranslationPlugin } from "riot/RiotTranslationPlugin";
 
+import { getPlatform } from "js/PlatformDetection";
+import { AppelflapConnect } from "ts/AppelflapConnect";
+
 import App from "RiotTags/App.riot.html";
 import "./scss/canoe.scss";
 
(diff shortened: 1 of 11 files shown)


src/riot/App.riot.html
```diff
@@ -53,10 +53,21 @@
...
+        } from "js/Events";
+        import {
+            isBrowserSupported,
+            areCompletionsReady,
+            getManifestFromStore,
+        } from "ReduxImpl/Interface";
         import { logPageView } from "js/GoogleAnalytics";
 
+        import { getPlatform } from "js/PlatformDetection";
+        import { AppelflapConnect } from "ts/AppelflapConnect";
```
PeteCoward, line 69: Why not put this in the app.js? It seems totally unrelated to interface, so can be removed from this UI file.

src/app.js
```diff
@@ -11,5 +14,34 @@ riot.install(function (component) {
...
+            riot.register("app", App);
+            riot.mount("app");
+        })
+        .catch((_) => {
+            /* Do nothing */
+        });
+} else {
+    riot.register("app", App);
+    riot.mount("app");
+}
```
PeteCoward, line 47: If it throws I'd rather not swallow it.
Using await avoids the repetition

```suggestion

    const lockResult = await lockCanoe();

    if (lockResult !== "ok") {
        // The lock didn't succeed, the app is probably still usable.
        // We don't yet understand what this means for the user and the app.
        // Should we show a small warning?
    }
}

riot.register("app", App);
riot.mount("app");
```

src/riot/App.riot.html
md8n: I tried putting this in app.js - and it definitely wants riot running first ...

md8n: But I did extract a lot of the associated code out into its own little library, and then made corresponding changes

src/app.js
md8n: This has all changed.

src/ts/CanoeHost.ts
```diff
@@ -0,0 +1,31 @@
...
+                        // The lock didn't succeed, the app is probably still usable.
+                        // We don't yet understand what this means for the user and the app.
+                        // Should we show a small warning?
+                    }
+
+                    riotStart();
+                })
+                .catch(() => {
+                    /* Do nothing */
+                });
```
PeteCoward, line 26: Why swallow exceptions like this?

PeteCoward: the whole code is effectively
```
if (inAppelflap()) {
    await this.LockCanoe();
}
riotStart();
```
And you don't even need the riot start, so don't need the await

md8n: Because we specifically do NOT want any exception here propagating, and therefore we must handle any exceptions even if that means just swallowing them.

This is a Node thing, but it's still general best practice for JS in the browser.

PeteCoward: I specifically do want to propagate an exception thrown here ( unless you are _expecting_ the exception to happen regularly )

PeteCoward (approved):

PR #312, opened 2021-01-05
Rename Resource to Asset, add state booleans

md8n:
We don't want `Resource`s in the Pages, so this renames them to `Asset`.

This also clarifies the two booleans that describe the state of an item including renaming one of them.


src/ts/Interfaces/IManifestItemState.ts
```diff
@@ -0,0 +1,14 @@
...
+export interface IManifestItemState {
+    /** Is this manifest, page or asset item available offline?
+     * @remarks This is used to indicate to the UI that a given item is ready for rendering
+     * @returns true if all items referenced by this manifest, page or asset item are available offline as well
+     */
+    isAvailableOffline: boolean;
+
+    /** Is this manifest, page or asset item complete (in itself) and ready for distribution?
+     * @remarks This is used to tell Appelflap that this individual item can be shared to other devices
+     * @returns true is this item itself is complete and the most recent that is known about amongst all the devices currently connected
```
PeteCoward, line 11: Canoe does not know about devices currently connected or what they have
```suggestion
     * @returns true is this item itself is complete - all descendant pages and assets are in the cache
```

src/ts/Interfaces/IAsset.ts
```diff
@@ -0,0 +1,11 @@
+import { IManifestItemState } from "./IManifestItemState";
+
+/** Describes a single asset, its type, and where to get it */
+export interface IAsset extends IManifestItemState {
```
PeteCoward, line 4: An asset does not have a `isPublishable` only a page. assets are cached inside page caches, only a page knows if it is complete/publishable.
It is fine to not extend from a single interface, and maybe just use an Asset.isCached member

PeteCoward: conversation resolved because the is{ublishable member was removed, but unresolved becasue it is still there through IManifestItemState.
Can you explain why you need isPublishable on an Asset?

Issue #314, opened 2021-01-06
DS Screen - Course Page


PeteCoward:
## Current state 
<img width="206" alt="Screenshot 2021-01-07 at 11 29 51" src="https://user-images.githubusercontent.com/1092029/103843750-ba625b00-50db-11eb-9a42-ad31927f2487.png">

### Previous
<img width="30" alt="Screenshot 2021-01-06 at 11 38 55" src="https://user-images.githubusercontent.com/1092029/103722667-cedb2080-5013-11eb-8218-d800e53eb298.png">

## Design system elements
<img width="190" alt="Screenshot 2021-01-06 at 11 41 12" src="https://user-images.githubusercontent.com/1092029/103722791-1eb9e780-5014-11eb-8a4a-773cb3cb58db.png">

## Required changes
- [ ] Header
  - [ ] back to courses link
  - [ ] settgins cog link ( where search icon is )
  - [ ] no image 
- [ ] Details
 - [ ] NDS consistent styling 
 - [ ] Button ( use the blue one not the purple ) 
- [ ] Lessons
  - [ ] NDS consistent styling 
  - [ ] completed card style
  - [ ] locked / unlocked status
  - [ ] progress bar
![coursepage](https://user-images.githubusercontent.com/1092029/103723434-655c1180-5015-11eb-834f-d953f2caddd8.jpg)

PeteCoward:
@katevoss @Juliana-t let me know if this kind of issue per page is useful

Juliana-t:
@PeteCoward @katevoss we can ask Daniel to prepare a default image for that banner on the top of the screen

Juliana-t:
@katevoss can we have the 'play' button to go to a lesson screen? If so, the right one is the blue circle and not this purple one

PeteCoward:
I don't see much consistency in the design for button colour. Do we know what each means? Until we do I'd go with Blue for all buttons

PeteCoward:
On the play button, just use whatever is there. If there is nothing, then no button needed here ( can be new issue )
On the header image I also recommend a new issue.

Juliana-t:
A screenshot with the 'completed' and the 'play' (to start or continue a lesson) 
![Screen Shot 2021-01-06 at 1 20 59 PM](https://user-images.githubusercontent.com/52807808/103728813-56c82700-5022-11eb-9f75-2c92ba8c639e.png)

Issue #311, opened 2021-01-05
Make images look good enough (for now)


katevoss:
We have lots of images in the redesign but letting admins upload images is beyond the scope of this leg of work.

From @PeteCoward 
Maximum I would suggest is to provide different default logos for course / lesson / section I think.
Minimum is as is the little sprout for all.

PeteCoward:
Some options for tasking and estimating improvements to this:

1. Selectable thumbnail images for courses and lessons
  - design provides X small assets for thumbnail use
  - existing content displays a single default image from that list
  - authors can select different images from the design provided images for both courses and lessons
  - these images are baked into the app code, not downloaded as content 
    - individual projects developers should be able to add/remove/delete images 
    - content authors cannot add/remove/delete images

2. Selectable thumbnail images for courses and lessons, which can be uploaded
  - existing content displays a single default image
  - authors or program staff can upload new images to an image library
  - authors can select different images from the image library to provide images for both courses and lessons
  - these images are content, need to be downloaded by the client this could impact network use
  - there is no control on image quality and suitability

katevoss:
Hi @DanielHeads, could you make four stock images with the dimensions of 45px by 45px, in SVG format? The images will be for course, lesson, resource, and activity.

Here is a general example (that doesn't have the exact dimensions):
![image](https://user-images.githubusercontent.com/2395211/103722721-92211200-4f9e-11eb-8738-98f8ea6dfb63.png)

DanielHeads:
Hi @katevoss, please find in this [link](https://drive.google.com/drive/folders/1Ysq42oNuPcafPUObg5FAm2Q35mFkcrn8?usp=sharing) the stock images in SVG. I did 6 different illustration 1 for course, 1 for lesson, and 3 for activity (content, objetive and text) If you need something more please let my know.

katevoss:
@DanielHeads these are great, thank you!

Juliana-t:
@DanielHeads illustrations are great!
I am just concerned that the blue tones used in 'Course' and 'Resource' illustrations are too close to the blue of the 'play' button. Please see the screenshot.
Maybe we can just modify colors of those 2 illustrations to lighter (or darker) tones.
![image (3)](https://user-images.githubusercontent.com/52807808/103838357-36ef3c80-50d0-11eb-8897-36993c0a7bf0.png)

DanielHeads:
[...]

PR #315, opened 2021-01-06
Course page and other updates

katevoss:
Closes https://github.com/catalpainternational/canoe/issues/314, closes https://github.com/catalpainternational/canoe/issues/311

Also closes https://github.com/catalpainternational/canoe/issues/260 and adds the Continue Learning button/card to the All Courses page. It's rough but it's the right idea, and I can easily hide it if push comes to shove.


src/scss/canoe.scss
```diff
@@ -31,25 +31,22 @@
 @import "scss/_logos";
 
 /* General */
-body {
+* {
```
katevoss, line 34: Removes the standard app margin and padding for all elements. Hopefully not too controversial.

src/scss/canoe.scss
```diff
@@ -31,25 +31,22 @@
...
     max-width: $tablet;
     margin: auto;
 }
 
-.profile,
-.settings {
-    background-color: $tan;
-}
-
+//TODO: delete these
```
katevoss, line 49: I don't think we'll need the individually-set background colors anymore. At least I hope we won't.

sandroroux (approved):

PR #340, opened 2021-01-12
Don't let workbox cache appelflap

### LINKED ISSUE #335: Stop workbox caching any Appelflap routes
## Description

All the routes used by Canoe to talk with Appelflap need to be excluded from workbox's control, otherwise confusing things are likely to happen.

md8n:
Closes #335 

# Description

To get Workbox to ignore the Appelflap routes, these routes have to be registered with Workbox, and this registration has to occur before the default handler is set with the `setDefaultHandler` method.

We use the `setDefaultHandler` method in `sw.js` therefore the route registration must take place in `sw.js` before this method.

# Solution

Because transpiling Typescript for a webworker has to be done slightly differently than for anything else, and because we don't want to muck around with multiple `tsconfig.json` files to pull this trick off, the parts of the `AppelflapConnect.ts` that the `sw.js` needs to know about so it can register the routes have been split out into a separate `RoutingAppelflap.js` file.
This lets us support registering the routes in the context of the service worker, and then later hit those routes with `AppelflapConnect`.

Unfortunately ava had a bit of a hissy fit because it couldn't find the typing file made for the `RoutingAppelflap.js` file, so there's a few big set of transpiler flags for eslint and typescript to get us over the hump for unit testing.


src/sw.js
```diff
@@ -62,6 +63,20 @@ registerRoute(
     "POST"
 );
 
+// Set up routes to Appelflap, if Canoe is not hosted by Appelflap this does nothing
+const initialiseAppelflapRoutes = () => {
+    const portNo = AppelflapPortNo();
```
PeteCoward, line 68: Isn;t appelflap port number a runtime defined variable?
If that is the case this may not work too well

PeteCoward: We could 
1) ask Wicher to change api so there is a common prefix `appelflap`
2) configure a simple regexp route `http:127.0.0.1:{\d, 4}/appelflap/`

OR ( more work )
update the default handling to network only

boerwastaken: The API prefix is currently `/api/`, it's not too late to make it something else.

@PeteCoward your regex is buggy, port number will be between 2^10 and 2^16 -1.

In both cases these form implicit dependencies which can lead to bugs.

A router could be considered which knows whether it's in Appelflap (reads the user agent) (and possibly port number) and based on that decides whether the URL is an Appeflap one and needs to be cached. If the SW is running in Appelflap and the request is for an Appelflap-URL (based on localhostness and potentially port number), don't cache.

It could also be considered to approach the problem from the other side, and become a little less Workbox trigger-happy and only route the stuff that you *know* needs to come from a Cache through an intercepted HTTP request (as opposed to being pulled straight from the Cache by frontend code that *already knows* something should be in the Cache (hey, we have a Manifest, don't we?)).

Use request interception only where you need it and so many problems simply go away.

src/js/RoutingAppelflap.js
```diff
@@ -0,0 +1,97 @@
...
+    },
+    bulkSubscribe: {
+        commandPath: `${AF_CACHE_API}/${AF_SUBSCRIPTIONS}`,
+        method: "POST",
+    },
+};
+
+export let AppelflapPort = -1;
+
+/** Get the port number that Appelflap is using or return -1 */
```
boerwastaken, line 68: Best to add a Github-link to the API docs (use a specific ref, not just "master") where the example code lives that you based this on.

src/sw.js
boerwastaken: > Isn;t appelflap port number a runtime defined variable?

Indeed, I did note in the API docs that the value is considered always stale. We don't know exactly how and when context is restored, the lifecycle of a ServiceWorker instantiation may be (or may become) different from the lifecycle of the built-in webserver.

src/ts/tests/AppelflapConnect.test.ts
```diff
@@ -55,29 +62,29 @@ test.beforeEach((t: any) => {
...
+    // const afc = t.context.afc as AppelflapConnect;
 
-    t.plan(2);
+    t.plan(1);
 
     t.is(
-        afc.portNo,
+        AppelflapPortNo(),
         t.context.testPort,
         "navigator has encoded portNo - implies Appelflap"
```
boerwastaken, line 72: with very low chance of false positives :clown_face:

src/js/RoutingAppelflap.js
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    const portNo = [...portword.substr(3)]
+        .map((el) => String.fromCharCode(el.charCodeAt(0) - 0x31))
+        .join("");
+
+    AppelflapPort = parseInt(portNo, 10);
+
+    return AppelflapPort;
+}
```
boerwastaken, line 97: Best to add a Github-link to the API docs (use a specific ref, not just "master") where the example code lives that you based this on.

md8n: Good idea - will do

src/sw.js
md8n: We could certainly change the generated routes for the Appelflap connection to use RegEx and therefore the port numbers can be handled that way.  I'd need to add some code to Regexify the various URLs to be hit.
Just setting the default handling to NetworkOnly would also help.  I don't know whether that's appropriate or not yet though ...

If the portNo does change though there's a very important consideration, that would invalidate the registered routes to Appelflap, and without restarting Canoe (or other significant actions) we'd lose what's defined for those routes.

I'm thinking that going for Option 2 (Regixifying) is the right choice for this scenario.  That removes portNo as a factor for the route registration.

md8n: And the regex for validating the port number looks like this ...
`(102[4-9]|10[3-9]\d|1[1-9]\d{2}|[2-9]\d{3}|[1-5]\d{4}|6[0-4]\d{3}|65[0-4]\d{2}|655[0-2]\d|6553[0-5])`

PeteCoward (commented):
I suspect something similar to the lines below would be sufficient.
If our code start calling out to things that false match that pattern, I'm ok with them using the network.
If we change the pattern, I'm ok with updating the static reference in the sw code

```
const appelfFlapRoute = RegExp("http:\/\/localhost:\d+\/appelflap\/");
registerRoute(appelfFlapRoute, new NetworkOnly(), "GET");
registerRoute(appelfFlapRoute, new NetworkOnly(), "POST");
```

src/sw.js
PeteCoward: > The API prefix is currently `/api/`, it's not too late to make it something else.

I think the `meta` endpoint is outside this pattern as of now

md8n: `meta` is supported.

PeteCoward (commented):
pending confirmation by @boerwastaken that the appelflap port number does not change without a reboot of the gecko view.
I'm assuming that would cause the serviceworker to rerun these route registrations.
I still prefer a simple single match for anything localhost request whose path starts with `appelflap/`

src/sw.js
boerwastaken: Are you referring to the large object storage facility of yesteryear? [That one is indeed behind /meta and /eikel](https://github.com/catalpainternational/appelflap/blob/5c6561d851477c6aaed98e5171f4799c621335b3/eekhoorn/src/main/java/org/nontrivialpursuit/eekhoorn/Const.kt#L4:L6) but I don't think using it is on the roadmap. There are no consumers anyway so those endpoints could be brought into the fold without much ado.

boerwastaken:
I've mentioned before that the port number should be considered immediately stale.
It's simple and cheap to retrieve the port number.
Retrieve it afresh every time you want to use it.
Except when you're not running in Appelflap, because that's not going to suddenly change behind the back of the runtime state for sure.
Making that work would be trivial in straight serviceworker code (`self.addEventListener('fetch', ...)`) but maybe not with those Workbox routes.


Alternatively putting the Appelflap API on `/appelflap` is fine by me except that I'll need to search and replace in the docs which I find boring but sufferable for the greater good.

src/js/RoutingAppelflap.js
```diff
@@ -0,0 +1,101 @@
+/** If Appelflap is running, it is always on localhost */
+export const AF_LOCALHOSTURI = "http://localhost";
```
boerwastaken, line 2: Better make that 127.0.0.1 . `localhost` may resolve to `::1` and I don't think we're doing IPv6 (nor do we need to). Better to sidestep having to resolve the host, avoid the risk of it returning an IPv6 `::1` and so avoid the [happy eyeball](https://en.wikipedia.org/wiki/Happy_Eyeballs) from kicking in.

boerwastaken (dismissed):
plz avoid resolving and the happy eyeballs algorithm

boerwastaken:
Note that I've just placed the whole of the API under `/appelflap` for the aforementioned greater good, so we can now live with a simple regex of `http://127\.0\.0\.1:[0-9]{4,5}/appelflap/`   without too much of a risk of false positives.

 ) for the sake of keeping the regex simple and understandable, we're not input-validating or anything, if someone comes up with invalid inputs `127.0.0.1:99999` or `127.0.0.1:0007` then that's their problem if it won't lead to joy.

please see [here](https://github.com/catalpainternational/canoe/pull/340#discussion_r556472869) to understand why I'm matching `127.0.0.1` instead of `localhost`.

please see catalpainternational/appelflap#34 for updated constants etc.

src/js/RoutingAppelflap.js
md8n: I'll have to test that, because there's translation of 127.0.0.1 to localhost that takes place behind the scenes.

boerwastaken: Likely reeolving localhost comes back with an A record for 127.0.0.1 and a quad A record for ::1.
What do you mean by "translation" and where does it live?

md8n: Tested it - it doesn't work (I had tried this before) so leaving it as 'localhost'.

boerwastaken: What are the symptoms of "doesn't work"? Are you saying it's impossible to make requests to 127.0.0.1 from an Appelflapped webapp?

md8n:
I ran into dramas while working on the next set of changes related to the manifest.  That :pr: references this one.

From the remote debugging console I could see workbox routing the first call to appelflap via `cacheFirst` and then later calls via `networkOnly`.

So my above earlier statement about `127.0.0.1` vs. `localhost` is probably incorrect, I suspect it is timing related with workbox not having properly finished registering routes before the first call to appelflap occurs. Although for some reason setting it to `localhost` is more 'reliable' than `127.0.01` for getting that first call to appelflap going via `networkOnly`.

For clarity I have moved all of the URLs that were defined in `sw.js` over to `urls.js`, and I have moved all of the actual registration of routes to `sw.js`, so now the appelflap specific route definitions in `RoutingAppelflap.js` are exactly that, just definitions.

I do recommend we stick with `localhost`, I don't recommend we spend any more time on this problem, but that we do keep a close eye on it.

PeteCoward (approved):
ok let's run with it  :)

PR #344, opened 2021-01-13
Manifest handling

md8n:
## Description

This begins the wiring in the  * new *  Manifest structure.

About the only things it doesn't do is:

- Flag the thing for publishing with Appelflap.
- Start traversing it to download the Pages that are listed within it. 
- Add the unit tests


src/index.js
```diff
@@ -42,26 +44,39 @@ subscribeToStore(() => {
...
 
+/** Sets a sessionStorage item signifying whether we are booting into a:
+ * - preseeded state (either through Appelflap cache injection, or autonomous buildup), or
+ * - blank slate.
+ * 
+ * We use the presence of the manifest as an indication for this.
+ * @remarks Can be used as a fence by awaiting its promise.
+ * @returns true when a manifest can be found, false otherwise.
+ */
+const record_bootstate = async () => {
```
PeteCoward, line 55: this function will almost definitely disappear. We don't need it if we move to a world where we just display content if we have it, having said that see comments below

src/index.js
```diff
@@ -42,26 +44,39 @@ subscribeToStore(() => {
...
+    const cacheNames = await caches.keys();
+    if (cacheNames.indexOf(MANIFEST_CACHE_NAME) === -1 || cacheNames.indexOf(MANIFESTV2_CACHE_NAME) === -1) {
+        return manifestIsExtant;
+    }
+
+    try {
+        manifestIsExtant = (await caches.open(MANIFEST_CACHE_NAME)).match(WAGTAIL_MANIFEST_URL) !== undefined;
+    } catch {
+        // Do nothing - manifestIsExtant is still false
+    }
```
PeteCoward, line 67: 1. I do not see how this code can throw an exception we would not want to immediately see
2. `(await caches.open(MANIFEST_CACHE_NAME))` is a `Cache` object,, calling `.match(..)` on it returns a `Promise` not undefined.

Like i said above this code will probably be deleted

PeteCoward (commented):
more review to come soon...

src/ts/Implementations/Manifest.ts
```diff
@@ -1,17 +1,124 @@
...
+                },
+            } as RequestInit;
+            const resp = await fetch(MANIFEST_URL, init);
+            if (!resp.ok) {
+                responseFailure = "Http error getting manifest";
+            } else {
+                return resp.json();
+            }
+        } catch {
+            responseFailure = "Network error getting manifest";
```
PeteCoward, line 101: this could happen in cases other than network error

src/index.js
md8n: I had to work with what I saw, and how it behaved up until these changes, and try not to alter that as much as possible.
So you get the above ...

src/ts/Implementations/Manifest.ts
md8n: Nope - Only actual network error would fall out here.  Everything else 404, 503, etc. gets handled by `fetch`

src/index.js
PeteCoward: as far as I can tell from reading it `(await caches.open(MANIFEST_CACHE_NAME)).match(WAGTAIL_MANIFEST_URL) !== undefined` should always evaluate to `false`
I haven't run it though, so I may be wrong :)

src/ts/Implementations/Manifest.ts
PeteCoward: there are about 10 lines of our code, and a call out to `getAuthenticationToken` and `resp.json()` that could throw an exception.

PeteCoward: well resp.json will return a promise so not get caught here, just swallowed two frames higher

src/index.js
md8n: As I said this was the previous code, with a new test added in, and then converted to its direct equivalent async / await form.

It is what it was

src/ts/Implementations/Manifest.ts
md8n: This part of it will have to be rewritten later (refactored out).
The main reason why that's not done here yet is because the 'correct' solution for such refactoring would involve pulling a LOT of code out of the JS stuff and converting it to TS, when we don't yet understand the full outcome of what these changes will require (which will mean another refactor later).
I'd prefer one refactor rather than two.
So I opted for a minimal amount of stuff moved over.

src/index.js
PeteCoward: yes the "bug" ( again it's just me reading and inferring behaviour ) is in the original code.
Does it work after you refactored it?

boerwastaken (commented):
wondering about some things, see inline comments.

src/index.js
md8n: I only saw it 'trip up' once while doing a step through, it takes a bit to set up, but yes it (seemed to) work.

src/riot/CacheDisplay.riot.html
```diff
@@ -62,7 +64,10 @@ <h4>Details</h4>
                 cachesInfo.forEach(q => {
                     switch (q.name) {
                         case 'workbox-precache-v2-http://localhost:8080/':
```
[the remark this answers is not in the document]
md8n: The only reason you're looking at this code is because I've added the `manifestv2` URL, and related bits.  I haven't changed any of the other code behaviour.
But in these cases, if there's no `q.size` then of course it will be coerced to `0`.

src/ts/AppelflapConnect.ts
```diff
@@ -16,14 +16,39 @@ import { AF_LOCALHOSTURI, APPELFLAPCOMMANDS, AppelflapPortNo } from "js/RoutingA
...
+
     private appelflapFetch = async (
         commandPath: string,
         commandInit?: RequestInit
     ): Promise<Response> => {
         const requestInfo = `${AF_LOCALHOSTURI}:${AppelflapPortNo()}/${commandPath}`;
 
-        /** When we finally have authorization via Appelflap, we'll change this */
-        const weHaveAuthorization = false;
+        const authorization = this.authHeader;
```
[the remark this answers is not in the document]
md8n: Currently this is set once and reused.  However it would be trivial to force it to refresh the credentials, and therefore the authheader.

canoe-environment-default.js
```diff
@@ -3,7 +3,8 @@
 
 module.exports = {
     GA_TAG: false,
-    API_BASE_URL: "http://localhost:8000",
+    API_BASE_URL: "http://127.0.0.1:8000",
+    NEW_MANIFEST_API_BASE_URL: "http://127.0.0.1:8000",
```
PeteCoward, line 7: this not needed surely
```suggestion
```

src/js/urls.js
```diff
@@ -1,11 +1,14 @@
 export const BACKEND_BASE_URL = `${process.env.API_BASE_URL}`;
+export const MANIFEST_BACKEND_BASE_URL = `${process.env.NEW_MANIFEST_API_BASE_URL}`;
```
PeteCoward, line 2:
```suggestion
```

src/js/urls.js
```diff
@@ -1,11 +1,14 @@
...
+export const MANIFEST_URL = `${MANIFEST_BACKEND_BASE_URL}/manifest/0.0.1`;
```
PeteCoward, line 3:
```suggestion
export const MANIFEST_URL = `${BACKEND_BASE_URL}/manifest/0.0.1`;
```

src/js/urls.js
```diff
@@ -1,11 +1,14 @@
...
 export const WAGTAIL_MANIFEST_URL = `${BACKEND_BASE_URL}/manifest`;
 export const MEDIA_PATH = "/media";
 
 export const ROUTES_FOR_REGISTRATION = {
     media: `${BACKEND_BASE_URL}/media/media/.+`,
     images: `${BACKEND_BASE_URL}/media/images/.+`,
     pagesv2: `${BACKEND_BASE_URL}/api/v2/pages/.*`,
+    manifestv2: `${MANIFEST_BACKEND_BASE_URL}/manifest/0\.0\.1`,
```
PeteCoward, line 11:
```suggestion
```

src/sw.js
```diff
@@ -13,6 +13,7 @@ import { precacheAndRoute } from "workbox-precaching";
 
 precacheAndRoute(self.__WB_MANIFEST);
 
+import { MANIFEST_CACHE_NAME, MANIFESTV2_CACHE_NAME } from "ts/Constants";
```
PeteCoward, line 16:
```suggestion
import { MANIFEST_CACHE_NAME } from "ts/Constants";
```

src/ts/Constants.ts
```diff
@@ -0,0 +1,5 @@
+/* A place to define various magic strings etc. that do not belong in urls.js */
+
+export const MANIFEST_CACHE_NAME = "manifest-cache";
+export const MANIFESTV2_CACHE_NAME = "manifestv2-cache";
```
PeteCoward, line 4:
```suggestion
```

PeteCoward (approved):

PR #351, opened 2021-01-18
Add support for credentials provided by appelflap

### LINKED ISSUE #354: Get credentials to use Appelflap API
Appelflap publishes a username / password set of data that needs to be provided by Canoe so that it can call any of the Appelflap APIs.

These values are made available in a subar subar kind of way by Appelflap.

Requirements:
[ ] - Get the credentials from Appelflap (as per its documentation)
[ ] - Process these credentials and inject them as an `Authorization` header into every request to Appelflap

md8n:
Closes #354 

This breaks out the 'use the special credentials provided by appelFlap' piece of work into its own :pr:
(It was in Manifest_Handling, but it's not directly related to Manifest Handling).

# Description
This includes the code to look for the `ecu` and `ecp` languages in Navigator, which are put there by Appelflap, and then process and use those codes to create an `Authorization` header for Canoe as it communicates with Appelflap.

# Testing
This has to be tested on device or emulator, running the latest version of Appelflap, and using the Firefox remote debugging so that you can set a breakpoint on this code and watch it step through whenever Canoe sends a request to Appelflap, which it will do at start up or restart.

### DIFF
--- a/src/ts/AppelflapConnect.ts
+++ b/src/ts/AppelflapConnect.ts
@@ -16,13 +16,11 @@ import { AF_LOCALHOSTURI, APPELFLAPCOMMANDS, AppelflapPortNo } from "js/RoutingA
 /* eslint-enable prettier/prettier */
 
 export class AppelflapConnect {
-    #authHeader = "Not set";
-
+    /** Get the Authorisation Header details that Appelflap requires
+     * @remarks See: https://github.com/catalpainternational/appelflap/blob/7a4072f8b914748563333238bb1a49ea527480bd/docs/API/determining-endpoint.md for more info
+     * @returns The auth header as `Basic btoaEncodedStuff` or 'None' if the credentials are not available
+     */
     get authHeader(): string {
-        if (this.#authHeader !== "Not set") {
-            return this.#authHeader;
-        }
-
         const credentialsExtract = (prefix: string) => {
             const rex = RegExp(`^${prefix}-[a-z]{5}$`);
             const match = navigator.languages.filter((word) =>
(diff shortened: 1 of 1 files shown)


src/ts/AppelflapConnect.ts
```diff
@@ -16,14 +16,39 @@ import { AF_LOCALHOSTURI, APPELFLAPCOMMANDS, AppelflapPortNo } from "js/RoutingA
...
+
     private appelflapFetch = async (
         commandPath: string,
         commandInit?: RequestInit
     ): Promise<Response> => {
         const requestInfo = `${AF_LOCALHOSTURI}:${AppelflapPortNo()}/${commandPath}`;
 
-        /** When we finally have authorization via Appelflap, we'll change this */
-        const weHaveAuthorization = false;
+        const authorization = this.authHeader;
```
boerwastaken, line 49: I'm unfamiliar with TS. Does this property access always result in the getter being run, or is that property somehow cached? Asking because the auth header may change over the lifetime of the JS objects in the runtime; which is why it needs to be re-derived (= getter is always run) every time.

md8n: the get is always run.  So if you want it cached, then that behaviour has to be within the get.

src/ts/AppelflapConnect.ts
```diff
@@ -16,14 +16,39 @@ import { AF_LOCALHOSTURI, APPELFLAPCOMMANDS, AppelflapPortNo } from "js/RoutingA
...
 
 export class AppelflapConnect {
+    #authHeader = "Not set";
+
+    get authHeader(): string {
+        if (this.#authHeader !== "Not set") {
+            return this.#authHeader;
+        }
+
+        const credentialsExtract = (prefix: string) => {
```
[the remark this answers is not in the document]
md8n: please check @remarks in `get` JS Docs header

src/ts/AppelflapConnect.ts
md8n: Which I've now changed so that it always gets the auth header details fresh each time.

src/ts/AppelflapConnect.ts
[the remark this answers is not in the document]
boerwastaken: 👍

boerwastaken (approved):
:+1:

PR #361, opened 2021-01-19
356 get cert signed

md8n:
# Description

Next step on from just getting the original unsigned cert from Appelflap.

This step sends that cert to the canoe-backend for signing, and it accepts the result back and tries to send it back up to Appelflap.

It does handle when the cert signing is rejected, but full and 'correct' behaviour (whatever that is) is not yet incorporated.

This also ensures that it doesn't try to get the cert signed until the user has logged into the app properly.

### DIFF
--- a/src/riot/App.riot.html
+++ b/src/riot/App.riot.html
@@ -65,7 +65,7 @@
         } from "ReduxImpl/Interface";
         import { logPageView } from "js/GoogleAnalytics";
 
-        import "ts/StartUp";
+        import { InitialiseCertChain } from "ts/StartUp";
 
         import HomePageJS from "js/HomePage";
         import CoursePageJS from "js/CoursePage";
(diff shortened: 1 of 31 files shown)


src/riot/App.riot.html
```diff
@@ -191,6 +200,19 @@
                 const page = await getPage();
                 const pageModel = getWagtailPageModelName(page);
 
+                const afc = this.state.canoeHost.appelflapConnect;
```
PeteCoward, line 203: I'd much rather keep this kind of code outside the UI components.
The app should set up the signing cert if:
in appelflap && no valid cert && logged in
The UI should have no business in that process

md8n: Partially done - The UI is still used to trigger this because it's the only thing watching the LOG_IN event currently, however I could change that

PeteCoward (approved):

PR #367, opened 2021-01-22
Reduxify Part One

PeteCoward:
This moves online/offline into redux and disables SiteDownloader 
SiteDownloader is going to get in the way of us developing  isAvailableOffline and makeAvalableOnline
The more we have in redux the better.
The less complexity in our UI the better

Commits are nicely followable


md8n (changes requested):
You need to change the name of the redux ducks `online.js` file to `Online.js` - the imports are case sensitive on Ubuntu

md8n (approved):

PR #369, opened 2021-01-25
Page Handling

md8n:
# Description

This adds an implementation for `Page` handling as part of the larger `Site` handling.

This is an interim step towards getting a reliable `isPublishable` test that examines a given page and determines if it has all of its assets as well.

This also makes changes to the Manifest implementation to keep it and the Page implementation more in sync, and to clarify some of what it does.


src/ts/Implementations/Manifest.ts
```diff
@@ -1,27 +1,32 @@
...
-    get pages(): Record<string, TPage> {
-        return this.data && this.data.pages ? this.data.pages : undefined;
+    get version(): string {
+        return this.data?.version || "";
     }
 
-    constructor() {
-        this.initialiseFromStore();
+    get pages(): Record<string, TWagtailPage> {
+        return this.data?.pages || { "0": Page.emptyItem };
```
PeteCoward, line 29: Why pretend to have a single empty page when you have no pages?
I get the change to version, that's ok. And 👍 for constructor up the top

src/ts/Implementations/Manifest.ts
```diff
@@ -81,48 +96,80 @@ export class Manifest implements IManifest {
     async getImages(): Promise<any[]> {
         const images = new Set(
             Object.values(this.data.pages)
-                .filter((page: any) => {
+                .filter((page: TPage) => {
```
PeteCoward, line 99: Why get all images from the manifest? Where is this supposed to be called?

src/ts/Implementations/Manifest.ts
```diff
@@ -81,48 +96,80 @@ export class Manifest implements IManifest {
...
+
+            // We could get false matches here, needs more work
+            if (page.loc_hash.indexOf(locationHash) === -1) {
+                return false;
             }
-        );
 
-        return matchingPages.length === 1 ? matchingPages[0][0] : "";
+            // Check the language matches
+            if (languageCode === "tet" || languageCode === "tdt") {
```
PeteCoward, line 134: please no hard coding of language codes at all!

src/ts/Implementations/Manifest.ts
md8n: The empty page is purely to meet typing requirements.  The way that `pages` is likely to get used this could save some dramas.
I'll explore an empty `Record` in the next pr though

src/ts/Render.ts
```diff
@@ -4,9 +4,10 @@
 // 3. try to fetch data if it has to ( display something while it does )
 // 4. not care if you are not logged in, if the data is present
 
-import { IManifest } from "./Interfaces/IManifest";
-import { resolveManifest } from "./Container";
-import { LoadingCallback, RenderCallback } from "./Callbacks";
+import { resolveManifest } from "ts/Container";
```
PeteCoward, line 7: I'm actually thinking this file won't be used anymore. I think it is unreferenced?

src/ts/Implementations/Manifest.ts
md8n: It was me trying to implement something like the wagtailapi code - I didn't take it back out, because it seemed it may be usable (I did reference in some test code for a while)/

src/ts/Implementations/Manifest.ts
md8n: Tell that to the people who used `tet` when they should have used `tdt` ;-)

src/ts/Render.ts
md8n: OK - I can check - but I'm still planning a 'clean up' :pr: to catch the things I missed.

src/ts/Implementations/Manifest.ts
PeteCoward: Actually the loc_hashes are guaranteed to be different for different languages, so I think there is no need for a language check here

PeteCoward (approved):

PR #371, opened 2021-01-25
Manifest refactor to assist render

PeteCoward:
as discussed

### DIFF
--- a/src/js/AuthenticationUtilities.js
+++ b/src/js/AuthenticationUtilities.js
@@ -88,10 +88,11 @@ export const isUserLoggedIn = () => {
 };
 
 export const userShouldLogin = () => {
+    const no_token = getAuthenticationToken() === null;
     const is_deauthed = localStorage.getItem(USER_IS_AUTHED_STORAGE_KEY) === "false";
     const is_firstboot = sessionStorage.getItem(EMPTY_SLATE_BOOT_KEY) === "true";
     const is_user_logged_in = isUserLoggedIn();
-    return is_deauthed || is_firstboot || !is_user_logged_in;
+    return no_token || is_deauthed || is_firstboot || !is_user_logged_in;
 };
 
 export const getUsername = () => {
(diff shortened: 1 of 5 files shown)


src/js/AuthenticationUtilities.js
```diff
@@ -88,10 +88,11 @@ export const isUserLoggedIn = () => {
 };
 
 export const userShouldLogin = () => {
+    const no_token = getAuthenticationToken() === null;
     const is_deauthed = localStorage.getItem(USER_IS_AUTHED_STORAGE_KEY) === "false";
     const is_firstboot = sessionStorage.getItem(EMPTY_SLATE_BOOT_KEY) === "true";
     const is_user_logged_in = isUserLoggedIn();
-    return is_deauthed || is_firstboot || !is_user_logged_in;
+    return no_token || is_deauthed || is_firstboot || !is_user_logged_in;
```
md8n, line 95: Had to add this in to get it to be able to get the manifest

src/ts/StartUp.ts
```diff
@@ -36,8 +36,16 @@ export const InitialiseCertChain = async (): Promise<void> => {
...
+            const hasCert = await (gt.certChain as CertChain).initialise();
+            if (!hasCert) {
+                throw new Error(
+                    "Publishing certificate chain could not be initialised. \
+                    The certificate endpoint may not be working, \
+                    Canoe is not running within Appelflap (the mobile host), \
+                    the user is not logged in \
+                    or the user may not have the correct permissions."
+                );
+            }
```
md8n, line 48: Somehow this had been rolled out, so I put it back in

md8n (approved):

src/js/AuthenticationUtilities.js
PeteCoward: you sure? backend fix for that token null got merged I think

src/ts/StartUp.ts
PeteCoward: It may be in `maifest_...` but merged in after I branched, so that it would not be in this branch until a merge with `maifest....`

PR #372, opened 2021-01-27
Wires up getting a page (for sync page)

md8n:
# Description

This wires up the riot 'sync' page and gets it to grab a page, using the new `Page` class, but still using the old page endpoint.

Obviously a lot of this code will change when the new Page api is available.  But for now this allows progress towards having a working `isPublishable` test for Pages.

### DIFF
--- a/src/ts/Implementations/Page.ts
+++ b/src/ts/Implementations/Page.ts
@@ -1,7 +1,6 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
 import {
     TAssetEntry,
-    TPage,
     TWagtailPage,
     TWagtailPageData,
 } from "ts/Types/ManifestTypes";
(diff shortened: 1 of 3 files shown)


src/ts/Implementations/Page.ts
```diff
@@ -108,12 +110,19 @@ export class Page implements TWagtailPage {
...
+            } else {
+                // Merge old and new
+                this.data = { ...this.data, ...pageData };
+            }
         }
     }
 
     async initialiseByRequest(): Promise<boolean> {
-        const resp = await fetch(this.api_url, {
+        const resp = await fetch(`${BACKEND_BASE_URL}${this.api_url}`, {
```
PeteCoward, line 125: might consider using https://developer.mozilla.org/en-US/docs/Web/API/Cache/put after this call to add to the cache ( without involving the sw )

src/ts/Implementations/Page.ts
```diff
@@ -108,12 +110,19 @@ export class Page implements TWagtailPage {
...
     initialiseFromStore(): void {
         if (this.pageId) {
-            this.data = getWagtailPageFromStore(this.pageId);
+            const pageData = getWagtailPageFromStore(this.pageId);
+            if (pageData?.api_url) {
+                // New format
+                this.data = pageData;
+            } else {
+                // Merge old and new
+                this.data = { ...this.data, ...pageData };
```
PeteCoward, line 119: Maybe consider just assigning the api return data into this.data.pageData or something like that, so you isolate what has come from where better.

PeteCoward (approved):

src/ts/Implementations/Page.ts
md8n: Good point - I'll try that in the next :pr:

src/ts/Implementations/Page.ts
md8n: At the moment they don't conflict (and are unlikely to), and this is 'better' (very subjective of course) to work with in a Typescript / Javascript sense

PR #374, opened 2021-01-28
Update Sync tag if binding

PeteCoward:


src/riot/App.riot.html
```diff
@@ -36,7 +37,6 @@
         
         <BottomMenu contentType="{state.contentType}"></BottomMenu>
```
md8n, line 38: You'll still need to fix this, but let's leave that for another :pr:

md8n (approved):

PR #377, opened 2021-01-28
Feature: Offline Sharing

PeteCoward:
This branch has the latest from our badly named https://github.com/catalpainternational/canoe/pull/288 branch, but with the FT-New-Design-System merged.


sandroroux:
@PeteCoward Should we point this at RC-2.0?

PR #378, opened 2021-01-28
Not loggedin no certchain for you

md8n:
# Description

stops trying to initialise the certchain (for publishing), until the user is in a logged in state.

PS. The user can't log in currently, we probably want to fix that.


PeteCoward (commented):
I'd prefer the initialiseAppelfap stuff to subscribe to store status, and do these things independently of everything.

md8n:
> I'd prefer the initialiseAppelfap stuff to subscribe to store status, and do these things independently of everything.

So would I, this is only about dealing with a regression introduced with the inclusion of 'guest' logins.  It needs proper fixing later, including something to delete the certs when a user logs out

PeteCoward (approved):

PR #379, opened 2021-01-29
Page is publishable second step

md8n:
# Description

This adds functionality to get Assets that are referred to by a Page. 
And then to check on the status of those Assets in determining whether a Page `isPublishable`.

To support the custom strategy upgrading workbox to v6+ was required.  That in turn required upgrading yarn.


src/ts/StartUp.ts
```diff
@@ -48,7 +48,7 @@ export const initialiseCertChain = async (): Promise<void> => {
             }
         } catch (e) {
             // No signed cert - change this to a proper message, if that is appropriate
-            console.info(e);
+            // console.info(e);
```
PeteCoward, line 51: YAY!

src/ts/Constants.ts
```diff
@@ -1,5 +1,12 @@
 /* A place to define various magic strings etc. that do not belong in urls.js */
 
+// Cache names
 export const MANIFEST_CACHE_NAME = "manifest-cache";
 export const PAGES_CACHE_NAME = "pages-cache";
```
PeteCoward, line 5: can we delete this?

md8n: yep

md8n: Still there. Let's get rid of it when we actually clean out sitedownloader and cachedisplay

src/ts/Implementations/Manifest.ts
```diff
@@ -66,91 +83,88 @@ export class Manifest implements TManifest {
...
+            this.status = "prepped";
+        }
+    }
+
+    get updatedResp(): Response {
+        return new Response(JSON.stringify(this.data));
+    }
+
+    get cacheKey(): string {
+        return this.fullUrl;
```
PeteCoward, line 131: should not be the fullUrl, but the storagecontainer name returned in the page entry in the manifest

PeteCoward (approved):

PR #381, opened 2021-02-01
More reduxy refactors

PeteCoward:


src/js/Routing.js
```diff
@@ -21,23 +20,31 @@ export function initialiseRouting() {
...
 
     // otherwise we need a manifest to understand what to render
     try {
         const manifest = await getValidManifest();
-        const page = hash.length ? manifest.getPageManifestData(hash) : manifest.getLanguageHome(getLanguage());
-        setRoute(page);
-    } catch {
-        setRoute("error");
+        var colonIndex = hash.indexOf(":");
+        let pageHash = hash, riotHash = "";
```
md8n, line 38: This style has fallen out of favour significantly, because it can't simply be 'scanned', it has to be read ... slowly.

src/js/Routing.js
```diff
@@ -21,23 +20,31 @@ export function initialiseRouting() {
...
+        if (colonIndex !== -1) {
+            pageHash = hash.substring(0, colonIndex);
+            riotHash = hash.substring(colonIndex + 1);
+        }
```
md8n, line 42: You could do all of this with a `split`

Issue #396, opened 2021-02-02
Profile (Settings) – Visual Design QA


andrewwapling:
Hi @Juliana-t ,

I have conducted visual design QA on the Canoe profile (settings) screen and have identified a range of issues that would be worth resolving from a front end implementation perspective. I have marked up detailed feedback as comments on top of the screen in question.

Please see detailed annotations of the marked up staging server: https://invis.io/N910399DW4EH

Please see reference design here: https://invis.io/4S10399MTBZQ

Hope this helps and kindly let me know if you have any queries at all.

Thanks again, Andy

PeteCoward:
I'd recommend for this screen addressing the following
- [ ] styling of labels
- [ ] styling of toggle
- [ ] styling of select
- [ ] styling of button

The overall layout should wait for the larger work item that is reorganising to include the settings as part of the profile screen.


Label / toggle / select 
<img width="205" alt="Screenshot 2021-02-03 at 16 20 37" src="https://user-images.githubusercontent.com/1092029/106712270-0d2c2580-663c-11eb-81eb-ff355f9faf5f.png">
Button
<img width="196" alt="Screenshot 2021-02-03 at 16 20 50" src="https://user-images.githubusercontent.com/1092029/106712263-0b626200-663c-11eb-8759-535c5f73e328.png">

PeteCoward:
also see my response to comments in https://projects.invisionapp.com/share/N910399DW4EH#/screens/443806506/comments

andrewwapling:
Hi @PeteCoward @Juliana-t @katevoss @megganeturner ,

Thanks very much, the FE improvements are looking good. I have marked up a range of small feedback points that would be great to action to further refine this screen.

Pleases see attached annotated PDF.

[ProfileSettings.pdf](https://github.com/catalpainternational/canoe/files/5986483/ProfileSettings.pdf)

Thanks again, Andy

PR #399, opened 2021-02-05
Profile (Settings) – Visual Design QA #396

megganeturner:
Addresses feedback in issue #396:
- labels are now semantic `<label>` tags, where they were a mix of heading and p tags. styles in line with invision spec
- updated the global `.btn-secondary` class to have inherit background-color
- custom down arrow for the select input, however as discussed the work for the custom dropdown menu will be done separately, as I suspect it would be a rather large chunk of work (:

Screen Shot 2021-02-05 at 4 37 10 pm

### DIFF
--- a/src/riot/Components/LanguageSelector.riot.html
+++ b/src/riot/Components/LanguageSelector.riot.html
@@ -1,10 +1,12 @@
 <LanguageSelector>
-    <h4>{ TRANSLATIONS.language() }</h4>
+    <label for="language">{ TRANSLATIONS.language() }</label>
 
-    <select id="language" name="language" onchange="{ setLanguage }">
-        <option value="tet" selected="{state.language=='tet'}">Tetum</option>
-        <option value="en" selected="{state.language=='en'}">English</option>
-    </select>
+    <div class="select-wrapper">
+      <select id="language" name="language" onchange="{ setLanguage }">
+          <option value="tet" selected="{state.language=='tet'}">Tetum</option>
+          <option value="en" selected="{state.language=='en'}">English</option>
+      </select>
+    </div>
 
     <script>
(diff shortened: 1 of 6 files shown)


src/riot/Components/LanguageSelector.riot.html
```diff
@@ -1,10 +1,12 @@
 <LanguageSelector>
-    <h4>{ TRANSLATIONS.language() }</h4>
+    <label for="language">{ TRANSLATIONS.language() }</label>
```
PeteCoward, line 2: suggest we change this to "Select your language" to be more consistent with design

PeteCoward: and it's a chance for @megganeturner to learn about our js translations framework.
- update the english version lower in this file
- run `yarn translations`
- ask someone who speaks tetum to provide a translation ( we usually do this all at once at the end )

katevoss (commented):
This looks good to me, though there may be further design tweaks. A note about the font size is that the Canoe DS often calls for font size 12px which is quite small for mobile devices:

The [general](https://uxdesign.cc/guide-for-designing-better-mobile-apps-typography-5796495ef86f) [recommendation](https://www.smashingmagazine.com/2011/10/16-pixels-body-copy-anything-less-costly-mistake/) is that you shouldn't use anything smaller than 16px @andrewwapling

PeteCoward (approved):

andrewwapling:
Hi @megganeturner @katevoss , 

The refinements are looking good. Please see additional commentary on visual review issue: https://github.com/catalpainternational/canoe/issues/396

Hope this is helpful and drop me a line if you have any queries at all.

Thanks again, Andy

PR #403, opened 2021-02-09
Storage container page and asset

md8n:
# Description

This makes a set of little tweaks to get both `Page` and `Asset` to use the `Page`'s storage_location for their cache name (as originally intended).

It also corrects how `Page` and `Asset` are connected using the internal `Id` within the `Asset`.  And it allows for the `Asset`'s `Id` to be a `string` or a `number` in the original manifest data.

### DIFF
--- a/src/ts/Implementations/Asset.ts
+++ b/src/ts/Implementations/Asset.ts
@@ -7,7 +7,6 @@ import { getBrowser } from "ts/PlatformDetection";
 
 // See ts/Typings for the type definitions for these imports
 import { BACKEND_BASE_URL } from "js/urls";
-import { MissingImageError } from "js/Errors";
 
 export class Asset extends PublishableItem<TAssetEntry> {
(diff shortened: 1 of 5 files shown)


PeteCoward (commented):
I'm seeing this make a weird request to `/media/` which causes a console error, not sure why it does that

md8n:
> I'm seeing this make a weird request to `/media/` which causes a console error, not sure why it does that

That occurs because it can't figure out the rendition for that asset correctly.  Have you got your canoe-backend updated properly?

md8n:
Can you make sure you have the very latest pull for this :pr: - I can't repeat that problem now, which was fixed with my second commit to this branch

PeteCoward:
Thanks @md8n I might have been running a commit behind.
See #405 for a fix for this when logged out

PeteCoward (approved):

Issue #382, opened 2021-02-02
Login – Visual Design QA


andrewwapling:
Hi @Juliana-t ,

I have conducted visual design QA on the Canoe login screen and have identified a range of issues that would be worth resolving from a front end implementation perspective. I have marked up detailed feedback as comments on top of the screen in question.

Please see detailed annotations of the marked up staging server: https://invis.io/M410394K8N3K

Please see reference design here: https://invis.io/PAZ7H36R435

Hope this helps and kindly let me know if you have any queries at all.

Thanks again, Andy

PeteCoward:
Recommend a pass touching most of Andy's feedback here.
Some questions...
- do we have button tapped interaction design?
- do we have deisgn for how the curve works on different screen sizes?

Juliana-t:
I couldn't find the design for button interactions nor for the responsiveness of the curve on the inVision project but I believe there is a DSM, I don't have access to it so we need to wait for @andrewwapling

Juliana-t:
We can move on with the review of other elements and leave this two for when Andrew gets back

katevoss:
@PeteCoward can you or @Juliana-t do a review of the comments that Andrew made and specify what changes to make for this issue, then list them here? Similar to what Peter did for the Settings changes in #396 Thank you!

andrewwapling:
Hi @PeteCoward @Juliana-t @katevoss @megganeturner,

Answers to your queries below:

1. Yes we have designed the tapped state of the buttons. The fully fledged DSM is still in progress however we have added the Primary button components please see: https://projects.invisionapp.com/dsm/catalpa/canoe-dsm

2. We have not yet tackled larger responsive breakpoints however I do believe these are on the roadmap for design including tablet and desktop.

I will review the FE refinements and post back if I see anything else worth attending to from a visual review perspective.

Thanks again, Andy

andrewwapling:
Hi @katevoss @megganeturner,

The visual refinements to login look great! The are only a few small issues that I've picked up listed below:

• Let's add the tapped interaction state for the primary button here (and globally across buttons) to add the friendly tactile feel we're aiming to achieve

• The '••••••••' in the password field appear too large let's drop these in size by about 40%

• On mobile device the fields look great on Google Chrome desktop there's a white block that appears behind the in-field labels which hides some of the inner shadow of the field itself (this issue isn't persistent in Safari). This white block is also visible behind the show / hide control. Can we check and fix this?

Let me know if you require any further details at all.

[...]

PR #404, opened 2021-02-09
382 login screen visual QA

megganeturner:
#### Addresses feedback in issue #382:
- logo size in header - increased slightly
- login page illustration size - increased slightly
- 'login' title size increased, colour changed
- made the gutter size consistent across screen widths, applied to inputs
- padding for input label
- decreased the height of input
- added margin to the top of the 'forgot my password' button
- updated the loader from the three dots to a simple spinner

#### Remaining work/questions for design:
- Andy asked for the focus state on the inputs, but I haven't been able to find an example of a focused input in the designs
- Andy also asked for a tapped interaction state on the button, but again I haven't been able to find an example of this in the designs

I'm happy to take creative license here, but also happy to wait until Andy is back to clarify these requests.

Screen Shot 2021-02-09 at 3 11 18 pm
Screen Shot 2021-02-09 at 3 11 41 pm
Screen Shot 2021-02-09 at 3 12 21 pm

### DIFF
--- a/src/scss/_loading.scss
+++ b/src/scss/_loading.scss
@@ -1,7 +1,7 @@
 LoadingDots {
   .loader {
-    margin: 1em;
-    display: inline-block;
+    margin: 2em auto;
+    text-align: center;
     width: 20px;
     height: 20px;
     border-radius: 50%;
--- a/src/scss/_loading.scss
+++ b/src/scss/_loading.scss
@@ -1,65 +1,29 @@
 LoadingDots {
-    .spinner {
-        margin: 1em auto;
-        width: 100px;
-        text-align: center;
+  .loader {
+    margin: 1em;
+    display: inline-block;
+    width: 20px;
+    height: 20px;
+    border-radius: 50%;
+    border-width: 0px;
+    border-right: 3px solid transparent;
+    animation: circ-anim 1.2s linear infinite;
+
+    &.dark {
+      border-top: 3px solid $gray-2;
     }
 
-    .spinner > div {
-        width: 17px;
-        height: 17px;
-        margin-right: 7px;
-        margin-left: 7px;
-
-        border-radius: 100%;
-        display: inline-block;
-        -webkit-animation: bounceDelay 1.4s infinite ease-in-out both;
-        animation: bounceDelay 1.4s infinite ease-in-out both;
-
-        &.dark {
-            background-color: $gray-2;
-        }
-
-        &.light {
-            background-color: $gray-7;
-        }
+    &.light {
+      border-top: 3px solid $gray-7;
     }
+  }
 
-    .app {
-        margin-top: 7em;
+  @keyframes circ-anim {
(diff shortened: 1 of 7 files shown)


src/scss/_login.scss
```diff
@@ -1,5 +1,5 @@
 Login .modal-background {
-    $content-width: 18.75em;
+    $content-width: calc(100vw - 40px);
```
katevoss, line 2: I think this change is what's causing the input fields and button to stretch too far on a wide screen. The max-width should probably be the same as the "swoop" in `.top-menu-container`
image

src/scss/_top_menu.scss
```diff
@@ -35,8 +35,9 @@ TopMenu {
 
     h4 {
         text-align: center;
-        color: $gray-3;
-        font-weight: 200;
+        color: $gray-1;
```
katevoss, line 38: You should check how this change impacts other pages, such as the lesson page:
image
The lesson page has its own Github issue so you can add a note that the course title will need to be adjusted to be smaller and lighter gray.

src/scss/_loading.scss
```diff
@@ -1,65 +1,29 @@
 LoadingDots {
-    .spinner {
-        margin: 1em auto;
-        width: 100px;
-        text-align: center;
+  .loader {
+    margin: 1em;
```
katevoss, line 3: The new spinner looks nice on the Login page but it's uncentered on other pages.

Loading All Courses:


Loading courses:

src/scss/_login.scss
```diff
@@ -154,6 +154,7 @@ Login .modal-background {
         border: none;
         padding-top: 15px;
         background-color: unset;
+        font-size: 16px;
```
katevoss, line 157: Login is one of two places where there is a `input` of `type="text"`, the other place is the `SearchBar`. Since the search bar style has not been finalized in the Design System, I think the font-size change should go in `_typography` to keep the input styles aligned (for now). Though now looking at the searchbar, I see it has its own specific font size so changing the typography of input shouldn't have any other impact.

Also we've been trying to keep font sizes in terms of `em`s rather than `px`, so this would be `1em`.

src/scss/_login.scss
megganeturner: Ah this does look bad! I hadn't tested on a big screen. Have resolved this now

Screen Shot 2021-02-10 at 12 01 33 pm

src/scss/_top_menu.scss
megganeturner: Good pickup! Have updated so that only the h4 on the login page has those changes

src/scss/_loading.scss
megganeturner: This was coming from the `display: inline-block` below, which wasn't necessary. Have fixed this up now

katevoss (approved):
Looks good!

PR #409, opened 2021-02-10
More appelflap wiring

md8n:
# Description

This PR is about telling Appelflap that a cache can be published.

It includes:
* Adding an actual `Publish` method to the `PublishableItem` super class.  So available to `Manifest`, `Page`, and `Asset` (although `Asset` can't fire it because `Asset`s always return `false` for `isPublishable`).
* Cleaning up version numbers
* Looking at different things in the Sync page


src/ts/Implementations/Asset.ts
```diff
@@ -27,6 +27,12 @@ export class Asset extends PublishableItem<TAssetEntry> {
         this.#pageCache = pageCache;
     }
 
+    /** Assets don't (yet) have a version, so we set them to a +ve value if they are `ready` */
```
PeteCoward, line 30: Assets don't need a version. If they change they will have different urls for cachebusting.

PeteCoward (approved):

src/ts/Implementations/Asset.ts
md8n: I've changed this to reflect that fact.  An Asset's `version` is a simple +/- 1 depending on whether the Asset is `ready`.

This means that all the functionality that checks whether `version` is at least greater than or equal to 0 still work.

Issue #384, opened 2021-02-02
Course Listing (Filters) – Visual Design QA and Translations update


andrewwapling:
Hi @Juliana-t ,

I have conducted visual design QA on the Canoe course listing (filters) screen and have identified a range of issues that would be worth resolving from a front end implementation perspective. I have marked up detailed feedback as comments on top of the screen in question.

Please see detailed annotations of the marked up staging server: https://invis.io/DP10395DMB8Y

Please see reference design here: https://invis.io/4G10395FNRQA

Hope this helps and kindly let me know if you have any queries at all.

Thanks again, Andy

katevoss:
I'm adding translations to this issue since it's a very small issue on its own.

andrewwapling:
Hi @katevoss @megganeturner ,

Due to the compromised implementation of the filters i.e. blend of modal filter styles and the in page graphical circular filter styles this issue would probably be best merged with 383 following: https://github.com/catalpainternational/canoe/issues/383

I have reviewed this one and added design recommendations to an annotated PDF in 383.

Hope this helps and let me know if you have any queries at all.

Thanks again, Andy

pedro16v:
Closed in favour of #393

PR #412, opened 2021-02-11
Style filters and update translations

katevoss:
Closes https://github.com/catalpainternational/canoe/issues/384

* Make the tags look like the design:

* Update translations and `yarn compilemessages`


megganeturner:
This looks great Kate! Just noting that on my end, clicking the filters doesn't add the `.selected` class, like it does in the temp build. I can't see any changes in this PR that could have caused this, so I'm not certain if something weird is going on with my build, or if I'm just missing a change ¯\_(ツ)_/¯  Either way, I'm happy to merge this and we can look at that later if need be 🙂

megganeturner (approved):

PR #414, opened 2021-02-11
389 feedback screen visual design qa

megganeturner:
Addresses feedback in #389:
- remove sun icon & add new subtitle 'Let us know how you feel'
- centre the icons in the buttons, add labels, and add active state
- tweaks to the card such as border-radius, 'x' colour, `<p>` colour

I'm still waiting on a few translations - if I get them by EOD I'll update this branch, otherwise I can add them to my next branch (which looks at the feedback confirmation screen, so closely related)

Screen Shot 2021-02-11 at 3 02 59 pm
Screen Shot 2021-02-11 at 3 03 23 pm

Issue #392, opened 2021-02-02
Cards – Visual Design QA


andrewwapling:
Hi @Juliana-t ,

I have conducted visual design QA on the Canoe card screens and have identified a range of issues that would be worth resolving from a front end implementation perspective. I have marked up detailed feedback as comments on top of the screen in question.

Please see detailed annotations of the marked up staging server: https://invis.io/FE10397XD5RQ

Please see reference design here: https://invis.io/KW1039632BJ7

Hope this helps and kindly let me know if you have any queries at all.

Thanks again, Andy

Juliana-t:
@andrewwapling please review the screenshots and add more feedback from your visual review

andrewwapling:
Hi @Juliana-t , 

No problem to add more feedback however  I couldn't see any screenshots attached. Can you please advise?

Thanks again, Andy

Issue #390, opened 2021-02-02
Feedback Confirmation – Visual Design QA


andrewwapling:
Hi @Juliana-t ,

I have conducted visual design QA on the Canoe feedback confirmation screen and have identified a range of issues that would be worth resolving from a front end implementation perspective. I have marked up detailed feedback as comments on top of the screen in question.

Please see detailed annotations of the marked up staging server: https://invis.io/MQ10396ZFSUP

Please see reference design here: https://invis.io/AW103975FPSB

Hope this helps and kindly let me know if you have any queries at all.

Thanks again, Andy

PR #418, opened 2021-02-12
390 feedback confirmation screen visual design qa

megganeturner:
Note: this was branched from #414 - probably best to get that one merged before reviewing this one.

Addresses feedback in issue #390 

- Tweaks to the heading copy
- Tweaks to lesson card styles, which I didn't realise had their own ticket until this morning's meeting (#392) - the changes in this PR just add the stack/shadow beneath the card. I suggest using #392 to move the buttons out and below the card as per designs
- update translations for feedback & feedback confirmation screens

I'm still waiting for the new image from Andy (it wasn't in the assets in invision) but will add that in once I receive it (:

Screen Shot 2021-02-12 at 3 41 58 pm


src/scss/_lesson.scss
```diff
@@ -495,7 +510,19 @@ LessonFeedback {
         .correct-flag {
             height: 6em;
             width: 6em;
-            margin: auto;
+            margin: 2em auto;
+        }
+    }
+
+    .message-container,
+    .thank-you-container {
```
katevoss, line 518: Both of these containers have very similar styles so let's use one class for all the elements. I think the only difference between the two classes is the `.thank-you-container` has the `.correct-flag` class.

src/scss/_lesson.scss
```diff
@@ -509,6 +536,23 @@ LessonFeedback {
     }
 }
 
+.message-container {
```
katevoss, line 539: I'm not sure if this was a merge kerfuffle but this seems redundant of the other rules above.

katevoss (changes requested):
Minor comments but looks good! Also some merge conflicts to fix.

katevoss (approved):

PR #421, opened 2021-02-16
392 card visual design qa

megganeturner:
Addresses feedback in issue #392 

moves the buttons out from the card for **most** cases, with two exceptions:
- exam honor code element (http://localhost:8080/#52/exam/0)
- test multiple choice 'check answer' (http://localhost:8080/#25/test/1)

These two cases are tricky because the buttons & their logic are inside sub-components. I'm not sure if it would be worth digging into these now, or if it would be better to come back to them - keen to hear your thoughts Kate!

I think I've checked everywhere else but please let me know if you find any others I've missed!

Screen Shot 2021-02-16 at 4 03 40 pm
Screen Shot 2021-02-16 at 4 03 10 pm
Screen Shot 2021-02-16 at 4 03 03 pm
Screen Shot 2021-02-16 at 4 02 46 pm
Screen Shot 2021-02-16 at 4 07 23 pm
Screen Shot 2021-02-16 at 4 03 56 pm

Cases where buttons are still within the card frame:
Screen Shot 2021-02-16 at 4 07 06 pm
Screen Shot 2021-02-16 at 4 04 43 pm


katevoss:
I think you made smart choices to leave the two cards-with-buttons as they are for now 👍🏻

katevoss (approved):
I noticed the spinner on the button is a little low, is that related to this PR? If not I'll make a separate issue for it


Other than that, looks good!

megganeturner:
Good pickup! I've fixed this now :) 

Screen Shot 2021-02-17 at 2 31 20 pm

PR #424, opened 2021-02-17
Manifest status

md8n:
# Description

Basic purpose is to provide a way to subscribe to status changes to the Manifest, a Page or an Asset related to their storage in the cache or in the redux store.  This is done by saving that `Status` into the redux store itself, then other parts of the code base can subscribe to the redux notifications and filter for the `Status` they are interested in.

## Clean up / Refactor
* Removed the `.source` member - didn't do anything useful
* Split some of the `Interfaces` and `Types` into separate, more manageable files.

## Added `Status` object
* Relevant redux bits and pieces (the reducer itself is in TS)
* Improved distinction between what's happening with the cache and what's going on with the redux store.
* Used `set`ters to ensure that changes to the `Status`' status values propagate through to the redux store.

## Things explored
* Tried adding the other status values `isValid`, `isAvailableOffline` and `isPublishable` to the `Status` object, however this would require a much more significant refactor so decided against it.

## Things not tried
* Did not try rewriting the code for Manifest, Page or Asset to get them to subscribe to `Status` changes in order to trigger their behaviours - I figured that this refactor was enough.
* Did not try getting any of the riot tags to show the `Status` changes.


PeteCoward:
@md8n I think this is causing the ready status to be unready while making a page request, even when the data is in the store, so the page is technically ready

md8n:
Not quite - that was a choice I made for that output on the sync page.
The fact is, that if you refresh the sync page directly, then the data will not by in the Redux Store, but of course it will be in the cache.
So the output on the sync page is deliberately showing the store status *before it then goes and fills it from the cache*

PeteCoward (approved):

Issue #394, opened 2021-02-02
Resources Detail – Visual Design QA


andrewwapling:
Hi @Juliana-t ,

I have conducted visual design QA on the Canoe resources detail screens and have identified a range of issues that would be worth resolving from a front end implementation perspective. I have marked up detailed feedback as comments on top of the screen in question.

Please see detailed annotations of the marked up staging server: https://invis.io/Z410398MR3N8

Please see reference design here: https://invis.io/G9108GUJAMZR

Hope this helps and kindly let me know if you have any queries at all.

Thanks again, Andy

PeteCoward:
Recommend we implement
- [ ] header bar as per design
<img width="192" alt="Screenshot 2021-02-03 at 16 48 00" src="https://user-images.githubusercontent.com/1092029/106714900-d1935a80-663f-11eb-95b1-3973dc433200.png">

We have to discuss the rest, what resources are in Canoe is a series of content cards ( like lesson content )
We can implement the Basic card display to look like the element below ( ignoring the scribbled out bits ), but we'd need to render all cards in a vertical scroll layout.
<img width="216" alt="Screenshot 2021-02-03 at 16 51 32" src="https://user-images.githubusercontent.com/1092029/106715473-73b34280-6640-11eb-94b4-be1cbb2360c9.png">

Juliana-t:
this is the correct reference screen > https://projects.invisionapp.com/share/TBYZBD5NFRC#/screens/429853510

andrewwapling:
Hi @PeteCoward @megganeturner @katevoss , 

This sounds like a good interim solution. I have also updated the issue with the correct screen design reference.

Let me know if there's any design queries or if I can help with anything else on this one.

Thanks again, Andy

PR #428, opened 2021-02-19
394 resources detail

megganeturner:
Addresses feedback in issue #394

Primary changes:
- passing a `noCardStyle` prop to the `LessonFrame` component to prevent resource page from inheriting card styles
- render all of the resource cards in one screen, where recently we were only rendering the first
- adding title to `topmenu` and back arrow instead of close cross

Smaller changes:
- switch the order of the image & heading on a `BasicCard` (this is consistent with the [newer designs for lessons](https://projects.invisionapp.com/d/main/#/console/20251865/430145918/inspect)) - also make the image full width
- tweaks to quote style (note that [the designs](https://projects.invisionapp.com/d/main/#/console/20251865/429853509/inspect) include a nicer quotation mark - InVision claims this is what quote marks look like in Tiempos Headline, but it's lying. I've changed the font for that span to be `cursive`, as I couldn't find any other websafe font with that style of quotation, and I figured that would be better than importing a whole other font just for one character - happy to discuss this though!
- bullets for lists are no longer the blue accent color, and are gray-1. also, the wagtail `listcard` list items (which come through as `<p>` tags) are now styled more similarly to our native `<li>` tags
- remove code to prevent box shadow appearing on completed cards (this is a small change David requested in our sprint planning call today - figured it would be easiest to implement in this one)

Course # 112 has all available cards (my url is http://localhost:8080/#112)

A thought... now that we are including the description on the course listings page, and it is truncated, we should probably also include that description on the actual resource detail page. Something to implement in the future (when we implement the tags and other features in the design).

localhost_8080_(iPhone 6_7_8)


src/riot/Components/QuoteCard.riot.html
```diff
@@ -1,10 +1,11 @@
 <QuoteCard>
+    <hr>
```
katevoss, line 2: This might need some work, it looks a bit odd in the normal Lesson Content cards but it's not clear from the mockups how it should look.
image
I think it's fine for now

src/scss/_lesson.scss
```diff
@@ -97,6 +97,41 @@ LessonFrame {
...
+        max-width: $tablet;
+
+        @media only screen and (min-width: $tablet) {
+            margin: 0 auto 1.5em;
+        }
+
+        ImageWrapper img.richtext-image {
+            width: calc(100% + (2 * 18px);
+            margin-left: -18px;
+            margin-right: -18px;
```
katevoss, line 112: Ah full bleed images, I've been meaning to try out the grid method as [described in this blog post](https://www.joshwcomeau.com/css/full-bleed/). If you care to look into it, be my guest. Otherwise negative margins get the job done 🤷🏻‍♀️

src/scss/_lesson.scss
```diff
@@ -97,6 +97,41 @@ LessonFrame {
...
+        }
+    }
+
+    .card-content > * + * {
```
katevoss, line 116: This is a bit cryptic, is there a simpler or specific way of writing this?

katevoss (approved):
Totally agree with your decision not to add a font just for one character. Sometimes InVision does lie! Would you add a comment on InVision to see if Andy can export it as an icon? If that would be easier than tracking down the correct font. What you have looks good for now 👍🏻 

Good thought about the showing the resource description in full, that's also worth a comment on InVision I think.

Overall I have a few comments but in the interest of time any changes can go into future PRs. Looks good!

src/scss/_lesson.scss
megganeturner: Yepp good point! I'll add a class and sneak this change into my next PR

PR #431, opened 2021-02-19
Show statuses in sync

md8n:
# Description

In the Sync page, this now shows a listing of the manifest and all pages defined in the manifest, and their respective statuses.

Also the `Sync Me` button will attempt to 'publish' all items that have the `isPublishable` status (i.e. it will tell AppelFlap to publish them).

The `fetchingManifest` redux action and related pieces has been removed because it is now surplus to requirements.

Several types have been renamed so they aren't so 'offensive' to Peter (and me for that matter).  They've also been split in different ways to make them more useful.

## Note:
The `PublishAll` function is in with the AppDataStatus class - I don't like it here.  Once we have the subscribe functionality available I strongly recommend that we break it all out into its own module.


PeteCoward (commented):
I'll run it later and if it runs approve

PeteCoward (approved):

PR #433, opened 2021-02-22
Sync all

### LINKED ISSUE #353: Canoe decide whether to publish or subscribe to the manifest
towards #300 

Do we have a manifest or not?

md8n:
Closes #353 

# Description

This calls all of the Appelflap endpoints related to publishing and/or subscribing a single item.

This includes unlocking the cache first (delete the insertion-lock) and then locking the cache again once completed.

There is an issue related to Appelflap integration, which is not the responsibility of this :pr:.
* subscription endpoints not yet implemented in Appelflap (404s returned)

Before doing the unlock mentioned above calling 'sync me' a second time would result in the following error.
* calling publish a second time for an item, after a refresh of the app via geckoview (get a 500). e.g. trying to publish the manifest a second time gets this error back from Appelflap `java.io.FileNotFoundException: /data/user/0/io.catalpa.canoe.demo/files/mozilla/default_profile/storage/default/http+++localhost+8080/cache/morgue/83/{5c9fb8e3-8361-48a8-a82d-e48716adf453}.final (No such file or directory)`

This :pr: 'handles' any issues such as the above (as far as the UI is concerned).

The UI for this is also very crude.


src/ts/AppDataStatus.ts
```diff
@@ -127,35 +141,121 @@ export class AppDataStatus {
...
-        const publishable = this.itemListings.filter(
-            (listing) => listing.isPublishable
-        );
+        const performable = this.itemListings.filter(filter);
 
-        const published: Record<string, boolean> = {};
-        for (let ix = 0; ix < publishable.length; ix++) {
-            const item = publishable[ix];
+        const performed: Record<string, TAppelflapResult> = {};
+        for (let ix = 0; ix < performable.length; ix++) {
```
boerwastaken, line 152: I'm not familiar with TS but traversing list/array-like structures using a for loop with an incrementing counter doesn't look very this-century? Is there no `for thing in collection` type idiom available?

src/riot/Screens/Sync.riot.html
```diff
@@ -102,10 +102,15 @@ <h5>ID:{statusListing.id} {statusListing.type}:{statusListing.title} <small>({st
             },
 
             async syncMe() {
+                const cacheUtilities = new CacheUtilities(appelflapConnect);
+                await cacheUtilities.unlock();
+
                 // Identify what needs to be subscribed to, and what needs to be published
                 const appDataStatus = this.state.appDataStatus;
                 const syncStatus = await appDataStatus.SyncAll();
```
boerwastaken, line 111: I don't think you want to be interrupted while doing your state changing operations, yet above you've called unlock() just before embarking upon state change endeavours, that's maybe not the best way to go about things.

To me it seems that you should do your subscription poking and publication poking on an ongoing basis, in the background. Packups do not interrupt anything (they may fail if you're using/modifying the cache, but they should not affect the user's experience).

Then IIRC it was decided that for now, Canoe would have a separate page where the lock is released and things like webapp reloads may happen. On that page you should basically do nothing, right?

src/ts/AppDataStatus.ts
```diff
@@ -127,35 +141,121 @@ export class AppDataStatus {
...
             if (item.type === "manifest") {
                 try {
-                    published["manifest"] = await this.manifest.publish(
+                    performed[this.manifest.cacheKey] = await action(
+                        this.manifest,
                         appelflapConnect
                     );
                 } catch {
-                    published["manifest"] = false;
+                    performed[this.manifest.cacheKey] = "failed";
```
boerwastaken, line 161: "the operation did not complete, because of an error" would be the Microsoft style, but if we're in charge, why hide what transpired? I'd say attach the exception too

src/ts/AppDataStatus.ts
md8n: This is a JS issue not a TS issue.  Anytime you use a `for thing in collection` type idiom you create a new scope.  Which is OK so long as you don't have an `await` within that scope.
So, if you need to `await` several things in some `enumerable`-ish thing, then it's good ol-fashioned `for` loops to the rescue to keep the scope the same.

src/riot/Screens/Sync.riot.html
md8n: This is that page

src/ts/AppDataStatus.ts
md8n: At this time this is really just providing some very basic feedback to the user (i.e. a dev) about what did (or didn't) succeed.
What a 'real' user gets to see in the future is still up in the air.

src/ts/AppDataStatus.ts
boerwastaken: Wow. And is there no kind syntactic sugar as an alternative to such for-loops?

src/riot/Screens/Sync.riot.html
md8n: Rather than attach the behaviour to entering the page itself.  I attached it to clicking a button, but otherwise it's still the same idea, by performing some action, the user indicates that they're ready for any update.

src/ts/AppDataStatus.ts
md8n: Nope

PeteCoward (approved):

PR #439, opened 2021-02-23
Handle page initialisation errors better

md8n:
# Description

If a page's cached entry is incorrectly formatted the following line from page's `initialiseFromResponse` will throw an error.
```
this.data = await resp.json();
```

However, the calling method expects a boolean result instead to indicate any problems (i.e. `false`).

This fixes that failure and returns the correct result.

## Note:
It may be desirable to change this line in `routing.js`
```
page.initialiseByRequest();
```

Because it will now reliably get a boolean indicating whether the page has been successfully initialised or not. e.g.
```
const isInitialised = await page.initialiseByRequest();
```

### DIFF
--- a/src/ts/Implementations/Page.ts
+++ b/src/ts/Implementations/Page.ts
@@ -235,7 +235,12 @@ export class Page extends PublishableItem<TWagtailPageData> {
 
     async initialiseFromResponse(resp: Response): Promise<boolean> {
         this.status.cacheStatus = "loading";
-        this.data = await resp.json();
+        try {
+            this.data = await resp.json();
+        } catch {
+            // We assume a bad cached entry, i.e. bad json
+            return false;
+        }
 
         // And store the page data in Redux
         this.StoreDataToStore();


PeteCoward (approved):

PeteCoward:
> const isInitialised = await page.initialiseByRequest();

That would be an unused variable failure if Routing was typescript no?
that line needs to change really as it should 
1. only use a request if the data is not in the store
2. use cached data preferentially to network

We should check `isInitialised` by querying the page object ( indirectly this should query the store )

md8n:
> That would be an unused variable failure if Routing was typescript no?

Correct - This was just a suggestion for that particular line of code.  How it should be handled after that I think needs further discussion.

Having said that I believe that this line should be `await`ed regardless.

PeteCoward:
> Having said that I believe that this line should be `await`ed regardless.

reasons?

PeteCoward:
> needs further discussion.

Do you have anything to add to
> that line... it should
> only use a request if the data is not in the store
> use cached data preferentially to network

md8n:
> reasons?

Because otherwise `routing.js` falls immediately through to routing to the given page. It should be given a chance to actually load the data and thereby effectively verify that the page exists in a usable form.

md8n:
> Do you have anything to add to
> 
> > that line... it should
> > only use a request if the data is not in the store
> > use cached data preferentially to network

Kind of - here's the chain of events that leads to this problem relative to `routing.js`:
* Initially we try `initialiseFromCache` for that page.
* We get a `response` from the cache for that page.
* But that `response` will not parse into the `.data` member for the `page` object (why? still sorting that out, but candidates include that the originally stored network response was in fact an error (so not json, but text), or that there's something really dodgy with the json, or that there's been some significant change in the json format, and the cached entry is the old format)

Now with this fix `routing.js` gets a simple boolean telling it that the page couldn't be initialised from the cache, and it should do a full network request instead.

PeteCoward:
Its a question of whtehr that fallback logic lives in routing, or in Page, or in a seperate module I think

Issue #443, opened 2021-02-23
Resources page


pedro16v:
# Overview 
Following a design review meeting, here's some fixes that would be great to get in as we transition to Release One work.

* Check the meeting's [InVision Freehand ](https://projects.invisionapp.com/freehand/document/ZUJn1e00N)
* Ping @andrewwapling in the comments if you need any further clarification

# Notes

## Listing
- Some font sizes look a bit smaller. Try:
  - [x] Line 1: 0.7 em
  - [x] Line 2: 1em
  - line 3 is fine
  - [x] Reduce padding/margin if increasing the font sizes makes things too large
- [x] Make search field taller
- [x] Move search icon to the right

## Resource detail
- [x] Header container: #F4F8FA
- [x] Background colour (should be lighter): #FAFDFF

Issue #440, opened 2021-02-23
Login screen


pedro16v:
# Overview 
Following a design review meeting, here's some fixes that would be great to get in as we transition to Release One work.

* Check the meeting's [InVision Freehand ](https://projects.invisionapp.com/freehand/document/ZUJn1e00N)
* Ping @andrewwapling in the comments if you need any further clarification

# Notes

- [x] Make fields taller
- [x] A bit more padding on the fields (45px)
- [x] A bit more padding on the button (40 to 45px)
- [x] Push the fields down a bit

PR #445, opened 2021-02-23
440 login further tweaks

megganeturner:
Addresses feedback in issue #440 (& a few additional comments from issue #382):
- fixes the white background on autofilled inputs with error (the background isn't quite the perfect colour as it's coming from a box-shadow, but this is about as close as I can get)
- increase input size/padding
- decreases the size of the password dots '••••••••'
- more padding on button
- more space above the form

also a small tweak to the resource detail page to update the background-color as per #443 

Screen Shot 2021-02-23 at 4 00 02 pm
Screen Shot 2021-02-23 at 3 59 41 pm

### DIFF
--- a/src/scss/_lesson.scss
+++ b/src/scss/_lesson.scss
@@ -95,11 +95,12 @@ LessonFrame {
 
 ResourceArticle {
     .no-card {
-        margin: 0 18px 70px;
+        padding: 0 18px 70px;
         max-width: $tablet;
+        background-color: $gray-10;
 
         @media only screen and (min-width: $tablet) {
-            margin: 0 auto 1.5em;
+            padding: 0 auto 1.5em;
         }
 
         ImageWrapper img.richtext-image {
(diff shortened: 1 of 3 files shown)


src/scss/_lesson.scss
```diff
@@ -95,11 +95,13 @@ LessonFrame {
...
     .no-card {
-        margin: 0 18px 70px;
+        padding: 0 18px 70px;
         max-width: $tablet;
+        background-color: $gray-10;
 
         @media only screen and (min-width: $tablet) {
-            margin: 0 auto 1.5em;
+            padding-bottom: 1.5em;
+            margin: 0 auto;
```
katevoss, line 104: I think this change causes the content to be hidden by the bottom menu on wide screens, tablet and wider.

katevoss (changes requested):
A couple tweaks but generally look good!

katevoss (approved):

PR #446, opened 2021-02-23
Refactor after sync all - separation of concerns

md8n:
# Description

A refactor of the code that works with both the cache and items (manifest, asset, and page).  There has been NO change in functionality except what was required for the refactor itself.

The `Sync All` branch is a convenient point in the code base to do a refactor around the cache behaviour.  This also sets up for better testing of the cache related functionality.

This splits out that functionality into a separate module, with the function signatures inverted to take an `item` as input and perform cache related actions on that `item`.

Matching changes are made in `routing.js`

Also, methods flagged as deprecated in `routing.js`, and that are not referenced anywhere else have been deleted, with corresponding cleanup in `Errors.js` and `WagTailPagesAPI.js`


src/js/Routing.js
```diff
@@ -17,10 +12,6 @@ const CANOE_SHORTCUTS = {
     'home': 'homepage',
     'resources': 'resourcesroot',
 };
-// pages from the CMS ( we might not use this )
-const IS_WAGTAIL_PAGE = /([\d]+)/; // should match '#3' and '#3/objectives'
-
-const IS_PAGE_PREVIEW = /^\?(.+)/;
```
PeteCoward, line 23: I'd rather you didn't remove things to do with the preview functionality, as it still needs to be re-implemented, leaving it in place makes that process easier

PeteCoward: It's ok, though, we can just go looking in other branches for it

src/js/WagtailPagesAPI.js
```diff
@@ -17,11 +17,6 @@ export async function fetchManifest() {
     return allPagesMetadata;
 }
 
-export async function fetchPageNoAuth(path) {
```
PeteCoward, line 20: this was added to the FT-New-Redesign branch to fix preview, and snuck in here with a merge. Why did you remove it?

src/js/Routing.js
md8n: I'll put it back - thanks for the heads up

src/js/WagtailPagesAPI.js
md8n: Because nothing in the client code referenced it - I'll put it back.

PeteCoward (approved):

PR #449, opened 2021-02-23
Optional Service Workers with Refresh to Load Changes

sandroroux:
Makes the service worker optional. Removes the loading screens and sw-related Redux code.

https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Add_to_home_screen


src/js/ServiceWorkerManagement.js
```diff
@@ -1,49 +1,64 @@
...
-        changeServiceWorkerState("controlling")
-
-        function checkForNewVersion() {
-            console.log("checking for new version");
-            if( wb ) {
-                wb.update();
+    wb.addEventListener("installed", (event) => {
+        if (event.isUpdate) {
+            const refreshForNewVersionMessage = gettext(
+                `New content is available! Click OK to refresh.`
```
PeteCoward, line 52: it's not new content, it's a new application versions

src/index.js
```diff
@@ -1,83 +1,13 @@
...
-    // indicate 'empty'
-    sessionStorage.setItem(EMPTY_SLATE_BOOT_KEY, !manifestIsExtant);
-
-    // indicate we've got everything
-    return manifestIsExtant;
-};
-
 initializeServiceWorker();
+
+import(/* webpackChunkName: "app" */ "./app.js");
```
PeteCoward, line 13: So, the main reason for this dynamic webpack chunked import and requiring of the service worker was to deal with the following wierd use of internet
1. User visits page for the first time, downloads the entire webpack compiled js codebase
2. User registers a service worker, the service worker then re-downloads the entire webpack compiled js codebase
3. User starts using the serviceworker and everything is fine.

Check in your devtools network tab. Before the browser was parallel downloading the app codeat the same time becasue they are both being requested by window scope and the serviceworker.
scope.
Does this happen again with this change?

src/js/ServiceWorkerManagement.js
sandroroux: Good point. I'll change this.

src/index.js
sandroroux: Yeah, here's what I think I see:

On a fresh install (after deleting site data), 
* `canoe.js` downloads at ~2.7 mb.
* `canoe.js` code-split imports app.js, which is ~1.2 mb.
* Precache downloads, which is ~1 mb.

I refresh. Everything loads from the service worker cache.

sandroroux: Here's the first load:
image

And the sw-starting refresh:
image

src/js/ServiceWorkerManagement.js
```diff
@@ -1,65 +1,69 @@
...
-        // This happens:
-        // a fresh first time visit once the sw is in control
-        // a refresh when there is an active sw to take control
-        changeServiceWorkerState("controlling")
-
-        function checkForNewVersion() {
-            console.log("checking for new version");
-            if( wb ) {
-                wb.update();
+    register("/sw.js", {
```
sandroroux, line 31: As an experiment, I switched to [`register-service-worker`](https://www.npmjs.com/package/register-service-worker), which matches [`workbox-window`](https://www.npmjs.com/package/workbox-window)'s functionality, but at a much smaller size (21 kB vs. 577 kB).

Vue uses `register-service-worker` in their `pwa` plugin. Workbox Window [might have more features](https://dev.to/webmaxru/workbox-4-implementing-refresh-to-update-version-flow-using-the-workbox-window-module-4e3c), but I don't think we need that flexibility.

PeteCoward: we need the range requests plugin that workbox provides, until we put in the work to get that capability back in other ways we need workbox I think

sandroroux:
@PeteCoward It's worth discussing the scale of our current service worker code. This branch reacts to the current implementation: that it is specialized, big, and complex.

Here, I try to simplify our implementation (no redux, no workbox-window, no need for preinstallation) and follow standards we see in other sites (e.g. cmd+R to run without the worker, prompt user when new versions are available). Canoe is getting big, so any simplification helps.

What do you think?

PeteCoward:
I like the reduction in size, but we have code that assumes we have a service worker controlling ( is available offline etc )
I like the being able to run without sw ( mainly for us ) but not the removal of knowing that in the redux state ( so code can run differently if controlled by sw or not )

sandroroux:
@PeteCoward Ah, understood. Is there any chance that code can be service-worker-detecting? As in, those elements and libraries only activate when an offline method is present.

PeteCoward:
I'm coming round to this, the benefits for us of the devtools just working as expected ( bypass etc ) are really valuable.
We can use the navigator interface to detect if we have the serviceworker behind us at any point, it's just a bit of a pain.

It needs some really careful testing though. I think one of the things workbox is supposed to do is smooth out cross browser oddities. Maybe I just imagine remembering reading that though :)

src/index.js
PeteCoward: > I'm having trouble reproducing it. If I'm following your instructions well, here's what I see:
> 
> On a fresh install (after deleting site data),
> 
> * `canoe.js` downloads at ~2.7 mb.
> * `canoe.js` code-split imports `app.js`, which is ~1.2 mb.
> * Precache downloads, which is ~1 mb.
> 
> I refresh. Everything loads from the service worker cache.

My point with this is that the precache download includes both the canoe.js and the app.js, double downloading. But I'm beginning to think that one time install inefficency is just not important enough to worry about

PeteCoward:
Making this draft, lets keep it around for discussion

PR #475, opened 2021-02-25
Override support

PeteCoward:
This seems to allow overriding of the svg referenced in the handlebars partial
Please test, also would be nice to check if we can easily just override the partial itself


katevoss (approved):
Success! I'll leave this PR open because it's still a draft but I think it can be merged.

webpack.config.js
```diff
@@ -207,5 +210,6 @@ module.exports = env => {
         environmentConfiguration.WEBPACK_CONFIG,
         productionWebpackConfig
     );
+    console.log(config);
```
PeteCoward, line 213:
```suggestion
```

PR #476, opened 2021-02-25
Use an svg icon which resizes better

PeteCoward:
This should resolve the irritating console error telling you the android 512 icon is the wrong size. It was the wrong size because the webpack-pwa-manifest plugin which promises to auto resize the necessary icons, does not do so when the input is a .ico file.

Not that because we normally build from the canoe parent project, this requires a similar change in that project's `canoe-project.js` config file, see the "mentioned" PR below


katevoss (commented):
The error is gone, that's really nice.

Is there any reason to keep the favicon.ico? If not then let's remove it.

Also I ran favicon.svg through the [SVG compressor](https://vecta.io/nano) and updated it.

megganeturner (approved):
Nice Peter! I'll leave this open in case you'd like to remove the `.ico` as Kate mentioned. Very nice to have one less thing in the console 😄

PR #477, opened 2021-02-25
Support for skipping the serviceworker installation

PeteCoward:
This allows a developer to avoid the painful refresh/wait/refresh/wait cycle bypassing service worker requirement.
It should not be used in production, or when testing serviceworker related features.

Edit your project level `canoe-environment.js` file and add this.
```javascript
    SKIP_SW: true,
```

### DIFF
--- a/canoe-environment-default.js
+++ b/canoe-environment-default.js
@@ -4,6 +4,7 @@
 module.exports = {
     GA_TAG: false,
     API_BASE_URL: "http://localhost:8000",
+    SKIP_SW: false,
     DONT_SHOW_COMPLETIONS_AFTER: "2020-2-15",
     APPLICATION_SERVER_KEY: "",
     WEBPACK_CONFIG: {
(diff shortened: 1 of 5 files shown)


katevoss:
This is an improvement in the service worker doesn't load 🎉 
However, after every local change the page refreshes and I have to login again. So overall an improvement but not quite desirable behavior.

sandroroux:
@PeteCoward How's this supposed to play with my PR?
https://github.com/catalpainternational/canoe/pull/449

PeteCoward:
> #449

It's supposed to just be a quick easy way to help devs dev easier, not to change any site behaviour.

PeteCoward:
> However, after every local change the page refreshes and I have to login again. So overall an improvement but not quite desirable behavior.

silly me, I only tested on the login page :facepalm:

PeteCoward:
Try it now @katevoss and @megganeturner

megganeturner (approved):
This works really nicely, thanks Peter!

PR #478, opened 2021-02-25
New subscribe API

md8n:
# Description

Uses the new Appelflap Subscription API, and corrects some terms.

There is still a unit test to fix.  Plus get the subscription watching the correct files (instead of everything).


boerwastaken:
I think the proof of the pudding will be in the eating of it - did you run it against Appelflap?

One thing that I'm missing is some kind of concurrency control. Appelflap supports the `If-Match` HTTP client header for `/subscriptions` so that you can make sure you're editing the latest known state and not some older copy. Think of it as how in CouchDB you need to pass the revision of the latest incarnation of a document in order to update it (thus making you testify that you have "seen" all other edits).

There are cases in which using that won't be necessary, three spring to mind

1. you always build the document completely from scratch, as opposed to editing the current state (but then how do you know the value of `injected_version` ?)
2. you have some kind of mutex somewhere that serializes concurrent edits
3. you know utterly sure that nothing will ever ever ever result in your api-adressing functions being called concurrently, and if I ask any random Catalpa engineer "will `setSubscriptions` ever be called concurrently" they'll answer "no of course not, because this and that reason, and see right here where Lee documented not to do that, and the reasons not to use straight HTTP primitives for this are X, Y and Z".

The robust, cautious, and unsurprising approach would be to simply implement the edit conflict prevention mechanism (basically, pass the Etag you got when GETing the doc into an If-Match header when POSTing it).

md8n:
> did you run it against Appelflap?

I only ran it against Appelflap - on an actual device.

As for the rest of your comments, yep they're all acknowledged, in the sense that I'm not currently concerned with concurrency (although I should be).  Currently how it works is that this behaviour is only accessible via the Sync page, and only when the user clicks the 'Sync Me' button, which does what it does in a synchronous matter making the user wait for the result.  So with the way things are now, it's fine.  But I'd prefer this to be something that wll ultimately be able to work 'in the background', so the problem will have to be resolved then.

boerwastaken:
> So with the way things are now, it's fine.

*fine-ish* ;-)

I'll put this up on canoe-engineering to see how things go — I'm passing this on to @PeteCoward for review, as I only checked some high-level things that I was concerned with.

boerwastaken:
Hmmm I can't find the "sync page" on https://learn.canoe-engineering.temp.build — where is it?

md8n:
> Hmmm I can't find the "sync page" on https://learn.canoe-engineering.temp.build — where is it?

You have to run this on an android device to see that (or an emulator of course) - It will show up as a fourth option on the bottom 'menu'.

PeteCoward (changes requested):
> `yarn test`
```FetchError {
    message: 'invalid json response body at  reason: Unexpected token o in JSON at position 0',
    type: 'invalid-json',
  }```

boerwastaken:
>>  Hmmm I can't find the "sync page" on https://learn.canoe-engineering.temp.build — where is it?

> You have to run this on an android device to see that (or an emulator of course) - It will show up as a fourth option on the bottom 'menu'.

Of course! Got it

boerwastaken:
nocanhaverelease

I'm on the sync page as shown above.
I have a subscription which lists as injectable a cache identity for which I happen to have a bundle ready to go, as testified to by `/ingeblikt/publications/` and `/ingeblikt/subscriptions/`.

A few remarks:

1. What does "Getting" signify... who's getting what?
2. What does "No syncs" signify? It sounds like a countable thing. What is the definition of a "sync"?
3. The insertion lock is acquired, but not released — Appelflap can't do the injection that it knows it should perform :disappointed:

boerwastaken:
4. It looks like it GETs to `/eikel-meta/status` - for Canoe which isn't currently using the large object storage facility, I've removed those HTTP handlers, so this 404s. What's the use case for any information that could formerly be gleaned from there?

md8n:
> 1. It looks like it GETs to `/eikel-meta/status` - for Canoe which isn't currently using the large object storage facility, I've removed those HTTP handlers, so this 404s. What's the use case for any information that could formerly be gleaned from there?

Because that was one of the first APIs that you implemented that I could get a response out of.

PeteCoward (approved):
OK, happy to get this in so we can iterate on the sync page behaviour and feedback when not working on the release one priorities

boerwastaken:
>> It looks like it GETs to /eikel-meta/status - for Canoe which isn't currently using the large object storage facility, I've removed those HTTP handlers, so this 404s. What's the use case for any information that could formerly be gleaned from there?

> Because that was one of the first APIs that you implemented that I could get a response out of.

You're not answering the question! :haircut_woman:

PR #480, opened 2021-02-25
interim code review PR to discuss some Appelflap <-> Canoe things

boerwastaken:
PR (not supposed to be merged) just to be able to comfortably discuss some things I'm seeing while laying eyes on the Canoe side of the Canoe <-> Appelflap interface.


src/js/RoutingAppelflap.js
```diff
@@ -0,0 +1,134 @@
...
+ * @remarks this MUST be 'localhost' not '127.0.0.1',
+ * because geckoview translates '127.0.0.1' to 'localhost' in the background
+ * which results in the route registration failing
+ */
+export const AF_LOCALHOSTURI = "http://localhost";
+
+export const AF_API_PREFIX = "appelflap";
+
+export const AF_EIKEL_API = `${AF_API_PREFIX}/eikel`;
+export const AF_META_API = `${AF_API_PREFIX}/eikel-meta`;
```
boerwastaken, line 11: Covering the Eikel API (large object storage facility) is not part of the scope of work for Canoe at the moment. It's not forbidden to implement it either ;-)
That said, `AF_META_API` is a misnomer, as it's really the Eikel meta-api. So it should be renamed to avoid confusion.

src/js/RoutingAppelflap.js
```diff
@@ -0,0 +1,134 @@
...
+    },
+    releaseLock: {
+        commandPath: `${AF_CACHE_API}/${AF_INS_LOCK}`,
+        method: "DELETE",
+    },
+    getCacheStatus: {
+        commandPath: `${AF_CACHE_API}/${AF_STATUS}`,
+        method: "GET",
+    },
+    doReboot: {
```
boerwastaken, line 42: I've changed this into soft-reboot and hard-reboot.
soft-reboot closes and reopens the session (sorta similar to when you'd close a tab and reopen it).
hard-reboot closes and reboots the gecko runtime (similar to closing and restarting your browser).

src/js/RoutingAppelflap.js
```diff
@@ -0,0 +1,134 @@
...
+        method: "GET",
+    },
+    savePublication: {
+        commandPath: `${AF_CACHE_API}/${AF_PUBLICATIONS}`,
+        method: "PUT",
+    },
+    deletePublication: {
+        commandPath: `${AF_CACHE_API}/${AF_PUBLICATIONS}`,
+        method: "DELETE",
+    },
```
boerwastaken, line 57: You can also getPublication nowadays (you'll get a ZIP file in the response).

src/js/RoutingAppelflap.js
```diff
@@ -0,0 +1,134 @@
...
+        .map((el) => String.fromCharCode(el.charCodeAt(0) - 0x31))
+        .join("");
+
+    return parseInt(portNo, 10);
+}
+
+/** Builds each Appelflap API route individually. These will not result in false positives */
+function initialiseSpecificRoutes() {
+    // Port number range is 2^10 to 2^16-1 inclusive - 1024 to 65535
+    const portRange = "(102[4-9]|10[3-9]\\d|1[1-9]\\d{2}|[2-9]\\d{3}|[1-5]\\d{4}|6[0-4]\\d{3}|65[0-4]\\d{2}|655[0-2]\\d|6553[0-5])";
```
boerwastaken, line 111: yikes, if you find yourself trying to catch arithmetic rather than lexical patterns with regexes, perhaps just use arithmetic? Check if numeric, cast to numeric, determine if it's in the expected range?

src/ts/AppelflapConnect.ts
```diff
@@ -0,0 +1,298 @@
...
+        const { commandPath, method } = APPELFLAPCOMMANDS.releaseLock;
+        return await this.performCommand(commandPath, { method }, "text");
+    };
+
+    public getCacheStatus = async (): Promise<any> => {
+        const { commandPath } = APPELFLAPCOMMANDS.getCacheStatus;
+        return await this.performCommand(commandPath);
+    };
+
+    public doReboot = async (): Promise<any> => {
```
boerwastaken, line 132: we have soft-reboot and hard-reboot now. I can't envision a scenario in which the webapp would want to inflict that on itself yet but it's incorporated in the eight layer indirection circus if we need it I guess?

src/ts/AppelflapConnect.ts
```diff
@@ -0,0 +1,298 @@
...
+    public getPublications = async (): Promise<TPublications> => {
+        const { commandPath } = APPELFLAPCOMMANDS.getPublications;
+        return (await this.performCommand(
+            commandPath
+        )) as Promise<TPublications>;
+    };
+
+    private publicationPath = (publication: TPublicationTarget) => {
+        const prepWebOrigin = encodeURIComponent(publication.webOrigin);
+        const prepCacheName = encodeURIComponent(publication.cacheName);
```
boerwastaken, line 146: that's changed from URI-encoding to base64-encoding, so `btoa()` away here

src/ts/AppelflapConnect.ts
```diff
@@ -0,0 +1,298 @@
...
+
+    /** Build the Version Min and Max headers.
+     * This method 'assumes responsibility' for the values provided to it.
+     * @param { TSubscription | TSubscriptionVersion } versionRange
+     * - A subscription or subscriptionVersion that identifies none, either or both a version min and a version max value
+     * @returns { Record<string, string> | undefined }
+     * undefined if there were no version min or max values, otherwise a header with the relevant values
+     * @throws { RangeError } If both a min and max are provided and they are incorrectly ordered
+     */
+    private buildVersionHeaders = (
```
boerwastaken, line 190: The subscriptions are now one and only one integral json doc, see 
https://github.com/catalpainternational/appelflap/blob/master/docs/API/update-subscription.md
No headers required!

src/ts/CacheUtilities.ts
```diff
@@ -0,0 +1,38 @@
+import { AppelflapConnect } from "./AppelflapConnect";
+
+export class CacheUtilities {
+    #afc: AppelflapConnect;
+
+    constructor(afc: AppelflapConnect) {
+        this.#afc = afc;
+    }
+
+    /** Get the status of the cache from Appelflap */
```
boerwastaken, line 10: It's not really the status of the cache. [See docs](https://github.com/catalpainternational/appelflap/blob/master/docs/API/cache-injection.md#operations---entity-information-block)

src/ts/tests/AppelflapConnect.test.ts
```diff
@@ -0,0 +1,555 @@
...
+    // global["navigator"] = buildFakeNavigator();
+    // const nfc = new AppelflapConnect();
+    // t.is(
+    //     nfc.portNo,
+    //     -1,
+    //     "navigator does not have encoded portNo - implies no Appelflap"
+    // );
+});
+
+/** meta status is not expected to be used by the Canoe-Appelflap cache API */
```
boerwastaken, line 87: indeed! It's not THE meta-status either. Rather it's the index of stored large objects (Eikels) and some operational metadata.

src/ts/tests/AppelflapConnect.test.ts
```diff
@@ -0,0 +1,555 @@
...
+        ],
+    };
+    fetchMock.get(testUri, state);
+
+    const result = await afc.getMetaStatus();
+    fetchMock.reset();
+    t.deepEqual(result, state);
+});
+
+test("Cache: lock", async (t: any) => {
```
boerwastaken, line 115: It's not really locking any Cache at all, so "Cache: lock" is a bit misleading.
Rather, it locks Appelflap so it won't do any injections.

src/ts/tests/AppelflapConnect.test.ts
```diff
@@ -0,0 +1,555 @@
...
+    t.deepEqual(successResult, testResponse);
+
+    fetchMock.get(testUri, authFailureResponse, { overwriteRoutes: true });
+    const result = await t.throwsAsync(afc.getCacheStatus());
+    t.is(result.message, authFailureResponse.statusText);
+
+    fetchMock.reset();
+});
+
+test("Cache: canoe reboot", async (t: any) => {
```
boerwastaken, line 179: The correct outcome of the test would be to quit the interpreter as that is what would happen if you'd call reboot ;-)

src/js/RoutingAppelflap.js
md8n: Thanks for the explainer - I'll change the names

src/js/RoutingAppelflap.js
md8n: This code isn't in use.  But it does work, and yes it was tested.  I left it in there because I didn't want to rewrite it later if we ever changed our minds...

src/ts/AppelflapConnect.ts
md8n: OK

src/ts/AppelflapConnect.ts
md8n: buildVersionHeaders is already gone from the most recent :pr:

src/ts/tests/AppelflapConnect.test.ts
md8n: Agreed - it's more of a "Please Don't Touch"

src/ts/CacheUtilities.ts
md8n: Or as your docs say:
> self-explanatory JSON document containing status information

src/js/RoutingAppelflap.js
```diff
@@ -0,0 +1,134 @@
+/** If Appelflap is running, it is always on localhost
+ * @remarks this MUST be 'localhost' not '127.0.0.1',
+ * because geckoview translates '127.0.0.1' to 'localhost' in the background
+ * which results in the route registration failing
```
boerwastaken, line 4: I still find this 
- vaguely defined (what is "translates"? where does it take place? which thing would have seen `127.0.0.1` were it not for this "translation"? why is it happening? what is "the background"?)
- improbable (may have to do with the vague definition)
- in need of a citation
- in need of expounding on what it is you're seeing that makes you think that this is going on

and furthermore
- why do pass `localhost` when the to-be-avoided behaviour (as you've tried to describe) would also have resulted in `localhost` being passed (somewhere, to something, somehow, for some reason) anyway?

boerwastaken: don't get me wrong, it's ok to see spooky/unexplained behaviour, it can happen, but then either state it with less certainty and confidence, and for reader's sake, separate observation from conclusion/interpretation and offer a simple test that shows the undesired behaviour, or... drill down on what's actually happening. Often taking steps towards shedding light on the problem makes it go away because it gets scared anyway!

boerwastaken:
A question. Does the webapp notice that operations (packing up for instance) are failing because of disk space shortage? And how does it deal with that?

Issue #473, opened 2021-02-25
Admins can see M&E data (weeks)


pedro16v:
## Conditions of Satisfaction
* Storing data
* Easy access to data
  * Option to setup dashboards in Google Analytics
  * Option to setup scheduled email reports 
  * Option to port G.A data into Google Data Studio
* Capture user’s feedback (smile / frown at the end of each lesson) on Wagtail
  * Test results saved to the Wagtail Database 
* Monitor impact of training
  * Completion of training modules
  * Trainees enrolled in the training program engagement with online platforms
    * % quizzes completed
    * Number of contributions to group discussions/webinars
    * % modules completed
    * % lectures watched (real time and recorded)
  * Disaggregated by (days): 
    * health worker type, 
    * experience, 
    * age, 
    * gender, 
    * health facility/department
  * Export data / tables (days)

### Update! (2021-05-11)
* Exam results
  * First score (first attempt)
  * Best score
  * Number of attempts

## Resources
* Engineering estimate: Weeks

## Notes
* Nicer interfaces can be pushed to Release Two.

PeteCoward:
@sandroroux to start on this I suggest that as discussed you concentrate on verifying that the data is being submitted and stored.

I think the best approach to this is to use https://docs.wagtail.io/en/stable/advanced_topics/customisation/custom_user_models.html

Individual projects would provide and configure a custom user model and custom user forms. This way the attributes are easy to change and add to per project.
You can play around in catalpa_canoe, ideally zero changes to canoe-backend should be required. 

For JID, we should try adding the following attributes

- [ ] health worker type ( foreign key to a model editable in the admin )
- [ ] experience ( no real clue so maybe another  foreign key to a model editable in the admin )
- [ ] age ( I would suggest integer Year Of Birth for this maybe current year minus 100 to current year )
- [ ] gender ( foreign key to a model editable in the admin - this list can be very dependent on culture )
- [ ] health facility/department ( foreign key to a model editable in the admin )

@Juliana-t can we discuss now which of the above should be editable for the JID deployment?
@KaraChesal we will need to discuss what attributes are needed for the PeP deployment

pedro16v:
> I'm using #492 to meet this Pedro-described requirement:
> 
> > Capture user’s feedback (smile / frown at the end of each lesson) on Wagtail
> > Test results saved to the Wagtail Database

It's actually a Juliana-described requirement 😁 
I'll make sure to work with @Juliana-t describing these requirements further.

[...]

PR #492, opened 2021-03-01
#488: Submits Lesson Feedback to User Actions

sandroroux:
Closes catalpainternational/canoe#488.

Contributes to https://github.com/catalpainternational/canoe/issues/473.

Requires `canoe-backend`'s `new_manifest` to merge the latest `RC-1.3`.


src/riot/Lesson/LessonFeedback.riot.html
```diff
@@ -37,24 +48,38 @@ <h5>{ TRANSLATIONS.value_feedback() }</h5>
         </template>
     </LessonFrame>
     <div if="{!state.madeFeedbackDecision}" class="bottom-buttons">
-        <button class="btn-secondary" onclick="{clickedNoThanks}">{ TRANSLATIONS.no() }</button>
-        <button class="btn-primary" onclick="{clickedSubmit}" disabled="{!state.activeSelection}">{ TRANSLATIONS.submit() }</button>
+        <button class="btn-secondary" onclick="{clickedNoThanks}">
```
PeteCoward, line 51: Can you set format only changed lines in visual studio please?
Especially while we have two feature branches in play, these non-change changes cause conflict headaches if someone changes one thing in these lines in the other branch

src/sw.js
```diff
@@ -28,7 +28,10 @@ registerRoute(
     })
 );
 
-registerRoute(new RegExp(ROUTES_FOR_REGISTRATION.images), new CacheAnyOrFetchOnly());
+registerRoute(
```
PeteCoward, line 31: Any actual changes in this file?

src/riot/Lesson/LessonFeedback.riot.html
sandroroux: Absolutely.

src/sw.js
sandroroux: Nope. I disabled workbox logging and removed the statement. Must have `prettier`ed the file on the way out.

src/js/Telemetry/index.js
```diff
@@ -0,0 +1,50 @@
...
+const postFeedback = async (feedback) => {
+    return await fetch(createPOSTRequest(ACTIONS_ENDPOINT, feedback));
+};
+
+const syncFeedback = async () => {
+    const feedbackKeys = await getFeedbackKeysFromIdb();
+    for (const feedbackKey of feedbackKeys) {
+        const feedback = await getFeedbackFromIdb(feedbackKey);
+
+        const response = await postFeedback(feedback);
```
PeteCoward, line 38: what is the behaviour if a feedback is posted, received by the server, but we do not get the ok response, it does not get deleted, then we try again?

sandroroux: That's my understanding.

Can the server create a user action and return a not-ok response?

PeteCoward: No, but the reply could be lost in the tubes on the way back.
I can't remember if the backend would error on that second attempt, I think it would due to a uuid clash, and then not return OK, therefore repeating for ever and ever :)

PeteCoward: No, but the reply could be lost in the tubes on the way back.
I can't remember if the backend would error on that second attempt, I think it would due to a uuid clash, and then not return OK, therefore repeating for ever and ever :)

sandroroux: I'll look into this. Thanks for describing that.

sandroroux:
@PeteCoward, I tried to handle the network errors. My approach feels a little loose. Let me know what you think.

src/js/Telemetry/index.js
```diff
@@ -27,16 +27,32 @@ const createPOSTRequest = (path, body) => {
 };
 
 const postFeedback = async (feedback) => {
-    return await fetch(createPOSTRequest(ACTIONS_ENDPOINT, feedback));
+    try {
+        return await fetch(createPOSTRequest(ACTIONS_ENDPOINT, feedback));
+    } catch (err) {
+        throw err;
+    }
```
PeteCoward, line 34: not sure how much point there is in a catch with just a throw inside it.
If you want to suppress the error, then catch but do not throw.
If you want to raise the error, just don't try.. catch

sandroroux: I think it's a code legibility thing. I'm basically saying, "this code throws, but I'm not interested in catching here. I'll pass it up to whatever uses me."

sandroroux: @PeteCoward, see here too:
https://stackoverflow.com/a/11853583

PeteCoward: from that link
> You catch an exception when you have something to do with it.

only the network error is an expected error here, and we know it might happen.
We also know that in theory not ok requests could occur.

If both those cases need to be handled in the same way we should raise a single fetch failed error in both cases to inform the calling code ( or catch the error and return a result if preferred )

The code as is just looks half implemented, and uglier and equal to no catching at all.

PeteCoward (changes requested):
This seems to have changes not related to the task
- why delete favicon.svg?
- why change serviceworker code at all ( you have added a `ROUTES_FOR_REGISTRATION.images` route duplication )

sandroroux:
@PeteCoward:
* favicon.svg throws an error that blocks yarn's build process.
* The service worker changes are an unintentional merge mistake.

I'll revert the service worker changes. Have you run into the favicon errors?

src/js/Telemetry/index.js
sandroroux: I've read through your comment and am going to have another go at this.

To start, I built my approach using [this](https://medium.com/to-err-is-aaron/detect-network-failures-when-using-fetch-40a53d56e36) article. (FYI, the article covers errors much more exhaustively than my code).

If I understand your point correctly, I want to treat three cases separately:
* a 400-range response will return a response object;
* a 500-range response will throw an error;
* a network failure will throw an error.

My strategy is to only save my local copy when the network fails. In other cases, I'll delete my local copy. The way that breaks down is:
* If the response is 200 (`response.ok === true`), I will delete my local copy. The server has the entry, so the local copy is unneeded.
* If the response is 400 (`response.ok === false`), I will delete my local copy. The local copy is probably malformed. To prevent repeat 400s, I'll delete my local copy, but I'll log the not-okay.
* If the network fails (`fetch` throws an exception), I will try the fetch again on the next poll. The network is spotty, so I should persist the upload until it succeeds or 500s (because of a double-entry error, for example).
* If the server throws a 500 (`fetch` throws an exception), I will delete my local copy. A server bug means we'll lose feedback data inadvertantly. Hopefully, we'd notice the sentry errors before too much data loss. If a double-entry error occurs, we will delete the local copy and the UUID clash will disappear.

PeteCoward:
> * favicon.svg throws an error that blocks yarn's build process.

no

src/js/Telemetry/index.js
PeteCoward: how does the server respond when you throw up a duplicate? can we stop that 500ing.
Expecting 500s is bad :)

sandroroux: Good point. Here's [a back-end PR](https://github.com/catalpainternational/canoe-backend/pull/130) that turns UUID errors into 400s.

sandroroux: Updates.

Now, that back-end PR returns 201s when devs double UUIDS.

Since my earlier comments, I changed the algorithm. Here's the current approach:
* If the response is 200 (response.ok === true), I'll delete my local copy. The server has the entry, so the local copy is unneeded.
* If the response is 400 (response.ok === false), I throw the error. The local copy is probably malformed. The error may signal a bug in feedback-submitting code, so I want the developer to know about it.
* If the network fails (fetch throws an exception), I'll try the fetch again on the next poll. Since the network is spotty, I repeat the upload until it succeeds.
* If the server returns a 500 (fetch throws an exception), I throw the error. A 500 may signal a server issue that I want the developer to know.

PR #500, opened 2021-03-05
fix bad conflict resolution merge

PeteCoward:


md8n (approved):

PeteCoward:
leaving this open for @katevoss to see in her morning, please merge if ok Kate

src/riot/Lesson/LessonContent.riot.html
```diff
@@ -11,7 +11,7 @@
         <template if="{props.description}">
             <p class="resource-description">{ props.description }</p>
         </template>
-        <template if="{props.resourceId === 'resources'}" each="{card in props.lessonContent.cards}">
+        <template if="{props.resourceId === 'resources'}" each="{card in state.cards}">
```
katevoss, line 14: The resource article still doesn't load because the data shape for a ResourcePage is different from a LessonPage. I can't find a simple fix to make it work, they seem to be very different and there are a lot of ways that they're intertwined. I think the logic and styling for Resource Articles needs to be separated from LessonContent, either on this PR or on one soon.

katevoss (changes requested):
The resource articles don't load so this isn't mergeable because we are deploying FT-Offlining so it can't have broken things. If we can have broken things on FT-Offlining then we can merge and make another PR to fix articles.

src/riot/Lesson/LessonContent.riot.html
PeteCoward: We should stop the resources page using LessonContent I think

PeteCoward:
My preference would be to merge this now, as it looks like lesson content is working.
Then assign someone ( or me ) to fix resource content, preferably by disassociating it form the components used to render lessons

Issue #491, opened 2021-03-01
Card Images: Offline Behaviour


PeteCoward:
Towards #472

## Conditions of Satisfaction

- [ ] when offline and card images can't be accessed card images fallback to the existing card defaults 
- [ ] once card images have been shown they are available offline

## Strategy

The aim is:
1. try to serve card images from the cache
2. if 1. fails try to get them from the internet
3. if 2. fails deliver a fallback image ( the one we use when card_image_url is null )
4. if 2. succeeds add that image to a cache
5. make sure the cache can be cleaned up if we run out of space ( these images are not educationally critical )

## Implementaion

1. Move the existing css background implementation of stock card images to an `<img>` tag based solution
  - this may mean we need to change the svgs to pngs, but somehow check that webpack exports them as actual image files ( currently svgs are auto data inlined into the css )
2. Add a new route / strategy to the serviceworker that handles these requests
  - The simplest way to do this is using a [CacheFirst](https://developers.google.com/web/tools/workbox/modules/workbox-strategies#cache_first_cache_falling_back_to_network) strategy... with some plugins
  - Add an [expiration plugin](https://developers.google.com/web/tools/workbox/guides/storage-quota?hl=en#what_configuration_options_are_supported) configured to use the cache name `card_images` and be `purgeOnQuotaError: true`
  - Add a [custom plugin](https://developers.google.com/web/tools/workbox/guides/using-plugins#custom_plugins) that reads the required fallback from the image request querystring 
```javascript
handlerDidError: async ({request, event, error, state}) => {
    //more to come here!
    return response;
}
```
to return an appropriate response from the cache

Issue #472, opened 2021-02-25
As a content manager, I can select a profile image to Courses and Lessons


pedro16v:
## Conditions of Satisfaction
* I can pick from a range of pre-selected images
* I can upload my own image
* Different implementations of Canoe will have different image sets
* There are default placeholders for when a image is not available

## Resources
* Engineering estimate: Days

PR #501, opened 2021-03-05
472 hero image implementation & 491 offline capability

megganeturner:
Addresses #472 & #491: implements images for cards, and adds in offline capability.

branches for testing: canoe-backend can be on `RC-1.3`, and catalpa_canoe is on `canoe-engineering`

#472:
following on from Peter's changes in the backend (https://github.com/catalpainternational/canoe-backend/pull/99):
- replaces `<span>` with `background-image` with an `` tag
- if a card has a value for the `card_image_url` attribute, we pass it through to the `<Card>` component and render it
- if a card doesn't have a value for `card_image_url` attribute, we display a fallback image
- append the fallback image as a queryParam to the imageUrl as `?cardFallbackImage={}` for offlining stuff
- fallback images have been randomised rather than being specific to a card type - this is a request from Andy, and is easy enough to change if we decide to revert back

#491:
Conditions of satisfaction:
- [x] when offline and card images can't be accessed card images fallback to the existing card defaults
- [x] once card images have been shown they are available offline

While I'm feeling a lot better about service workers in general they are still very much out of my comfort zone, so any feedback here is very welcome 😁 

- register a route to cache images with the `?cardFallbackImage=` query param attached, to cache to `card-images-cache`
- extend this route with a custom plugin, which defines the following callbacks:
&nbsp;&nbsp; - `fetchDidFail` - "Called when a network request fails, most likely due to a NetworkError. Note that this does not get called when a response with an error status, like 404 Not Found, is returned from the network."
&nbsp;&nbsp; - `fetchDidSucceed` - "Called when a network request is successful, regardless of what the HTTP status is of the response."
&nbsp;&nbsp; - `handlerDidError` - "Called if the handler was unable to provide a valid response from any source."
- additionally, defining `setCatchHandler`, which "is triggered when any of the other routes fail to
generate a response." From what I can see, it seems to get fired when the server is offline

All of these are doing the same thing: providing the fallback url based on the queryParams on the image url. This may be overkill, but some of them only get called in certain situations and I think this covers all our bases.

You'll need to add images to a couple of courses & lessons in your local instance of wagtail. To test the image fallback, you can change line 36 of `Components/Card.riot.html` to something like
`return this.props.imageUrl ? `not/an/image.png?cardImageFallback=${fallbackImg}` : fallbackImg;`

Relevant documentation:
- https://developers.google.com/web/tools/workbox/guides/using-plugins?hl=en#custom_plugins
- https://developers.google.com/web/tools/workbox/guides/advanced-recipes#comprehensive_fallbacks


src/riot/Components/Card.riot.html
```diff
@@ -13,4 +14,21 @@ <h5 class="clamp-lines">{props.title}</h5>
...
+        import stockImg3 from "img/stock_image_3.png";
+        import stockImg4 from "img/stock_image_4.png";
+        import stockImg5 from "img/stock_image_5.png";
+
+        export default {
+            getCardImageUrl() {
+                const stockImages = [stockImg1, stockImg2, stockImg3, stockImg4, stockImg5];
+                const fallbackImg = stockImages[Math.floor(Math.random() * stockImages.length)];
+
+                return this.props.imageUrl ? `${getMediaUrl(this.props.imageUrl)}?cardImageFallback=${fallbackImg}` : fallbackImg;
```
katevoss, line 30:
```suggestion
                return this.props.imageUrl
                    ? `${getMediaUrl(this.props.imageUrl)}?cardImageFallback=${fallbackImg}`
                    : fallbackImg;
```

src/riot/Components/Card.riot.html
```diff
@@ -1,6 +1,7 @@
 <Card>
     <a class="card {props.status}" href="#{props.link}">
         <div class="card-image">
+            <img crossorigin="anonymous" src="{getCardImageUrl()}" class="card-hero-image">
```
katevoss, line 4:
```suggestion
            <img crossorigin="anonymous" src="{getCardImageUrl()}" class="card-hero-image" />
```

katevoss (commented):
I haven't looked at the sw stuff yet but the riot stuff looks good so far 👍🏻 

Sometimes I get these img not found errors, I'll look more into it next week
image

PeteCoward (commented):
Why am i seeing stock_image_1.png here as empty files? Are they empty or an accidental add/commit?

src/sw.js
```diff
@@ -31,6 +33,53 @@ registerRoute(
...
+    function(request) {
+        return request.url.searchParams && request.url.searchParams.has('cardImageFallback')
+    },
+    new CacheFirst({
+        cacheName: "card-images-cache",
+        plugins: [
+            cardFallbackPlugin,
+            new ExpirationPlugin({
+                maxEntries: 50,
+                maxAgeSeconds: 30 * 24 * 60 * 60,
```
PeteCoward, line 69: Not sure these lines are wise! No need to expire based on number of entries or age

PeteCoward (commented):
Why the move from named stock images to numbered ones?
Naming is helpful I think :)

megganeturner:
Hey @PeteCoward - if you click on the 'display rich diff' button you'll be able to see the images. They've also been given numbers as Andy requested they be randomised, rather than having 1 image for a course, another for a lesson etc.:

src/sw.js
```diff
@@ -65,8 +65,6 @@ registerRoute(
         plugins: [
             cardFallbackPlugin,
             new ExpirationPlugin({
-                maxEntries: 50,
```
PeteCoward: oh dear this was a bad idea of mine!
`WorkboxError: You must define either config.maxEntries or config.maxAgeSecondsin workbox-expiration.Plugin.constructor`

PeteCoward: I suggest we just set a large maxEntries

megganeturner: Ah! That's my mistake for not testing again! Is 1000 too large? Not large enough?

PeteCoward (changes requested):
I'm seeing requests to `/media//media/images/Triage_E_pJhasJq.2e16d0ba.fill-200x100-c50.format-jpeg_k7hJpfK.jpg` which always 404, and always show the fallback.
So the fallback is working, but the real image is not :)

src/sw.js
PeteCoward: maybe a maxAge equivalent to 5 years is better

PeteCoward:
Also we are caching the image with a number of different fallback urls as part of the key. This means we are storing the same image many times over. I'll pull and see if I can fix.

megganeturner:
After discussing with Pete, I've de-randomised the fallback images - we can discuss this with wider team at design-eng sync on Wednesday :)

I'll take a look at the 404 issue with  `/media//media/images/Triage_E_pJhasJq.2e16d0ba.fill-200x100-c50.format-jpeg_k7hJpfK.jpg` tomorrow morning!

PeteCoward (commented):
Looks good, I'll test it once more...

PeteCoward (approved):

PR #503, opened 2021-03-08
Fix card-image path and re-oreder sw routes

PeteCoward:
- use `process.env.API_BASE_URL` to place the backend domain in front of the image path ( before we had media//media paths )
- move the normal content image handler under the card image handler ( so the card images don't get caught by it )
- make sure the card image responses are cached without the fallback querystring ( so only one gets cached )


PeteCoward:
I merged conflicts, leaving the non-randomness in place.
One technique for "sticky" randomness would be
```
function stickyFakeRandom(seed) {
    var x = Math.sin(seed) * 10000;
    return x - Math.floor(x);
}
```
If you call this function to generate your random number, passing in the id of the page represented by the card, it will produce a fake random number that will always be the same for the same page

megganeturner (approved):

PR #506, opened 2021-03-09
Learning activities

md8n:
# Description

This gets the basic Learning Activities wiring up done on the client side.

Some of these changes will need to be moved from where they are or altered to point at  * new *  Learning Activities pages :
* App.riot.html
* BottomMenu.riot.html
* Manifest.ts

The change to AllCoursesPage.ts could be simply rolled back once the new pages exist, however it does make the existing code more robust.


PeteCoward:
I'm going to make this draft. 2 things that might be good to keep from this are the CanoeEnums entries and maybe the bottom nav hide show code

PR #509, opened 2021-03-10
504 offlining audit fixes

PeteCoward:
To be run against canoe-backend `list_requirements_in_manifest` from https://github.com/catalpainternational/canoe-backend/pull/111

- Adds Page.ts accessors for tags and cardImage, removing the CoursePage.ts tags function
- Fixes the tag list which was broken by a bad merge


katevoss (approved):
This looks good! I'm going to merge so I can use this in my branch today.

Issue #504, opened 2021-03-09
Audit and fix frontend issues on FT-Offlining


katevoss:
- [x] Pull ResourceArticle out of LessonContent (and related cleanup) 
     - [x] make ResourceArticle a vertical scroll of cards
     - [x] Resource articles, not loading
     - [x] resource tags not loading
- [x] Profile page, loading courses (needs manifest)
- [x] ~Video and audio cards, not loading~ this was a server error, media was not being correctly transcoded, pull redesign db and media to retest
- [x] Tags list, not loading - addressed in #509
- [x] back arrow link broken (ex. from ResourceArticle back to Resources)
- [x] consistency (ex: use [`showTagList()` to determine if the page is loaded and tags exist in all instances](https://github.com/catalpainternational/canoe/blob/FT-Offlining/src/riot/WagtailPages/AllCoursesPage.riot.html#L17))
- [x] remove logic to [clear state between answer selections](https://github.com/catalpainternational/canoe/blob/3ea4846d622f5fac57575c36d88f779e82341bba/src/riot/Lesson/TestSingleAnswer.riot.html#L79) 
- [x] Make "you're not logged in" banner look better
    <img src="https://user-images.githubusercontent.com/2395211/110655630-3a18ae80-818d-11eb-97a7-6bd64aa33413.png" width=230>
- [x] Video and Audio cards throw error `Uncaught (in promise) TypeError: Cannot read property 'mediapath' of undefined`

PeteCoward:
I'll take the Profile page stuff, and am looking for volunteers for the Resources issues.
The clear state between answer selections is low priority, but feel free!

PeteCoward:
The video and media path stuff...

They should work in the lessons, but only when a transcode has succeeded, we may need to deal better with transcode failures
They may not work yet in resources, one for tomorrow

PR #511, opened 2021-03-11
Resource article fix

katevoss:
Adds to #504 

This PR takes `<LessonContent>` out of `ResourceArticle.riot.html` and now ResourceArticle has its own content cards.

#### What doesn't work (I think unrelated to this PR):
- Video and Audio cards return an error such as
     
    If you see this error, refresh the page and it should clear.
- ImageWrapper also doesn't work, not sure why
- Filtering Resources by tags or search query doesn't work
- The first time you go to the Resource page, you won't see the description or card type on the Resources list. After you visit the Resource Article, you'll see it. 

#### What should work:
- Lesson Content (except Video and Audio)
- back buttons and exit buttons across all lesson/course/activities/resources
- Resource articles displaying multiple cards in a scroll


src/riot/Lesson/LessonTest.riot.html
```diff
@@ -74,15 +74,14 @@ <h3>{ state.cardNumber }. { state.card.question }</h3>
...
-/*
+            /*
                 window.addEventListener(ON_ANSWERED_TEST_QUESTION, event => {
                     logATestAnswer(event.detail);
                 const correctAnswers = card.answers.filter(answer => answer.correct);
                 this.update({
                     isSingleAnswer: correctAnswers.length === 1,
                 });
-*/
+            */
```
PeteCoward, line 84: I'll see if this can be deleted, maybe I forgo

src/riot/Screens/Resources.riot.html
```diff
@@ -83,9 +86,8 @@ <h6 if="{ resource.ready }">{resource.description}</h6>
             onBeforeUpdate() {
                 this.state.isLoading = !this.state.page.ready;
             },
-            /*
+
             async onMounted() {
-                this.toggleLoadingMessage();
                 const articles = (await getResources()).map(
```
PeteCoward, line 91: this should be accessing the child pages on the ResourcesList model, I'll try and fix, it was commented out for a reason. I should have deleted it :)

src/riot/Lesson/LessonTest.riot.html
PeteCoward: ...t, or maybe i left it in because i had not verified it's functionality has been replaced

PR #512, opened 2021-03-11
504 list pages fixes

PeteCoward:
Edits all of the list pages to instantly render their lists from the manifest data. Merging into @katevoss 's #511 for now

Sometimes this means there is no need for a loader, we may revisit that and show loaders for only the bits that do ( as in the course exam )

Also applies @megganeturner 's card images on Teaching/Learning list pages ( not resources yet )

PLan is to get both this and RC-2.0 up for review tomorrow morning

### DIFF
--- a/src/js/AuthenticationUtilities.js
+++ b/src/js/AuthenticationUtilities.js
@@ -108,9 +108,11 @@ export const isGuestUser = () => {
 };
 
 export const getUserId = () => {
-    return getUser().userId;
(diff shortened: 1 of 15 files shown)


katevoss:
I added a bunch of these commits to my PR https://github.com/catalpainternational/canoe/pull/513
It seemed like the easiest way to build on the work.

PeteCoward:
ok closing this

PR #529, opened 2021-03-19
Try a refresh on every route

md8n:
# Description

We want the manifest checked (and refreshed) regularly, as this drives the updating process for the pages and assets (and what's in the store and caches).

One point where we always check the Manifest is when routing.  By slightly tweaking this behaviour we can get it to try refreshing the manifest for every route, currently the manifest is not refreshed if the copy in the store `isvalid`.

The service worker was set to do `StaleWhileRevalidate` for any request to the manifest API endpoint and considering we were effectively doing that anyway because we're checking the redux store and testing `isValid`, this strategy has been changed to `NetworkFirst`.


src/js/Routing.js
```diff
@@ -75,10 +75,13 @@ async function getValidManifest() {
     const manifest = new Manifest();
     if (!manifest.isValid) {
         await InitialiseByRequest(manifest);
-        return manifest;
     } else {
-        return Promise.resolve(manifest);
+        // Don't await this delierately,
+        // in effect we'll try refreshing the manifest in the background
+        InitialiseByRequest(manifest);
```
PeteCoward, line 81: imagine the situation where the user has manifest version X in the store
someone updates the server editing an image, manifest X+1 is now available
A user navigates to a page, this code requests the new manifest, an async op
The page itself potentially tries to update from the server
The page riot tag is mounted/updated according to the old manifest info ( either with or without the page api data )
Then the manifest request async resolves, we have a new manifest in the store.
Then the page api request returns, the riot updates.
The riot is now rendering a page which may not be in the current manifest, causing errors ( like missing images )

This makes me think 2 things...
1. We may need to stage that new manifest before applying it to the store
2. We may need to find a way to deal with the situation where the page data has updated, but the manifest hasn't yet. requesting specific revisions from the API maybe

2 is tricky. 

1 should be reasonably straight forward.
  - implement the Last-Modified handling so we know when we have a new version ( 200 response not 304 )
  - add a stagedManifest in the store, which we switch to on next route

md8n: These were all the same concerns that I had last Friday, thanks for writing it up though.

The reason why I choose async update on route was because the route path always checked the manifest.  I realised from that that routing had to be a part of the manifest refresh process.  But then that raises all of the other scenarios as you described above.

I like your option 1 of staging the manifest, regardless of whether routing or polling caused it to be refreshed, and then if it has been updated:
1. Display a toast message to the user that the learning / teaching data has been updated, and if they want to refresh they need to return to the home screen.
2. As soon as they enter any of the 'home' screens the staged manifest is applied, and normal operations continue from there ...

PeteCoward: as long as the manifest is applied on route op, not an async return, we are good for now.
We can chat with design about whether they want UI in the process

PR #538, opened 2021-03-24
Load audio and video properly

### LINKED ISSUE #537: Any 'media' in a page doesn't get loaded properly
Audio and Video in pages doesn't get loaded properly, instead a request to the `/media` API endpoint without any further route information is sent.

md8n:
Closes #537 

# Description

Media (Audio and Video) were not being loaded properly from their references in the manifest.

This fixes that(with corresponding changes from canoe-backend).

Also fixed, certain long redundant functions have been removed.


src/ts/Implementations/Asset.ts
```diff
@@ -1,11 +1,18 @@
...
-import { TAssetEntry } from "../Types/AssetTypes";
+import { TAssetEntry, TRendition } from "../Types/AssetTypes";
 import { PublishableItem } from "./PublishableItem";
 import { UpdateCachedItem } from "./CacheItem";
 
 import { AppelflapConnect } from "../AppelflapConnect";
-import { JPEG_RENDITION, WEBP_BROWSERS, WEBP_RENDITION } from "../Constants";
+import {
+    JPEG_RENDITION,
+    MP4A_RENDITION,
```
boerwastaken, line 10: I'd think the media renditions are supposed to be dynamic - we should be able to introduce a new one without having to update the clientside code. Does the use of these constants imply that we're instead going for a limited and fixed set?

md8n: This behaviour is a leftover from before. But it does allow us to 'handle' people who insist on using Safari.
Personally I agree that would should be more flexible, but I do think that should be a separate :pr:

src/js/Routing.js
```diff
@@ -57,6 +57,16 @@ async function route(hashWith) {
...
+        if (!manifest) {
+            route = {
+                page: {
+                    type: "manifest_error",
+                    error: `Manifest retrieval failure. Error: resetting the manifest in the cache failed, probably due to CORB issues`
+                }
+            };
+            return;
+        }
+
```
PeteCoward, line 69: `!manifest` I can't see how that could happen at all

PeteCoward: manifest would have to be null or undefined or empty, but it's the result of `new Manifest()`

src/js/WagtailPagesAPI.js
```diff
@@ -1,13 +1,8 @@
 import { BACKEND_BASE_URL } from "js/urls";
-import { token_authed_fetch, unauthed_fetch } from "js/Fetch";
+import { unauthed_fetch } from "js/Fetch";
```
PeteCoward, line 2: none of our FT-Offlining ode should be using the methods in this file. Is that correct?

src/ts/Implementations/Asset.ts
```diff
@@ -54,12 +61,24 @@ export class Asset extends PublishableItem<TAssetEntry> {
...
-        return getBrowser().name === "Safari" ? JPEG_RENDITION : WEBP_RENDITION;
+        const browser = getBrowser().name;
+        switch (this.type) {
+            case "image":
+                return browser === "Safari" ? JPEG_RENDITION : WEBP_RENDITION;
+            case "video":
+                // Safari does support webm to a limited degree
+                return browser === "Safari" ? MP4V_RENDITION : WEBM_RENDITION;
+            case "audio":
+                return browser === "Safari" ? OPUS_RENDITION : MP4A_RENDITION;
```
PeteCoward, line 77: for both video and audio, we cannot guarantee having these renditions. In that case we need to fall back to use "original"

src/js/Routing.js
md8n: I'll pull this one out.  It was being caused by webpack having an almighty brain fart while watching for code changes.

src/ts/Implementations/Asset.ts
boerwastaken: Wouldn't it be better if we wouldn't have to hardcode all this stuff?
According to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video , you can have multiple `<src>` elements (each having a distinct `type` attribute) in a `<video>` element, and the browser will choose the type it finds palatable.

Can't we reduce the job of the frontend to picking the appropriate dimension? say 240p, 360p, 480p.

But in order to precache videos, we'll need to know which format the browser likes *upfront*. For that then we'll probably need to probe: insert a small probe multisource video element (off-screen), and see which one the browser ends up loading (serviceworker helps here). We wouldn't have to probe all the time, just when the browser UA changes, say because of a browser engine version upgrade.
It is a bit of a hacky route, but if we want it, there IS a way to avoid hardcoding relations between browsers and the media types they can play.

src/js/WagtailPagesAPI.js
md8n: Unfortunately not true, the ImageWrapper.riot tag and I think one other still do.

PeteCoward:
Ok, I'm getting it in because the backend now delivers the manifest format this requires.
On Wicher's comments hard coding is good enough for now, but using mulitple sources would be better, maybe not the funky probing yet though.

PR #543, opened 2021-03-25
467-question card with images

megganeturner:
If questions come through with an image attribute, render it in the card

Screen Shot 2021-03-25 at 2 51 50 pm

Screen Shot 2021-03-25 at 4 53 41 pm

### DIFF
--- a/src/riot/Components/BottomMenu.riot.html
+++ b/src/riot/Components/BottomMenu.riot.html
@@ -64,6 +64,9 @@
                 if (this.page.type === 'coursepage' && !this.riotHash) {
                     return false;
                 }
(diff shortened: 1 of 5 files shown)


src/riot/Components/BottomMenu.riot.html
```diff
@@ -64,6 +64,9 @@
                 if (this.page.type === 'coursepage' && !this.riotHash) {
                     return false;
                 }
+                if (this.page.type === 'coursepage' && this.riotHash[0] === 'exam') {
```
katevoss, line 67: Aha you fixed the issue I saw, I thought I'd caused it 😅

PR #545, opened 2021-03-26
Multi choice state fix

megganeturner:
Previously when there were multiple multi-choice questions in a lesson/exam, the state wasn't being reset correctly between them and the answers stored in state for question B were actually the answers from question A. This should resolve that by explicitly resetting the `hiddenAnswers` attribute.

Also a small addition to the course exam page to prevent the honor code flashing up on screen while the calculations were being done for the route.


katevoss (approved):
Looks good! I'm not merging yet because this is a convenient base branch that is functional (while FT-Offlining is broken) 😁

PR #550, opened 2021-03-29
Refactor caching

PeteCoward:
A simplification of some backend of the frontend components

- removes duplication of state values in components attributes
- adds ability to make things offline and remove them ( unstyled )
- breaks the sync page ( but enables re-engineering it more simply

@md8n let's discuss


md8n:
No problem with breaking the sync page - It's about at the stage where it needs serious re-engineering anyway, because we now understand how the interaction with Appelflap will work

md8n (approved):

PR #552, opened 2021-03-29
Point TeachingActivity at the correct field

md8n:
# Description

TeachingActivity's members were pointing at `data` when they needed to be pointing at `storedData`

The code is written in that way (optional chaining `?.` plus an `||` default) to stop eslint complaining.


md8n:
superseded

PR #554, opened 2021-03-29
Caching ex unified media renditions

md8n:
# Description

Get the rendition handling standardised.

We should do a specific test for Audio.  Everything else 'seems' OK.


src/ts/Implementations/Asset.ts
```diff
@@ -34,18 +34,31 @@ export class Asset extends PublishableItem {
...
     get url(): string {
-        return `${process.env.API_BASE_URL}/media/${this.platformSpecificRendition.path}`;
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        return Asset.Url(this!.entry as TAssetEntry);
     }
+
+    /**
+     * The platform specific media url of this asset item
+     */
+    static Url(asset: TAssetEntry): string {
```
PeteCoward, line 47: can you explain why you did this?

PeteCoward: Url and url, I don't like

md8n: Because the 'assets' that we get returned are not Asset objects, just the Asset data.
So by flinging that data into a static method we can still get all that we need.

src/ts/Implementations/Asset.ts
```diff
@@ -110,14 +132,18 @@ export class Asset extends PublishableItem {
                 renditionSpec =
                     browser === "Safari" ? OPUS_RENDITION : MP4A_RENDITION;
                 break;
+            case "pdf":
+                renditionSpec = "original";
```
PeteCoward, line 136: are pdfs uploadable as images?

md8n: Nope - but this is me being complete - Technically PDFs are media, but this code won't be tripped over at present (I have checked that).

It would be nice (in the future) to have PDFs handled through this code as well, rather than the 'special' case that they are currently.

PR #556, opened 2021-03-30
Restore JWT for Page and Manifest

md8n:
# Description

This restores the use of a JWT bearer token for `Page` and `Manifest`.

It also restores the type for these requestOptions to `RequestInit` and fixes a bug in `PublishableItem.ts` `getResponseFromCache` where the wrong data was being passed for the `caches.match` query options

### DIFF
--- a/src/ts/Implementations/Asset.ts
+++ b/src/ts/Implementations/Asset.ts
@@ -12,6 +12,9 @@ import {
 } from "../Constants";
 import { getBrowser } from "../PlatformDetection";
 
+// See ts/Typings for the type definitions for these imports
+import { BACKEND_BASE_URL } from "js/urls";
+
 /** A asset ( binary resource ) than can be cached
  */
 export class Asset extends PublishableItem {
@@ -62,10 +65,18 @@ export class Asset extends PublishableItem {
     /**
      * The options to make an asset request
      */
-    get requestOptions(): any {
+    get requestOptions(): RequestInit {
+        const headers: any = {
+            "Content-Type": this.contentType,
+        };
+
         return {
             cache: "force-cache", // assets are (almost always) invariant on filename
-        };
+            headers: headers,
+            method: "GET",
+            mode: "cors",
+            referrer: BACKEND_BASE_URL,
+        } as RequestInit;
     }
 
     /**
@@ -92,6 +103,23 @@ export class Asset extends PublishableItem {
      */
     get type(): string | undefined {
         return this.entry?.type;
+    }
+
+    /** The mime type of this asset */
+    get contentType(): string {
+        const browser = getBrowser().name;
+        switch (this.type) {
+            case "image":
+                return browser === "Safari" ? "image/jpeg" : "image/webp";
+            case "video":
+                // Safari does support webm to a limited degree
+                return browser === "Safari" ? "video/mp4" : "video/webm";
+            case "audio":
+                return browser === "Safari" ? "audio/ogg" : "audio/mp4";
+            default:
+                // Default is for JSON, and we don't care about the browser
+                return "application/json";
+        }
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/ts/Implementations/Asset.ts
```diff
@@ -62,10 +65,18 @@ export class Asset extends PublishableItem {
...
+            "Content-Type": this.contentType,
+        };
+
         return {
             cache: "force-cache", // assets are (almost always) invariant on filename
-        };
+            headers: headers,
+            method: "GET",
+            mode: "cors",
+            referrer: BACKEND_BASE_URL,
```
PeteCoward, line 78: no need for `referrer` , it just happens

src/ts/Implementations/Asset.ts
```diff
@@ -62,10 +65,18 @@ export class Asset extends PublishableItem {
     /**
      * The options to make an asset request
      */
-    get requestOptions(): any {
+    get requestOptions(): RequestInit {
+        const headers: any = {
+            "Content-Type": this.contentType,
```
PeteCoward, line 70: these are get requests, they don't need a content type

src/ts/Implementations/PublishableItem.ts
```diff
@@ -12,15 +12,23 @@ const logger = new Logger("ContentItem");
...
-    abstract get requestOptions(): any;
+    abstract get requestOptions(): RequestInit;
     /** Brief descriptive string for logs */
     abstract get str(): string;
 
+    /** The options used to query the caches for this item */
+    get cacheOptions(): MultiCacheQueryOptions {
+        return {
+            cacheName: this.cacheKey,
+            ignoreVary: true,
```
PeteCoward, line 28: why add ignoreVary ?

src/ts/Implementations/Asset.ts
```diff
@@ -62,10 +65,18 @@ export class Asset extends PublishableItem {
...
+        };
+
         return {
             cache: "force-cache", // assets are (almost always) invariant on filename
-        };
+            headers: headers,
+            method: "GET",
+            mode: "cors",
```
PeteCoward, line 77: these two are defaults, but for explicitness, sure

src/ts/Implementations/PublishableItem.ts
md8n: It's the most common flag added to cache query options to get the 'VARY' header ignored.
Surprisingly frequent how often this will filter out a cached repsonse

src/ts/Implementations/Asset.ts
md8n: removed

src/ts/Implementations/Asset.ts
md8n: not always I found - especially when I'm testing locally

src/ts/Implementations/PublishableItem.ts
PeteCoward: but this just prevents us doing things we might want to do.
If the server tells us that a response varies on a particular header, why not used the default behaviour which is to respect it?

md8n: Practical experience (and a lot of other people's experience) says to add this particular value in by default.

src/ts/Implementations/Asset.ts
PeteCoward: it doesn't always get added? What happens when it doesn't?

src/ts/Implementations/PublishableItem.ts
PeteCoward: Hearsay doesn't cut it. Explain better please. But later. For now it's getting merged due to time pressure, but I will follow up with you.

PeteCoward (approved):

PR #557, opened 2021-03-30
Add course list title and description

katevoss:
If a content editor has added a title and/or description to the All Courses list then we'll show it.

_NOTE_: you must run this with this backend branch `home-page-body-no-richtext` (https://github.com/catalpainternational/canoe-backend/pull/143)


src/riot/WagtailPages/AllCoursesList.riot.html
```diff
@@ -1,9 +1,9 @@
 <AllCoursesList>
     <GuestBanner></GuestBanner>
     <TopMenu extrastyleclasses="with-swoop">
-        <h3>{ TRANSLATIONS.allCourses() }</h3>
+        <h3>{ page.title ? page.title : TRANSLATIONS.allCourses() }</h3>
```
PeteCoward, line 4: this might work too
```suggestion
        <h3>{ page.title || TRANSLATIONS.allCourses() }</h3>
```

src/ts/Implementations/Specific/AllCourses.ts
```diff
@@ -96,4 +96,12 @@ export default class AllCourses extends Page {
             this.countFinishedLessonsAndExams
         );
     }
+
+    get title(): string {
+        return this.data.title;
```
PeteCoward, line 101:
```suggestion
        return this.manifestData.title;
```
is more immediate, available from the manifest

PeteCoward: Actually all pages already have a title, you can remove this whole method and it will still work

src/riot/WagtailPages/AllCoursesList.riot.html
katevoss: Huh, the Learning Home page title is required by Wagtail so this is a moot point.

src/riot/WagtailPages/AllCoursesList.riot.html
```diff
@@ -97,14 +97,19 @@ <h5 class="section-title">{ TRANSLATIONS.allCourses() }</h5>
...
                         ),
                 },
 
                 onBeforeMount(props, state) {
                     this.page = getRoute().page;
                 },
 
+                onMounted() {
+                    this.$("#all-course-description").innerHTML =
+                        this.page.description || this.TRANSLATIONS.coursesIntro();
```
katevoss, line 110: @PeteCoward the description is available when I render `{ page.description }` in the DOM but it returns `undefined` in `onMounted`. I put it here to remove the `<p>` tags so it doesn't look like this:
image

Why is it returning undefined in onMounted? I saw this happen when other class properties return undefined but should have a value.

PeteCoward: We should remove the p tags on the backend by making it a simple text field not a richtext field
It's undefined on mounted because the description is in the page data not the manifest, we might change that too ( it is why the resource descriptions and search does not work on first load

PeteCoward:
Making this draft until it's edited to support the removal of richtext

megganeturner (approved):

PR #558, opened 2021-03-30
'Original' rendition for audio video

md8n:
# Description

The 'audio' and 'video' 'original' renditions have a different name, they are `original-audio` and `original-video` respectively.

So we need special code just for them ... when did you last see a `while` loop?

Also, the ugly differential casing the @md8n introduced has been wound back, because we we're looking directly at that method anyway.

### DIFF
--- a/src/ts/Implementations/Asset.ts
+++ b/src/ts/Implementations/Asset.ts
@@ -41,14 +41,14 @@ export class Asset extends PublishableItem {
      */
     get url(): string {
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        return Asset.Url(this!.entry as TAssetEntry);
+        return Asset.url(this!.entry as TAssetEntry);
     }
 
     /**
      * The platform specific media url of this asset item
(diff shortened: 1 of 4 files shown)


src/ts/Implementations/Asset.ts
```diff
@@ -161,23 +168,32 @@ export class Asset extends PublishableItem {
...
+                // however it would be nice to have it available in the near future
                 renditionSpec = "original";
                 break;
             default:
                 renditionSpec = "original";
         }
-        if (!asset.renditions[renditionSpec]) {
-            // Force it back to original if we don't have that rendition
-            renditionSpec = "original";
+        while (!(renditionSpec in asset.renditions)) {
```
PeteCoward, line 178: oh my, testing this now

md8n: I know right!

I went "but, but, no one is allowed to use while loops any longer!"

PeteCoward (approved):

PR #559, opened 2021-03-30
FE offline content

megganeturner:
Couldn't find a ticket for this one, but follows on from Peter's work in #550 

InVision: https://projects.invisionapp.com/d/main?origin=v7#/console/20251865/446572478/inspect?scrollOffset=3870

Lessons are available to "download" (cache) from a course page. A user can opt to download one lesson at a time, or the whole course.

I included a simple loading animation for the download button, but the download is lightning fast ⚡  (I tested with a slow 3G connection from the network tab in dev tools, but it was still very speedy) so you probably won't see it - I've left it in there in case someone is on a reallyyyy slow connection. You can change `download-state` to `download-state loading` to test this.

Screen Shot 2021-03-30 at 4 41 32 pm

Screen Shot 2021-03-30 at 4 42 47 pm


katevoss (approved):
Looks good, just a few small nitpicks:
* when the LessonDownload modal is open, I can still scroll the course page in the background
* I think there should be an X to explicitly close out of the modal, like you added for the fullscreen images

PeteCoward:
how hard would it be to add the same to Resources?

megganeturner:
> how hard would it be to add the same to Resources?

Shouldn't be too difficult! I'll work on this today

PR #561, opened 2021-03-30
Wrangling renditions preferred vs. smallest

md8n:
# Description

This handles the new rendition values that can show up.  It also does a limited amount of processing to determine 'preferred' and 'smallest' rendition when there is more than one available to determine an actual 'preferred' rendition to download.

### DIFF
--- a/src/ts/Constants.ts
+++ b/src/ts/Constants.ts
@@ -5,11 +5,25 @@ export const EMPTY_SLATE_BOOT_KEY = "empty_slate_boot";
 // Rendition Identifiers
 // - see canoe-backend repo - TranscodeDefinition.json
 // Image Renditions
-export const WEBP_RENDITION = "width-800|format-webp";
-export const JPEG_RENDITION = "width-600|format-jpeg";
+export const WEBP_RENDITION = "width-800|format-webp"; // .webp
+export const JPEG_RENDITION = "width-600|format-jpeg"; // .jpeg
 // Video Renditions
-export const WEBM_RENDITION = "webm-mono-360p-25fps-256kbit";
-export const MP4V_RENDITION = "video-streamcopy-mp4mux";
+export const ORIG_V_RENDITION = "original-video"; // Any extension
+export const H264_360P_RENDITION = "H264-360p-28"; // .mp4
+export const H264_480P_RENDITION = "H264-480p-28"; // .mp4
+export const H265_360P_RENDITION = "H265-360p-28q"; // .mp4
+export const H265_480P_RENDITION = "H265-480p-20q"; // .mp4
+export const VP9_360P_RENDITION = "VP9-360p-35q"; // .webm
+export const VP9_480P_RENDITION = "VP9-480p-20q"; // .webm
+export const WEBM_RENDITION = "webm-mono-360p-25fps-256kbit"; // .webm
 // Audio Renditions
-export const OPUS_RENDITION = "opus-mono-32kbit"; // AKA ogg
-export const MP4A_RENDITION = "audio-streamcopy-mp4mux";
+export const ORIG_A_RENDITION = "original-audio"; // Any extension
+export const OPUS_RENDITION = "opus-mono-32kbit"; // .ogg
+// Other Renditions
+export const ORIG_RENDITION = "original"; // Any extension
+
+// Inactive (Previous/Unused Transcodes)
+export const AV1_360P_RENDITION = "AV1-360p-48q";
+export const AV1_480P_RENDITION = "AV1-480p-32q";
+export const MP4MUXV_RENDITION = "video-streamcopy-mp4mux";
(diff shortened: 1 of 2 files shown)


src/ts/Constants.ts
```diff
@@ -5,11 +5,25 @@ export const EMPTY_SLATE_BOOT_KEY = "empty_slate_boot";
...
-export const WEBM_RENDITION = "webm-mono-360p-25fps-256kbit";
-export const MP4V_RENDITION = "video-streamcopy-mp4mux";
+export const ORIG_V_RENDITION = "original-video"; // Any extension
+export const H264_360P_RENDITION = "H264-360p-28"; // .mp4
+export const H264_480P_RENDITION = "H264-480p-28"; // .mp4
+export const H265_360P_RENDITION = "H265-360p-28q"; // .mp4
+export const H265_480P_RENDITION = "H265-480p-20q"; // .mp4
+export const VP9_360P_RENDITION = "VP9-360p-35q"; // .webm
+export const VP9_480P_RENDITION = "VP9-480p-20q"; // .webm
+export const WEBM_RENDITION = "webm-mono-360p-25fps-256kbit"; // .webm
```
boerwastaken, line 18: I disappeared this one. See https://github.com/catalpainternational/canoe-backend/blob/FT-Offlining/wagtailmedia_transcode/fixtures/TranscodeDefinition.json or https://admin.canoe-csp.catalpa.build/django-admin/wagtailmedia_transcode/transcodedefinition/ for the current definitions.

We should probably at some point agree on a mini-format.

src/ts/Constants.ts
```diff
@@ -5,11 +5,25 @@ export const EMPTY_SLATE_BOOT_KEY = "empty_slate_boot";
...
 // Audio Renditions
-export const OPUS_RENDITION = "opus-mono-32kbit"; // AKA ogg
-export const MP4A_RENDITION = "audio-streamcopy-mp4mux";
+export const ORIG_A_RENDITION = "original-audio"; // Any extension
+export const OPUS_RENDITION = "opus-mono-32kbit"; // .ogg
+// Other Renditions
+export const ORIG_RENDITION = "original"; // Any extension
```
boerwastaken, line 23: Also disappeared

src/ts/Implementations/Asset.ts
```diff
@@ -131,47 +163,108 @@ export class Asset extends PublishableItem {
...
+                        rendDefs.splice(
+                            rendCount,
+                            0,
+                            VP9_360P_RENDITION,
+                            VP9_480P_RENDITION,
+                            WEBM_RENDITION
+                        );
+                        break;
+                    case "Firefox":
+                        rendDefs.splice(rendCount, 0, WEBM_RENDITION);
```
boerwastaken, line 191: VP9 is the codec in webm. 
https://trac.ffmpeg.org/wiki/Encode/VP9

VP9 is the format of choice for Firefox (until AV1 starts working well, but there's also the question of whether it'll be hardware accelerated on older phones. VP9 is hardware-accelerated on many boards AFAIK)

src/ts/Constants.ts
PeteCoward: I was thinking to add the extension to the rendition dict in the manifest, would make life simpler no?

src/ts/Constants.ts
PeteCoward: we had handling for original to original-audio|video in a PR i merged this afternoon, @md8n does this delete that?

md8n: It replaces it - it didn't last long

src/ts/Constants.ts
md8n: A lot of them have the same extension, because that's the container.

src/ts/Implementations/Asset.ts
md8n: According to caniuse it's not supported yet.

src/ts/Constants.ts
boerwastaken: it'll either be mp4 or webm I suppose, I don't think we want any funky containers. Besides nothing should care much about the extension. And it's already in the source url.

But there is a case to be made for a more descriptive miniformat than the format's title. Width, height, runtime length, audio ka lae, codecs used, ...
Currently that's all stuffed into the transcode definition's title. We can live with that but it's not ideal in terms of conveying information to the client to let it make informed decisions on what to retrieve (without maintaining a full copy of the server's catalogue as constants in harder-to-update frontend code).

src/ts/Implementations/Asset.ts
boerwastaken: well it plays perfectly in my Firefox....
you might be looking at the wrong page.
According to caniuse 95.56% of browsers can play this. Everyone wants to play along and support a decent and royalty-free format. Everyone, except Apple.

I see you're adding VP9 for Chrome, but not for Firefox. Am I looking at it the wrong way?

PR #562, opened 2021-03-30
Fix up login banner

katevoss:
- When a user hasn't authenticated they see a banner to Login
- If they go to the Settings page then they'll see a button to Login
- When a user is logged in, the button logs them out
- Translation updates too


src/riot/Components/GuestBanner.riot.html
```diff
@@ -1,9 +1,10 @@
 <GuestBanner>
     <div if="{state.isBannerVisible}" class="{state.isBannerClosing ? 'close' : ''}">
         <a if="{!state.isBannerClosing}" class="cross" onclick="{closingBanner}"></a>
-        <p>{ TRANSLATIONS.guest() }</p>
-        <p>{ TRANSLATIONS.data() }</p>
-        <p><a class="btn btn-primary" href="#login">login</a></p>
+        <span>
+            <p>{ TRANSLATIONS.data() }</p>
+            <a class="btn btn-primary" href="#login">Login</a>
```
megganeturner, line 6: Should this also be TRANSLATIONS.login()?

megganeturner (approved):
Nice Kate, much neater banner! Just one word needs to be switched to a translation and this is good to go (:

src/riot/Components/GuestBanner.riot.html
katevoss: Good catch!

PR #580, opened 2021-03-31
579 JID multi choice feedback text bug

### LINKED ISSUE #579: Image multiple/single select questions break
_Also known as checkboxes and multiple choice._

The popup modal shows an icon only but no text feedback. An error can be seen in the console. Also, the text feedback "Find and check..." seems misplaced

The similar question types without image don't show this error.

---

correct answer | wrong answer | misplaced text
-- | -- | --
Screenshot 2021-03-31 at 17 28 59 | Screenshot 2021-03-31 at 17 28 29 | Screenshot 2021-03-31 at 17 29 39

**Console shows this when I click** `Check answers`
Screenshot 2021-03-31 at 18 14 53

megganeturner:
closes #579 

There's a bug in JID ([here](https://canoe-csp.catalpa.build/#/site/learn/test/lesson-test:content:2)) which is preventing users progressing past multiple-choice questions. I believe this is due to the format the FE expects from the api not lining up with the API (I'm not certain why this is only happening in JID).

This change adds some extra handling for when the `feedback` attribute is not present in an answer response, which should hopefully resolve the issue.

### DIFF
--- a/src/riot/Lesson/TestPopupModal.riot.html
+++ b/src/riot/Lesson/TestPopupModal.riot.html
@@ -53,22 +53,34 @@
 
             correctTitleMessage(selectedAnswer) {
                 // title for correct answer feedback
-                return selectedAnswer.feedback.title || gettext("Great");
+                if (selectedAnswer.feedback) {
+                    return selectedAnswer.feedback.title || gettext("Great");
+                }
+                return gettext("Great");
             },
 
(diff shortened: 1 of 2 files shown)


katevoss (approved):
Ah yes, this was my fault. The reason it only happens in JID is that's the only place my hastily-made PR was deployed to (https://github.com/catalpainternational/canoe/pull/570). Good fix!

PR #581, opened 2021-03-31
Feedback data includes course name and comment field.

sandroroux:


src/ts/Implementations/PublishableItem.ts
```diff
@@ -35,19 +35,19 @@ export abstract class PublishableItem {
      * @throws Error if network failure or response not OK
      */
     async getResponseFromNetwork(): Promise<Response> {
-        logger.log("using network for %s:%s", this.str, this.url);
+        // logger.log("using network for %s:%s", this.str, this.url);
```
sandroroux, line 38: @md8n Is there a way to disable these?

PeteCoward: Just jump into Logger, and disable all logs by commenting out one line.
I intended to add capability in Logger.ts to control them in config and runtime but had to swtich to doing something more targetted to real progress.
Being able to deploy to staging with listen this this named logger at this log level .
Being able to on production enter something in the console which sets those log level and name filters.

src/riot/Lesson/LessonFeedback.riot.html
```diff
@@ -150,27 +156,34 @@ <h5>{ TRANSLATIONS.value_feedback() }</h5>
...
-                const lesson = this.props.lessonName;
+                const { lessonName, courseName } = this.props;
                 const feedback = feedbackString;
-                const feedbackComment = document.getElementById("feedback").value;
+                const comment = document.getElementById("feedback").value;
 
-                saveFeedbackToIdb({ lesson, feedback, date: new Date() });
+                saveFeedbackToIdb({
+                    lesson: lessonName,
+                    course: courseName,
```
PeteCoward, line 165: please use the Course id, not the name. The name can change.
Send the name if you want, but in the analysis views we should always use the pk to look up the name in a join

src/ts/Implementations/PublishableItem.ts
PeteCoward: Also, you can just set your log level filter in the dev tools to hide them. And set a filter on contents to filter for the ones you want to see.

PeteCoward (changes requested):
- add / use course id
- revert logger comment outs

src/riot/Lesson/LessonFeedback.riot.html
sandroroux: I agree with the idea, but implementing the JOINs before I switch to hamahon may be tough. Let's chat about what my goal for this should be.

PeteCoward: for now, just add the id in here, then we can use either in the backend, and switch to the better method when we've worked out the sql required

sandroroux: Will do.

sandroroux: Done.

src/riot/Lesson/LessonContent.riot.html
```diff
@@ -89,8 +89,8 @@ <h3>{state.card.question}</h3>
     />
     <LessonFeedback
         if="{ !state.cardNumber && state.stage == 'feedback' }"
-        lessonName="{props.lesson.title}"
-        courseName="{props.lesson.course.title}"
```
PeteCoward: in this situation we should just pass the lesson through, and access the attributes from inside the tag, not a problem fo now though

PeteCoward (approved):

src/riot/Lesson/LessonFeedback.riot.html
```diff
@@ -149,28 +158,44 @@ <h5>{ TRANSLATIONS.value_feedback() }</h5>
...
             sendFeedback(feedbackString) {
-                const lesson = this.props.lessonName;
+                const { lessonId, courseId } = this.props;
+                const { comment } = this.state;
                 const feedback = feedbackString;
-                const feedbackComment = document.getElementById("feedback").value;
 
-                saveFeedbackToIdb({ lesson, feedback, date: new Date() });
+                saveFeedbackToIdb({
+                    lesson: lessonId,
```
boerwastaken, line 176: on the backend, for analysis, we'd really like a page revision as the sole topic. No need for lessonId or courseId, as the backend can tell that from the page revision. It's not forbidden to leave them in though, if needed clientside.

boerwastaken, line 176: Does this end up in a post to the UserAction endpoint? On the backend, for analysis, we'd really like a page revision as the sole topic. No need for lessonId or courseId, as the backend can tell all that (and more, basically everything there is to know) from the page revision. It's not forbidden to leave them in though, if needed clientside.

PR #601, opened 2021-04-19
592 lesson card size fixes

megganeturner:
Mobile browsers have menus that can appear/disappear only on scroll. They take up a lot of space and sometimes don't pay attention to `100vh` calculations, causing things to appear 'below the fold' (this is especially frustrating when your navigation is at the bottom of the screen).

This PR adds changes to detect the OS/browser of a device, and adjust the height of the lesson cards slightly so that the bottom navigation in lessons doesn't disappear below the bottom menu.

I've tested these changes on lots of devices/browsers across browserstack & it seems to do the job!

Safari on iOS before:

Safari on iOS after:

Chrome on Android before:

Chrome on Android after:

### DIFF
--- a/src/riot/Components/LessonFrame.riot.html
+++ b/src/riot/Components/LessonFrame.riot.html
@@ -26,7 +26,7 @@
             <span class="cross gray"></span>
         </a>
     </TopMenu>
-    <div class="frame-background">
+    <div class="frame-background {userAgent()}">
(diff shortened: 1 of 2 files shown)


PeteCoward (commented):
Ideally this kind of stuff should go in our platform detection code, see `PlatformDetection.get_platform` and it's use in Settings.riot.html`
Might be worth looking to see if that supports this use?

PeteCoward (approved):

PR #603, opened 2021-04-19
Minor clean out of some redundant stuff

md8n:
# Description

Removes some previously deprecated code that's definitely no longer used


PeteCoward (commented):
I may merge this after the actions_in_redux

PeteCoward (approved):

PR #627, opened 2021-04-20
Support Django session auth in Canoe

PeteCoward:
deployed to
https://canoe-redesign.temp.build/
https://admin.canoe-redesign.temp.build/
and
https://jid-staging.temp.build/
https://admin.jid-staging.temp.build/
and
https://pep-staging.temp.build/
https://admin.pep-staging.temp.build/

see https://github.com/catalpainternational/jid/pull/3

- sets `credentials: "include"` on fetches
- refactors initialisation
  - does synchronous initialisation setting state pre mount
  - mounts riot
  - does asynchronous initialisation setting state as it completes
  - riot reports that state through loading dots with messages
 - remove unnecessary service workering
 - stops sending some things to GA that we can now send to our servers


src/app.js
```diff
@@ -11,20 +11,32 @@ import { initialiseUserActions } from "js/UserActions";
...
+
+// Asynchronous initialization
+// Initialise things that require network calls or async ops
+initialiseIdentity()  // Am I logged in
+    .then(() => {
+        // perform independent actions that require login in parallel
+        return Promise.all([
+            initialiseUserActions(), // read actiion from api and idb
+            initialiseFeedback(),    // send any unsent feedback
+            initialiseCertChain(),   // initialise the appelflap sharing cert
```
PeteCoward, line 38: @md8n this I'm not sure will work when it drops into appelflap, but I think the right place for the call to initialise it is here.
I think we can detect isInAppelFlap above in the sync section, but need to request a cert here asynchronously.
We can decide whether it's worth waiting for the return before routing or not.

src/riot/App.riot.html
```diff
@@ -1,14 +1,19 @@
 <App>
+    <LoadingDots
+        if="{ !isAuthenticationDetected }"
+        extrastyleclasses="dark app"
+        message="Checking who you are..."
```
PeteCoward, line 5: need to tRANSLATION this

sandroroux (approved):

PR #634, opened 2021-04-23
Teaching Activity completion persistance

### LINKED ISSUE #458: I can mark Learning Activities as complete
## Conditions of Satisfaction
* Marking the Activity as complete will take me to the feedback screen
* This option is always accessible while navigating the Activity

## Resources
* Engineering estimate: Days (2)
* * Reference design: [Learning Activities - Detail](https://projects.invisionapp.com/share/6V102V1VURS9#/screens/443710125)

PeteCoward:
Fixes #458 

This uses the `page.complete` mechanism to query and set completion of Teaching Activities.

When `page.complete` is set to true the mechanism built into all `Page` instances in previous PRs, stores the completion in idb, sends it to the api ( which will resync if offline when you come back online ) and sets redux state.
When you get `page.complete` you read from redux state.
On initial page load we read from the api into idb and from idb into redux state.

- completing all activities in a topic completes that topic ( and uncompleting one, uncompletes the topic )
- LongCard now supports the same behaviour as Card reducing prop passing ( progress bar capability is present but commented )

TEst:
- set a lesson activity to complete, refresh, it should still be complete
- set the same lesson activity to incomplete, refresh it should still be incomplete
- login, logout, the same completions should be set
- check the completion ticks on the topic page, and the teaching home comply with expectations

You can also check the data stored in your actions idb, and your UserAction server side database


src/riot/Teaching/LongCard.riot.html
```diff
@@ -1,23 +1,47 @@
...
-                src="{getCardImage()}"
+                src="{ getCardImage() }"
                 class="card-hero-image"
             />
             <span class="card-status"><span class="icon"></span></span>
         </div>
         <div class="card-text">
-            <h5>{props.title}</h5>
+            <h5>{ title() }</h5>
+            <!-- <progress
```
katevoss, line 13: Why is this here but commented out?

katevoss (changes requested):
I marked a Teaching activity as complete, refreshed the page, and it was incomplete again. I also don't see any entries in my actions IndexedDB

https://user-images.githubusercontent.com/2395211/116599326-f1ac8000-a8f5-11eb-80e9-1ce5738500e8.mov

src/riot/Teaching/LongCard.riot.html
PeteCoward: I wasn't sure if we wanted it in the long card, but I wanted to make sure it supported it. I can remove

PeteCoward:
I merged latest, and removed the commented out code
I see your screencapture, and raise you mine. It works fine on my machine.

https://user-images.githubusercontent.com/1092029/117247524-5d07cc00-ae79-11eb-8505-8cc263c6b67b.mov


Will be on https://pep-staging.temp.build for your review too

katevoss (approved):

PR #640, opened 2021-04-26
594 video thumnails

### LINKED ISSUE #594: Fix: video thumbnails is not being shown
Thumbnail of the video is not being displayed.
Tester was using:
IPhone 11 and Safari

Also, a tester with a good bandwidth and connection, reported that the video took a long time to load on both the smartphone and computer

Screen Shot 2021-04-12 at 5 18 45 PM

PeteCoward:
Fixes #594
Accompanied by https://github.com/catalpainternational/canoe-backend/pull/159

- Simplifies access to Asset from Page class enabling simpler riot code
- Exposes thumbnail and duration members on Asset
- Simplifies and standardises riot media cards, uses new thumbnail and duration members

### DIFF
--- a/src/riot/Lesson/VideoCard.riot.html
+++ b/src/riot/Lesson/VideoCard.riot.html
@@ -1,8 +1,10 @@
 <VideoCard>
     <div>
         <video
-            if="{ state.src }"
-            src="{ state.src }"
+            if="{ asset() }"
+            src="{ asset().url }"
+            duration="{ asset().duration }"
+            poster="{ asset().thumbnail }"
             width="100%"
             crossorigin="anonymous"
             controls="true"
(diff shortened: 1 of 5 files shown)


PeteCoward:
@megganeturner this touches ( quite heavily ) the riot code in ImageWrapper, AudioCard, and VideoCard hence requesting your review
@md8n this touches ( lightly ) the Asset access code from Page hence requesting your review

Remember to pull catalpainternational/canoe-backend#159 before running

megganeturner (approved):

src/ts/Types/PublishableItemTypes.ts
```diff
@@ -4,16 +4,11 @@ import {
     TItemType,
 } from "ts/Types/CanoeEnums";
 
-/** The id fields for a manifest, page or asset item */
-export type TItemId = {
-    api_url: string;
```
md8n, line 9: We still have api_url in use in `CacheItem.ts` (i.e. everything) and `Page.ts` - so I'd recommend leaving this alone

md8n (commented):
Have a look where we are using `api_url`, we'll probably still need it defined in the type, or at least preferably want it explicitly defined in the type.

src/ts/Types/PublishableItemTypes.ts
PeteCoward: I think only one type should actually an api_url attribute: `TWagtailPageData`

PeteCoward: The CacheItem code needs a careful look at, I think everything it does is now available on Manifest.ts and Page.ts

PeteCoward: reverted this commit. The whole typings of those data objects is out of sync with reality right now, leaving it for another PR

src/riot/Lesson/VideoCard.riot.html
```diff
@@ -1,8 +1,10 @@
 <VideoCard>
     <div>
         <video
-            if="{ state.src }"
-            src="{ state.src }"
+            if="{ asset() }"
+            src="{ asset().url }"
+            duration="{ asset().duration }"
```
boerwastaken, line 6: Duration is a readonly attribute it seems:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video

PeteCoward: I wondered about that, but wasn't sure it meant it can only be set from the src once loaded, or just cannot be updated after the source was loaded. Writing it didn't cause an error. Easy to remove if we see it happening.

PR #642, opened 2021-04-27
Make icon colors themeable

katevoss:
As a part of theming Hamahon, I updated some of the icon colors so that they are themable https://github.com/catalpainternational/hamahon/issues/142

For example these icons:
image


PeteCoward (commented):
Will leave for Meggan. My thoughts are I wonder if we can inherit this from the project config background color or theme colour

megganeturner (approved):

PR #644, opened 2021-04-28
Set pdf download link to 'download'

PeteCoward:
no new tab, just download, not sure this is what is required on the device.
the old behaviour is open in a new tab and download, the new behaviour is just download.
Possibly worth thinking about opening in a new tab but displaying, or just same tab and displaying


pedro16v:
@Juliana-t please let @PeteCoward know where some relevant content is to test this code please

Juliana-t:
@pedro16v already sent that info to @PeteCoward :)

PR #655, opened 2021-04-29
Add wagtail page revision ID to completion actions

PeteCoward:
Adding it to the exam scores is a little more complicated, looking into it...


boerwastaken (approved):
:+1:

PR #661, opened 2021-05-03
useractions: fix :bug: , and store timestamps as integer (ms past unix epoch)

boerwastaken:
- Fix bug where "one year ago" meant "one year ago" only *once a year*, on new year's eve.
- use unixtime milliseconds for the `date` timestamp (which should've been called `timestamp` imo); JS's `.toString()` is awful for storing datetimes, also if the browser locale is invariant, which it won't be. Any problem with this was swallowed away serverside as the timestamp would be replaced by the server's if the client-supplied timestamp couldn't be parsed, so we couldn't even see how often this went wrong; plus the origin information of the timestamp (server or client) was discarded. Working on fixing all that up in canoe-backend on the way to solidifying the ground for data analysis.

### DIFF
--- a/src/js/actions/ActionsStore.js
+++ b/src/js/actions/ActionsStore.js
@@ -33,7 +33,7 @@ export const EXAM_SCORE_TYPE = `${EXAM_ACTION_TYPE}.finalScore`;
 export const saveAndPostAction = (actionType, data) => {
     const action = {
         type: actionType,
-        date: new Date(),
+        date: new Date().valueOf(),
         uuid: make_uuid32(),
         ...data,
     };
--- a/src/js/actions/actions_idb.js
+++ b/src/js/actions/actions_idb.js
@@ -1,6 +1,7 @@
 import { openDB, deleteDB, wrap, unwrap } from "idb";
 
 const action_store_name = "actions";
+const ONE_YEAR = 365 * 24 * 60 * 60 * 1000
 
 // store an action of a certain type
 export async function writeAction(action) {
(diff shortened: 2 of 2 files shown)


sandroroux:
@boerwastaken Looks good, the changes make sense. Why was ignoring actions from before the current year a bug?

boerwastaken:
> Why was ignoring actions from before the current year a bug?

The comment reads:
```js
// return false - no change made; ignore actions from more than a year ago
```
It doesn't say "before the current year". It says "more than a year ago".
Either the behaviour is a bug or the comment is a bug, because the code didn't do what the comment promised it'd do ;-)

I'm siding with the comment because setting different criteria for event discards depending on which side of the sun the earth is on is behaviour that raises my eyebrows :raised_eyebrow:

src/js/actions/actions_idb.js
```diff
@@ -38,11 +39,8 @@ export async function ensureAction(action) {
...
 
-    const thisYear = new Date().getFullYear();
-    const beginningOfThisYear = new Date(thisYear, 0, 1);
-
-    // return false - no change made; ignore actions from prior years.
-    if (key || action.date < beginningOfThisYear) {
+    // return false - no change made; ignore actions from more than a year ago
+    if (key || action.date < (new Date() - ONE_YEAR)) {
         return false;
     }
```
PeteCoward, line 45: There is no reason for this to stay here, it's old Matenek/Olgeta code that I don't think we need anymore
```suggestion
```

src/js/actions/ActionsStore.js
```diff
@@ -33,7 +33,7 @@ export const EXAM_SCORE_TYPE = `${EXAM_ACTION_TYPE}.finalScore`;
 export const saveAndPostAction = (actionType, data) => {
     const action = {
         type: actionType,
-        date: new Date(),
+        date: new Date().valueOf(),
```
PeteCoward, line 36: I just need to verify that date isn't used anywhere else. The action is returned from the function for use in setting redux state for completions. I think we are good though

PeteCoward (approved):

PR #668, opened 2021-05-10
Sociallearning minorfixes

PeteCoward:
Backend api has change, this updates to change the frontend urls used
Also sometimes users have no first or last name ( usually us ) this just uses user_username instead if that happens

published to https://learn.canoe-redesign.temp.build


PeteCoward:
my name is actually `undefinedundefined`

Issue #630, opened 2021-04-22
Empty structure causes Canoe to fail


pedro16v:
If no items have been added to `Translatable Root` (in other words, if there's no home pages for Courses or Learning Activities) then Canoe had nothing to display. 

We could display a static page to make it fail more elegantly. Maybe a maintenance page that could be used somewhere else? @Juliana-t @KaraChesal @andrewwapling  ?

The browser console shows this:
`TypeError: Cannot read property 'getLanguagePageType' of undefined`

![Screenshot 2021-04-22 at 14 46 32](https://user-images.githubusercontent.com/2813677/115668831-89ecb700-a382-11eb-9aa8-eda9f4a404f7.png)

pedro16v:
I should add that this is something that no user will ever have to worry with unless we promote Canoe installations with absolutely no content.

PeteCoward:
We used to have a "low priority' label that was meant for situations like this

andrewwapling:
Hi @pedro16v yes it could be good to have a designated design solution for a' 404' or 'Oops, something went wrong' to handle this (and any other critical failure) in an elegant way. Shall we move this into 'to do' for 'design'? I can then create a global design direction for this so that we have it in the DSM if needed. Thanks, Andy

pedro16v:
@Juliana-t agrees with that. Thanks @andrewwapling .

andrewwapling:
Ok great thanks for confirming @pedro16v @Juliana-t , I'll create a design direction for this and post back soon.

andrewwapling:
Hi @pedro16v @Juliana-t ! @PeteCoward as promised please see 404 error screen for the unlikely event we encounter an error that's fatal to a users experience: https://invis.io/FR10XKHUJ5EB

pedro16v:
Thanks @andrewwapling. We'll try to add this to the pipeline as soon as we get rid of the current big milestones.

megganeturner:
Peter has added some basic error handling over in PR #677, and set up the site to handle three scenarios:
- manifest didn't load (server error - site content cannot be retrieved)
- the requested page doesn't exist (user error - usually produced by an incorrect url)
- any other unknown error

This design is great Andy, but I think we should remove the explicit `404` from the illustration (or have another illustration for server side issues)

[...]

PR #677, opened 2021-05-14
Handling some errors

### LINKED ISSUE #675: First boot throws if backend server is unavailable
On a195153d (current FT-Offlining branch), if I start the webpack frontend stuff but not the backend server, it'll throw like so:

```
Logger.ts?6da3:20 ContentItem: request failed for Site Manifest:http://127.0.0.1:8000/manifest/v1
Routing.js?2753:72 Uncaught (in promise) TypeError: Cannot read property 'getLanguagePageType' of undefined
    at eval (webpack-internal:///…rc/js/Routing.js:91)
    at Array.forEach (<anonymous>)
    at route (webpack-internal:///…rc/js/Routing.js:86)
```

My expectation would be that yes, it might hang and show a spinner (which could/should be way more informative, telling us it's trying to contact the backend server or something — remember, this stuff is supposed to be used *exactly* in locales with less than ideal network connections!).
But then when the backend server is reachable again, it should complete the boot. Currently it's dead in the water once this has happened, you have to reload the page (another reason to keep a reload button handy in the Canoe Android app).

PeteCoward:
fixes #675 but maybe needs more visual love

Handles and displays basic error messages for the following 2 errors

1. Manifest cannot be retrieved from the server or the cache ( turn your runserver off and empty your storage to reproduce )
2. Page cannot be found ( goto http://YOUR_FRONTEND_HOST:PORT#does_not_Exist to reproduce as long as you already have a manifest or your server is running )

### DIFF
--- a/src/js/Routing.js
+++ b/src/js/Routing.js
@@ -49,7 +49,14 @@ async function route(hashWith) {
         try {
             manifest = await getValidManifest();
         } catch (err) {
-            route = {page: {type: "manifest_error", error: `Manifest retrieval failure. Error: ${err}`}};
+            setRoute({
+                page: {
+                    type: "error",
+                    errorType: "no_manifest",
+                    error: err,
+                }
+            });
+            return;
         }
         if( !isAuthenticated() && REQUIRE_LOGIN )
         { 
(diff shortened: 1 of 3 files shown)


megganeturner:
Nice Peter! Much better than an infinite loading screen 😄  I added some very basic styles to more closely align #630 (but added some comments there - something I can come back to later)

megganeturner (approved):

PR #681, opened 2021-05-17
Feedback logging format update

PeteCoward:
@boerwastaken per your request updated feedback format to align with other events.

I want to check whether we can merge this more into the actions framework so the date an uuids are set in one place, will chat to @sandroroux  about it


src/riot/Lesson/LessonFeedback.riot.html
```diff
@@ -168,24 +167,13 @@ <h5>{ TRANSLATIONS.value_feedback() }</h5>
...
-                    date: new Date().valueOf(),
-                });
-
-                logLessonFeedback({
-                    lessonId,
-                    courseId,
-                    feedback,
-                    comment,
-                });
+                };
```
sandroroux, line 176: This looks good, but I'm not sure I agree with the semantics. In theory, a person can submit multiple pieces of feedback per page (and one doesn't replace the prior). Shouldn't the call site reflect that?

Something like:
```js
this.lesson.submitFeedback(...);
```

PeteCoward: Yes, i think you are correct, just semantics, it works like that but doesn't read like that because of the setter

PR #692, opened 2021-05-26
639 pep flash cards

megganeturner:
Requires https://github.com/catalpainternational/canoe-backend/pull/186 in the backend

After creating tips & linking them to the appropriate Plan / Teach tabs in wagtail, they'll come through to the front end:

The summary card (corresponds to the `Content` tab in wagtail) displays the title, short & long descriptions, and the card image if there is one:

Screen Shot 2021-05-26 at 3 40 17 pm
Screen Shot 2021-05-26 at 3 40 46 pm

Clicking into the 'View all tips' CTA takes you through the flash cards (`Tips Content` tab in wagtail) (this is effectively a lesson flow):
Screen Shot 2021-05-26 at 3 40 27 pm

At the end there is a completion screen which will redirect back to the activities page (with the correct tab selected):
Screen Shot 2021-05-26 at 3 40 35 pm


src/riot/Lesson/LessonContent.riot.html
```diff
@@ -97,40 +99,82 @@
...
                     finish: () => gettext("Finish"),
                 },
+
+                onBeforeMount(props, state) {
+                    const tipTabs = ['plan', 'teach', 'extend'];
+                    const returnTab = this.route.riotHash[1];
+
+                    if (returnTab && tipTabs.includes(returnTab)) {
+                        this.state.returnTab = returnTab;
+                    }
```
PeteCoward, line 116: putting things in the beforeMount is a little vulnerable to situations where you navigate between 2 pages where this tab is mounted ( which admittedly almost never happens ).
If you were on a page with this tag mounted, then navigated to another where this tag was mounted, this code would not run, the tag is updated not mounted. Not sure if that would cause bugs, just thought I should mention it as this is what I think everytime I see mount code in riot :)

src/riot/Teaching/FlashCard.riot.html
```diff
@@ -0,0 +1,49 @@
+<FlashCard>
```
PeteCoward, line 1: check indentation in this file please

src/riot/Teaching/TeachTab.riot.html
```diff
@@ -1,41 +1,62 @@
 <TeachTab>
```
PeteCoward, line 1: check indentation please

src/riot/Lesson/LessonContent.riot.html
megganeturner: Agreed! Have moved this into a getter :)

PR #703, opened 2021-06-01
Local appelflap testing setup

md8n:
# Description

Instructions about the precise needs of Appelflap when doing fully local testing


src/sw.js
```diff
@@ -79,6 +80,15 @@ registerRoute(
...
 
+registerRoute(new RegExp(ROUTES_FOR_REGISTRATION.subscribe), new NetworkOnly());
+registerRoute(new RegExp(ROUTES_FOR_REGISTRATION.subscribe), new NetworkOnly(), "POST");
+
+registerRoute(new RegExp(ROUTES_FOR_REGISTRATION.appelflapPKIsign), new NetworkOnly(), "POST");
+
+// webpack-dev-server communicates over this endpoint. Without this clause, the
+// service worker caches these requests and breaks webpack-dev-server.
+registerRoute(new RegExp(ROUTES_FOR_REGISTRATION.socketInfo), new NetworkOnly());
+
```
PeteCoward, line 91: network only should be the default according to my understanding of current code so no need for these. Am I wrong?

canoe-environment-default.js
```diff
@@ -3,7 +3,8 @@
 
 module.exports = {
     GA_TAG: false,
-    API_BASE_URL: "http://127.0.0.1:8000",
+    // To do local testing with Appelflap, API_BASE_URL MUST be "http://localhost:8000"
+    API_BASE_URL: "http://localhost:8000",
```
PeteCoward, line 7: Wicher seems able to run with this as 127.0.0.1, I think until we work out what is wrong with your settings we should hold off on these changes which affect everyone and mean we have to change domains and ports across projects.
Can you keep the changes you require to dev until then in local config changes?

src/sw.js
md8n: I haven't seen anything that says `NetworkOnly` is the current default. I'm pretty sure that has be a specific setting somewhere (normally in the sw code).

canoe-environment-default.js
md8n: Wicher has been able to get the differing port numbers working (from what I read), as opposed to `localhost` vs. `127.0.0.1`.

To test this GeckoView / FireFox must be used.

PeteCoward: from slack engineering_bero channel "I'm running the backend server on 127.0.0.1:8010, frontend on 127.0.0.1:8011"

src/sw.js
PeteCoward: Unless you specify a different strategy, then "Without a default handler, unmatched requests will go against the network as if there were no service worker present." https://developers.google.com/web/tools/workbox/reference-docs/latest/workbox-routing_setDefaultHandler.js?hl=en

canoe-environment-default.js
boerwastaken: True, https://catalpa.slack.com/archives/C0122BDJ7FA/p1623162345044500

PeteCoward:
closing as discussed on slack

PR #717, opened 2021-06-04
bugfix for rendering tip cards + style tweaks

megganeturner:
Bug surfaced in sprint review just now - tip cards were failing with a `cannot read property 'card_image' of undefined` error

Getters don't seem to have access to `props` when a component initially loads, so we need to be really explicit when using props in  conditionals

Also adds some very minor tweaks to styles

### DIFF
--- a/src/riot/Teaching/FlashCard.riot.html
+++ b/src/riot/Teaching/FlashCard.riot.html
@@ -36,7 +36,7 @@
 			},
 
 			get cardImageUrl() {
-				return this.props && this.props.card.card_image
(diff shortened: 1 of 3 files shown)


PeteCoward:
I opened an [issue](https://github.com/riot/riot/issues/2908) for this in riot. It is fixed in the latest riot

PeteCoward (approved):

megganeturner:
> I opened an [issue](https://github.com/riot/riot/issues/2908) for this in riot. It is fixed in the latest riot

Oh nice one!

PR #725, opened 2021-06-08
"new" course/lesson badges

megganeturner:
adds a 'new' badge for course/lesson cards
mocked in the new state for now, still TODO

Design reference: https://projects.invisionapp.com/d/main/#/console/20251865/429853515/inspect

### DIFF
--- a/src/ts/Implementations/Page.ts
+++ b/src/ts/Implementations/Page.ts
@@ -230,10 +230,9 @@ export class Page extends PublishableItem implements StorableItem {
 
     get isNew(): boolean {
         return (
-            !this.complete &&
             this.progressStatus === "not-started" &&
-            Number(new Date(this.version / 1000)) >
-                Number(Date.now() - 1000 * 60 * 60 * 24 * 7)
+            new Date(this.version / 1000).valueOf() >
+                Date.now() - 1000 * 60 * 60 * 24 * 7
         );
     }
 
--- a/src/ts/Implementations/Page.ts
+++ b/src/ts/Implementations/Page.ts
@@ -228,6 +228,11 @@ export class Page extends PublishableItem implements StorableItem {
         return `Page ${this.title}`;
     }
 
+    get isNew(): boolean {
+        // TODO implement this properly when the backend is done
+        return true;
+    }
+
     /**
      * Check if the page is in the correct cache
      * @returns true if this page, assets, and children is cached in the correct cache , false if not
--- a/src/ts/Implementations/Page.ts
+++ b/src/ts/Implementations/Page.ts
@@ -229,8 +229,12 @@ export class Page extends PublishableItem implements StorableItem {
     }
 
     get isNew(): boolean {
-        // TODO implement this properly when the backend is done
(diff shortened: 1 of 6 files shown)


src/ts/Implementations/Page.ts
```diff
@@ -228,6 +228,11 @@ export class Page extends PublishableItem implements StorableItem {
         return `Page ${this.title}`;
     }
 
+    get isNew(): boolean {
+        // TODO implement this properly when the backend is done
+        return true;
```
PeteCoward, line 233:
```suggestion
        return !this.complete && new Date(this.version / 1000) > Date.now() - (1000 * 60 * 60 * 24 * 7);
```

PeteCoward: This might show the badge if not complete and published less than a week ago, can you test this @megganeturner

src/ts/Implementations/Page.ts
```diff
@@ -229,8 +229,12 @@ export class Page extends PublishableItem implements StorableItem {
     }
 
     get isNew(): boolean {
-        // TODO implement this properly when the backend is done
-        return true;
+        return (
+            !this.complete &&
+            this.progressStatus === "not-started" &&
```
PeteCoward, line 234: `!this.complete` might be redundant, if it is not started, then it cannot be complete

src/ts/Implementations/Page.ts
```diff
@@ -229,8 +229,12 @@ export class Page extends PublishableItem implements StorableItem {
...
+            Number(new Date(this.version / 1000)) >
```
PeteCoward, line 235: reason for the Number casts? Just curious

src/ts/Implementations/Page.ts
megganeturner: Ah of course 🤦‍♀️

src/ts/Implementations/Page.ts
megganeturner: Prettier yelled about using the `>` operator to compare a `Number` to a `Date`

PeteCoward: ahh ok, might be more sensible to use the `Date.valueOf()` method rather than `Number()`

src/ts/Implementations/Specific/Course.ts
```diff
@@ -152,11 +152,11 @@ export default class Course extends Page {
     }
 
     get minimumExamScoreDecimal(): number {
-        return this.storedData.exam_pass_score / 100;
+        return this.storedData?.exam_pass_score / 100;
```
PeteCoward, line 155: I don't think this change is necessary, still compiles without it anyway? Did you do it to avoid compilation/prettier errors, or runtime bugs?

megganeturner: for a runtime bug - it was causing this console error and a hanging loader on the courses page

Screen Shot 2021-06-11 at 10 49 17 am

PeteCoward (approved):

PR #730, opened 2021-06-09
Re-enable the basic sync page functionality

md8n:
# Description

Re-enable the basic sync page functionality.  But not the actual sync functionality.

Remember that this will all go away and be largely hidden from the user's view, this is an interim step on that 'journey'.


md8n:
made redundant

PR #731, opened 2021-06-09
tidy up of sync / cache related stuff - temp only

md8n:
# Description

This was stuff that connects loosely with sync page functionality re-enabling, that I 'cleaned up' as I was reviewing how to turn back on the sync functionality.

Ultimately most of this will 'go away', but some of the stuff (e.g. the bits related to 'url' of the cached item) will definitely continue and have a direct relationship to sync behaviour.


md8n:
made redundant

PR #732, opened 2021-06-09
Override toString (don't use `get str`)

md8n:
# Description

Minor fix changing custom `str` `get` accessor and replacing it by overriding `toString()`, which is the preferred Javascript approach for this kind of thing.

Plus two other very small tweaks in preparation for next step - which is exploring `Availability` value and determining if it is still relevant, before moving on to actual Publishing.


md8n:
made redundant

Issue #662, opened 2021-05-05
Portrait images take up too much vertical space


pedro16v:
- [x] portrait images take up too much vertical space - is there a way to crop this to be square (with option to expand?)

pedro16v:
The embedded video requirement now has its own issue here #761

pedro16v:
> marking an activity as complete does not persist, and does not show any status on the cards

This one was solved as part of some other general fix to Bero.

pedro16v:
This one is being solved by using a recommendation guide for now

PR #734, opened 2021-06-09
max-height for images with option to expand, hide footer when expanded

megganeturner:
Part of #662 

Set a max-height for images in cards, to save on vertical real estate for portrait images.

shorter in the card:

with option to expand and see full height of image:

also fixes a bug i noticed on activity pages which had the footer on top of the full-screen image

### DIFF
--- a/src/riot/Components/ImageWrapper.riot.html
+++ b/src/riot/Components/ImageWrapper.riot.html
@@ -62,6 +62,15 @@
                 },
 
                 toggleImageFullScreen() {
+                    // stacking rules mean we can't position the image over the top of the footer,
+                    // so we cheat by making the footer invisible when we're in full-screen mode
+                    const footerEl = document.getElementsByTagName('footer')[0];
+                    if (footerEl && footerEl.style.visibility !== 'hidden') {
+                        footerEl.style.visibility = 'hidden'
+                    } else {
+                        footerEl.style.visibility = '';
+                    }
+
                     this.update({ imageFullScreen: !this.state.imageFullScreen });
                 },
             }
--- a/src/scss/_lesson.scss
+++ b/src/scss/_lesson.scss
@@ -278,6 +278,8 @@ ActivityPage {
         margin-right: -20px;
         margin-top: -30px;
         border-radius: 8px 8px 0 0;
+        max-height: 85vmin;
+        object-fit: cover;
     }
 
     img.richtext-image.full-screen {
@@ -319,6 +321,10 @@ ActivityPage {
         display: flex;
         align-items: center;
         justify-content: center;
+
+        > span {
+             mix-blend-mode: difference;
+        }
     }
 
     ImageWrapper .contract-button {
@@ -326,10 +332,6 @@ ActivityPage {
         position: fixed;
         top: 1em;
         right: 1em;
-
-        > span {
-             mix-blend-mode: difference;
-        }
     }
 }
 


PeteCoward:
I'm a little worried that this will just cause another issue "I uploaded this image and now it looks ugly because it's cropped to square". Your ginger example is the happiest of happy paths for this. Imagine the behaviour with an actual portrait, or a full height image of a person.
It might be worth bringing this up in a design sync.

boerwastaken:
One thing that comes to mind is to force content authors to set a Wagtail focal point / square crop for too-rectangular images, and then we serve these focal point / square coordinates out through the manifest. And then the frontend needs to do a bit of arithmetic with all that to get the square view to center on the right part of the image. I made this sound easy, but I don't know if Wagtail is flexible enough to allow implementing all this.

Juliana-t:
Yes, agree with @PeteCoward - not sure if it is making more harm than good (if we are not making it easy/mandatory to set a focal point for the image). This is only on resources or lesson cards/content cards as well? If the second, that would be a big drama for our clients and we should not proceed with this solution.
Bringing @andrewwapling to the discussion!

PeteCoward:
Card images are delivered to the frontend at set ratio, and so focal points are applied in their rendering. The image must be cropped if it is not at that card ratio ( twice as wide as it is high ). The focal point guides wagtail to not crop out the important bit.

Lesson content images are delivered to the frontend as the ratio they are uploaded at, and so focal points are not relevant for their rendering. The image is just resized, not cropped in any way.

We can deliver any selected focal point down to the client to help in the client side "cropping" of the delivered image.

But really we should look at specific examples of portrait sized content to advise this choice. Most of the time just resizing the image before upload might be the best option. If you really WANT an image "taller" than the device display ratio, you probably want it displayed full height by default ( users will not necessarily see/use the expand ). If the problem is that it hides test content, then maybe we could look at options to put the text content above the image?

andrewwapling:
Hi @megganeturner and team, Firstly great work on the initial implementation.

There were some comments floating around against other GH issues that covered this very same topic around image rules and requirements and user controlled cropping with the CMS.

As far as the DSM stands currently for the activity cards and the resource detail screen we have catered for images that utilise the classic 16:9 ratio. The rationale behind this from a design perspective is to both conserve vertical space, create a clean and consistent UI and also to utilise a well known and familiar ratio for images i.e. refer social media titans Facebook, Twitter et al. The other benefit is that this format has the ability to work equally well for videos and images alike as it is a well used format.

16:9 example: https://invis.io/J9Z5F9LDQV8

This being said we've also accommodated the option for a square option should users prefer this. The rationale for this is likewise to conserve vertical real estate.

Square example: https://invis.io/TUZ5F8WXZ28

In both of these scenarios users can easily tap on the expand icon to view a full un-cropped version of the image in a modal window which can be important in situations where more highly detailed diagrams and infographics are utilised.

From a usability perspective I would highly recommend adding the ability to add cropping preferences in the CMS. This would avoid scenarios like @PeteCoward pointed out where users upload images that may not make sense when cropping to a default centre focal point.

Apart from adding custom cropping (which may be complex to implement) we could also offer a range of predefined crop focus settings (as @boerwastaken has alluded to above) such as:

• top left
• top right
• centre
• bottom left
• bottom right etc. 

This could be represented in the CMS as a simple graphic tool selector control using arrows with a toggle that allows a user to select their preferred ratio. This would give content moderators quick and intuitive control over how they would like images displayed. By offering users the option to select 16:9 or square formats it would also go some way to help achieve maximum flexibility for a range of image types, without overwhelming content moderators with exhaustive custom image cropping controls. In the vast majority of instances I believe the predefined set of cropping focal points in the CMS would certainly go a long way to help avoid undesirable image cropping on the front end.

PeteCoward:
> I would highly recommend adding the ability to add cropping preferences in the CMS

The problem is not that the wagtail focal point is difficult to use (it exists but people don't use it, for lesson content images it does nothing so why would they?)

We don't crop images to square or 16:9 because originally we assumed that most people would want their whole image shown as large as possible ( device width limited ). This could be important with graphics and charts.

If we do want to force all images to be cropped to 16:9 or square for layout reasons, we would do that client side ( so we don't have 2 images being downloaded and cached ). But this will probably result in people saying "My image is not showing how I want it, I don't want users to have to click to see the un-cropped version".

My advice for this is to make a decision based on actual known content in JID or PeP. There is no example in the original issue which we can look at to say "Here is a valid use of a long portrait ratio image in a card... for this example we should display it like this..." That decision might not be that it should always be 16:9 or square on first view.

Juliana-t:
Thanks @PeteCoward and @andrewwapling 
Andrew, I understand the design perspective on maintaining consistency across cards + saving vertical space. But the reality, and having in mind a HCD approach, not always content/information can feet on the 16:9 nor on the square standard format. When we have illustrative elements that we can manipulate, we are trying to get as close as the 16:9 ratio, but in a lot of cases, this isn't a feasible solution. As Peter said, especially on graphics/tables but also on some other cases where the image is the key piece of content of a card. In those cases, cropping an image has a negative impact on the core functionality of Bero - providing elearning content/passing a concept to the learners.
Andy, please see some examples below.

Having said that, it is my opinion that we shouldn't be cropping images unless it is a 'clear' action from the content management side. If we are hiding important parts of an image on the app, the learner may skip relevant concepts/information and that is really not a good pedagogical design principles (don't assume learners will clip on the button to enlarge images in all cards, most of them will probably never or almost never use that feature).
C2-P6
C3A-P64
C4-P47
C6-P6

C3A-P24

C2-P34

andrewwapling:
Hi @Juliana-t @PeteCoward @megganeturner @boerwastaken @pedro16v thanks for raising your concerns here. Thanks also for supplying actual content, this is really helpful.

In an effort to bring this into the real world for us to review in context I've created a dedicated prototype that uses this real world content so that we can see how it's working within the design structure. This will help us ensure the rules we are implementing provide the best experience for people in the widest range of scenarios.

Please see image test prototype: https://invis.io/RU116Y5I9SV3

Based on this test primary recommendations and observations:

• You'll note that the vast majority of detailed images work best in the square the format. No cropping is required here as the image can be resized in proportion to the width of the container and centred both vertically and horizontally to ensure equilibrium
•  Less detailed photographic based imagery is best suited to 16:9 ratio as it reduces unnecessary invasion of vertical space
• Portrait based imagery such as A4 or Letter sized images can suffice with a square format however would be more legible at an extended portrait based size that invades additional vertical space. No cropping is required in any of these scenarios here as the image would be resized in proportion to the height of the container and vertically and horizontally centred to achieve equilibrium

Secondary recommendations and observations:

• Square is universally the best format for all image types (photographic, detailed and portrait orientation). Portrait based images in this format will be impacted by lack of legibility and accessibility issues if heavily detailed. The enlarge view can counter this
• Heavily detailed images should be discouraged for the best UX from a HCD perspective as they run the risk of being illegible and cause accessibility issues. ALL images, diagrams and infographics should be created with a mobile first methodology to enhance the learning experience for users. With the caveat that this may not always be possible though should be aspired to. The vast majority of these images clearly fail accessibility tests from both a text size and colour contrast perspective 
• Would recommend some smart scripting to decipher if an image is portrait in orientation and render this image in the portrait container to enhance legibility and accessibility
• 16:9 format is best suited to either photographic, decorative or wide format landscape orientation images
• No cropping is required in any of the above scenarios (16:9 - 310px x 174px, Square - 310px x 310px, Portrait - 310px x 370px)

Hope this helps and look forward to hearing your thoughts, Andy

PR #735, opened 2021-06-09
Use BACKEND_BASE_URL where possible

md8n:
# Description

While doing work on restoring isAvailable it became apparent that `process.env.API_BASE_URL` was being used in several places when it should be `BACKEND_BASE_URL` everywhere except for the actual canoe-environment.js and the urls.js

This grabs the files that otherwise have nothing with the upcoming isAvailabile work and ensures they're standardised to `BACKEND_BASE_URL`


md8n:
Made redundant

PR #736, opened 2021-06-09
Get all the `url`s correct, critical for cache keys

md8n:
# Description

Over time the member `fullUrl` largely got replaced with just `url`, and `api_url` got a little lost along the way.

This just gets them all refreshed.

There is also some cleaning up of JSDoc comments (saving a few lines here and there).

This is part of the work towards `isAvailability` and then `isPublishable` (i.e. syncing)


md8n:
Made redundant

Issue #733, opened 2021-06-09
Re-enable wagtail Preview


PeteCoward:
OK this is complicated, stay with me...
Wagtail preview works by 

1. hitting [preview_on_edit post](https://github.com/wagtail/wagtail/blob/v2.10.2/wagtail/admin/views/pages.py#L848) validating the page form, and saving the form data into the django session
2. if valid open [preview_on_edit get](https://github.com/wagtail/wagtail/blob/v2.10.2/wagtail/admin/views/pages.py#L861) in a new tab which reads the page form from the django session ( and validates it ) the saves it with commit=False (not saving to db)
3. calls [page.make_preview_request](https://github.com/wagtail/wagtail/blob/8d60e98d71a7332af2f08d5b9d75303b0af679af/wagtail/core/models.py#L1560)
4. make_preview_request builds an internal request, served by [page.serve_preview](https://github.com/wagtail/wagtail/blob/8d60e98d71a7332af2f08d5b9d75303b0af679af/wagtail/core/models.py#L1702) which normally would just set request.is_preview and ensures no caching and forwards to `page.serve` ( which detects preview and reads data from the session ) 
 
BUT this is where `wagtail_headless_preview` comes in and because our pages inherit from `HeadlessPreviewMixin` it overrides [page.serve_preview](https://github.com/torchbox/wagtail-headless-preview/blob/main/wagtail_headless_preview/models.py#L94) to 
1. save to the db [PagePreview model](https://github.com/torchbox/wagtail-headless-preview/blob/main/wagtail_headless_preview/models.py#L12) the session, and page data and a token
2. renders a [template](https://github.com/torchbox/wagtail-headless-preview/blob/main/wagtail_headless_preview/templates/wagtail_headless_preview/preview.html) which iframes the `settings.HEADLESS_PREVIEW_CLIENT_URLS` site with token and content type in the query
3. The frontend site then used to detect that as a preview view, and hit the preview api which overrides [get_object](https://github.com/catalpainternational/canoe-backend/blob/RC-3.0/catalpa_wagtail_tools/previewable_api.py#L39) to apply the saved preview info

So given the above, where do we go from here...

I figured it might be nice to pop preview in an iframe suitably sized to display as it looks on a device.
I figured as we now use the same session auth front and back, we can dispense with wagtail_headless_preview, always nice to drop a dependency

How do we do that?

[...]

PR #737, opened 2021-06-09
Wagtail Preview frontend

PeteCoward:
supporting https://github.com/catalpainternational/canoe/issues/733
together with https://github.com/catalpainternational/canoe-backend/pull/212
implemented in JID https://github.com/catalpainternational/jid/pull/14

If the canoe url has the "preview" serach param, then 
- always request a new manifest, do not use cache
- pass the "preview" search param to the manifest api
- always request a page api update, do not use cache
- pass the "preview" search param to the page api


sandroroux:
@PeteCoward I'm still seeing the same problems. I'll describe them here.

## Can't change course image.
I change the course image but don't see it appear in the preview.
out

PR #740, opened 2021-06-10
Clean TS imports and JSDoc

md8n:
# Description

Across the TypeScript code this cleans up the imports used, to ensure better intellisense in various IDEs.

The `ts` webpack src path is only retained for some of the JS code, Riot pages, and Typings. 

It also cleans up several JSDoc comments, ensuring consistent formatting and fixing some spelling mistakes and grammatical errors, this is primarily cosmetic, but it does mean that several of the comments will make more sense, again these turn up in intellisense in various editors.

### DIFF
--- a/src/ts/Implementations/Asset.ts
+++ b/src/ts/Implementations/Asset.ts
@@ -88,7 +88,7 @@ export class Asset extends PublishableItem {
         return this.#page.cacheKey;
     }
 
-    /** The options to make an asset request*/
+    /** The options to make an asset request */
     get requestOptions(): RequestInit {
         return {
             cache: "force-cache", // assets are (almost always) invariant on filename
--- a/src/ts/Implementations/Asset.ts
+++ b/src/ts/Implementations/Asset.ts
@@ -46,8 +46,7 @@ const RENDITION_PREFERENCE = [
  * whereby we will take the preferred rendition over the smallest rendition */
 const SIZE_DIFF_PERCENT = 0.02;
 
-/** A asset ( binary resource ) than can be cached
- */
+/** A asset ( binary resource ) than can be cached */
 export class Asset extends PublishableItem {
     /** id of this asset in the manifest page */
     #id: string;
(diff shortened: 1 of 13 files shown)


src/ts/Implementations/Manifest.ts
```diff
@@ -58,7 +53,7 @@ export class Manifest extends PublishableItem implements StorableItem {
         return reqInit as RequestInit;
     }
 
-    /** StorableItem implementations */
+    // StorableItem implementations
```
PeteCoward, line 56: this is a point where we can use `#region` as in [here](https://github.com/catalpainternational/canoe/blob/RC-3.0/src/js/redux/Interface.js#L18)

src/ts/Implementations/Asset.ts
```diff
@@ -46,8 +46,7 @@ const RENDITION_PREFERENCE = [
  * whereby we will take the preferred rendition over the smallest rendition */
 const SIZE_DIFF_PERCENT = 0.02;
 
-/** A asset ( binary resource ) than can be cached
- */
+/** A asset ( binary resource ) than can be cached */
```
PeteCoward, line 49:
```suggestion
/** An asset ( binary resource ) than can be cached */
```

src/ts/Implementations/Asset.ts
```diff
@@ -87,24 +82,20 @@ export class Asset extends PublishableItem {
...
+     * @remarks Currently uses the page cache
      */
     get cacheKey(): string {
         return this.#page.cacheKey;
     }
 
-    /**
-     * The options to make an asset request
-     */
+    /** The options to make an asset request*/
```
PeteCoward, line 91:
```suggestion
    /** The options to make an asset request */
```

PeteCoward (approved):

src/ts/Implementations/Manifest.ts
md8n: Done

PeteCoward (approved):

PR #742, opened 2021-06-10
Use JS toString in preference to custom method

md8n:
# Description

JS has a property called `toString` which is what gets invoked whenever an object is 'dumped' as a string.

It is considered preferable to override this if a custom string representation is wanted for a certain object.

As we're using this for a lot of logging related to cache handling etc. it's also preferable to do this minor tweak.

### DIFF
--- a/src/ts/Implementations/Page.ts
+++ b/src/ts/Implementations/Page.ts
@@ -92,7 +92,7 @@ export class Page extends PublishableItem implements StorableItem {
                 this.saveToStore(manifestData);
             })
             .catch((err) => {
-                logger.warn("%s:%s deserialize %o", this.str, this.url, err);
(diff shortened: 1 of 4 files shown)


src/ts/Implementations/Page.ts
```diff
@@ -134,7 +134,7 @@ export class Page extends PublishableItem implements StorableItem {
     }
 
     get title(): string {
-        return this.manifestData?.title || "";
+        return this.manifestData?.title || this.toString() || "";
```
PeteCoward, line 137: circular? title calls toSting calls title etc?

md8n: I didn't find that in practice. But that is likely to be because we've got manifestData.title in all cases.
I'll change it back.

PeteCoward (approved):

PR #744, opened 2021-06-10
url cleanup + pageType + manifest cacheKey

md8n:
# Description

Legacy code used `fullUrl` and `api_url` members to identify where to get resources from, and under what name to save them in the cache.

Over time `fullUrl` was mostly changed  to just `url`.  Also the structure of the `Page` related information within the `Manifest` was standardised, including the usage of `api_url`.

This :pr: :

* Cleans up the `PageType` type to provide better typing support, obviously this includes the `api_url` member.  With the consistency in the manifest this reliably works, and ensures that TypeScript can provide significantly better 'compile' time type support.
* The `PublishableItem` abstract base class is slightly refactored to make much better use of `api_url` and define a better `url`.  The refactoring was to pivot the code from `url` to `api_url`, which is a better fit for its usage in cache interactions.
* `Asset` has special handling for `api_url` so this was refactored to match.
* `Manifest` had its `fullUrl` member removed (it's not used).  Because api_url and cacheKey are interrelated the `ManifestCacheKey` was also updated at this time from `canoe-manifest` to `bero-manifest`.
* `CacheItem` had all of its references to `fullUrl` replaced with `url`.  However it should be noted that most of this code is already deprecated and will likely disappear very soon.

### DIFF
--- a/src/ts/Implementations/CacheItem.ts
+++ b/src/ts/Implementations/CacheItem.ts
@@ -168,7 +168,7 @@ export const InitialiseFromCache = async (
     item: IPublishableItem
 ): Promise<boolean> => {
     const itemCache = await AccessCache(item.cacheKey);
-    if (!itemCache || !item.api_url) {
+    if (!itemCache || !item.backendPath) {
         item.status.cacheStatus = "prepared";
         return false;
     }
--- a/src/ts/Implementations/CacheItem.ts
+++ b/src/ts/Implementations/CacheItem.ts
@@ -19,7 +19,7 @@ const AccessCache = async (cacheKey: string): Promise<Cache> => {
 };
 
 /** Build a request object we can use to fetch the item
- * The item must have a valid `fullUrl`
+ * The item must have a valid `url`
  */
 const BuildRequestObject = (item: IPublishableItem): Request => {
     const headers: any = {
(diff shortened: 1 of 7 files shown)


src/ts/Implementations/CacheItem.ts
```diff
@@ -19,7 +19,7 @@ const AccessCache = async (cacheKey: string): Promise<Cache> => {
 };
 
 /** Build a request object we can use to fetch the item
- * The item must have a valid `fullUrl`
```
PeteCoward, line 22: I'm not sure we have a use for any code in this file anymore, can you explain why we need it?

src/ts/Implementations/PublishableItem.ts
```diff
@@ -9,16 +12,21 @@ export enum UpdatePolicy {
...
 
 /** A network backed content item that can be stored and retrieved from a named cache
- * The item can be queeried for cache status and added and remove from the cache
+ * @remarks The item can be queried for cache status and added and removed from the cache
  */
 export abstract class PublishableItem {
-    /** The url to retrieve this item from */
-    abstract get url(): string;
+    /** The api_url of this item as recorded in the manifest */
+    abstract get api_url(): string;
```
PeteCoward, line 19: the assets are not retrieved from an API, hence the change to `url`

PeteCoward (changes requested):
review naming as discussed in slack

src/ts/Implementations/CacheItem.ts
md8n: As per slack - I want it to be a :pr: on its own when it comes to deleting this file.

PeteCoward (approved):

PR #745, opened 2021-06-11
Reactivate sync page

md8n:
# Description

Reactivate the Sync Page (but not the 'sync me' button)

### DIFF
--- a/src/riot/Screens/Sync.riot.html
+++ b/src/riot/Screens/Sync.riot.html
@@ -15,13 +15,13 @@
                         <h5>ID:{statusListing.id} {statusListing.type}:{statusListing.title} <small>({statusListing.version})</small> </h5>
                     </div>
 
-                    <div if="{statusListing.api_url === statusListing.cacheKey}" class="sync-body">
-                        <span><small>Store URL/Cache Key:</small> { statusListing.api_url }</span>
+                    <div if="{statusListing.backendPath === statusListing.cacheKey}" class="sync-body">
+                        <span><small>Store Path/Cache Key:</small> { statusListing.backendPath }</span>
                     </div>
-                    <div if="{statusListing.api_url != statusListing.cacheKey}" class="sync-body">
-                        <span><small>Store URL:</small> { statusListing.api_url }</span>
+                    <div if="{statusListing.backendPath != statusListing.cacheKey}" class="sync-body">
+                        <span><small>Store Path:</small> { statusListing.backendPath }</span>
                     </div>
-                    <div if="{statusListing.api_url != statusListing.cacheKey}" class="sync-body">
+                    <div if="{statusListing.backendPath != statusListing.cacheKey}" class="sync-body">
                         <span><small>Cache Key:</small> { statusListing.cacheKey }</span>
                     </div>
 
--- a/src/riot/Screens/Sync.riot.html
+++ b/src/riot/Screens/Sync.riot.html
@@ -50,14 +50,11 @@
         import { AppelflapConnect } from "ts/AppelflapConnect";
         import { CacheUtilities } from "ts/CacheUtilities";
         import { CachePublish } from "ts/CachePublish";
-        import { Manifest } from "ts/Implementations/Manifest";
 
         export default {
             state: {
                 statusListings: [],
                 publications: "",
-                manifest: "",
-                page: "",
                 appDataStatus: new AppDataStatus(),
                 appelflapConnect: null,
             },
@@ -87,7 +84,6 @@
                 this.update(({ publications: JSON.stringify(publications) }));
 
                 // Get the Item Status Listing
-                await this.state.appDataStatus.Initialise();
                 await this.state.appDataStatus.BuildList();
                 this.update({statusListings: this.state.appDataStatus.itemListings});
             },
--- a/src/riot/Screens/Sync.riot.html
+++ b/src/riot/Screens/Sync.riot.html
@@ -15,24 +15,17 @@
                         <h5>ID:{statusListing.id} {statusListing.type}:{statusListing.title} <small>({statusListing.version})</small> </h5>
                     </div>
 
-                    <div
-                        class="sync-body"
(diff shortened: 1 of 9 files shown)


src/riot/Screens/Sync.riot.html
```diff
@@ -15,24 +15,17 @@ <h3>{ TRANSLATIONS.latest() }</h3>
                         <h5>ID:{statusListing.id} {statusListing.type}:{statusListing.title} <small>({statusListing.version})</small> </h5>
                     </div>
 
-                    <div
-                        class="sync-body"
-                        if="{statusListing.storeStatus !== 'unset' || statusListing.cacheStatus !== 'unset'}"
-                    >
-                        <span>Store URL:{ statusListing.api_url }, status:{ statusListing.storeStatus }</span>
+                    <div if="{statusListing.backendPath === statusListing.cacheKey}" class="sync-body">
```
PeteCoward, line 18: why this logic? cacheKey may not equal backendPath at all. Currently it does but only becasue the manifest returns `storage_container` as the same as `api_url`. That could change

src/riot/Screens/Sync.riot.html
```diff
@@ -94,8 +87,8 @@ <h5>ID:{statusListing.id} {statusListing.type}:{statusListing.title} <small>({st
                 this.update(({ publications: JSON.stringify(publications) }));
 
                 // Get the Item Status Listing
-                // await this.state.appDataStatus.Initialise();
-                // await this.state.appDataStatus.BuildList();
+                await this.state.appDataStatus.Initialise();
```
PeteCoward, line 90: The sync tab already has access to the manifest as part of the route `getRoute().manifest`
https://github.com/catalpainternational/canoe/blob/RC-3.0/src/js/Routing.js#L71

You should pass that manifest reference into any extra class you need for this, rather than as below newing up another

src/riot/Screens/Sync.riot.html
```diff
@@ -94,8 +87,8 @@ <h5>ID:{statusListing.id} {statusListing.type}:{statusListing.title} <small>({st
...
+                await this.state.appDataStatus.BuildList();
```
PeteCoward, line 91: the manifest already contains a list of pages, why duplicate that list and not use the source directly?

src/ts/AppDataStatus.ts
```diff
@@ -51,77 +49,57 @@ export class AppDataStatus {
...
         let isPublishable = false;
 
         const pageData = getPageDataFromStore(pageId);
-        status.storeStatus = pageData ? "ready" : "unset";
 
         if (inCache) {
-            const page = new Page(this.manifest, pageId, statusId);
-            await InitialiseFromCache(page);
-            status.cacheStatus = page.status.cacheStatus;
+            const page = new Page(this.manifest, pageId);
```
PeteCoward, line 81: please use ( or enhance ) the existing Page init methods on the manifest, so that all page initialisation goes through the same path.

src/ts/AppDataStatus.ts
```diff
@@ -51,77 +49,57 @@ export class AppDataStatus {
...
-    */
-    /*
 
     async BuildList(): Promise<void> {
         this.itemListings = [];
-        this.itemListings.push(this.ManifestListing());
+        this.itemListings.push(await this.ManifestListing());
 
         const pageIds = Object.keys(this.manifest.pages);
         const cacheKeys = await CacheKeys();
```
PeteCoward, line 105: Why the CacheKeys class, what info do you need more than we already have available on the Page class?

src/riot/Screens/Sync.riot.html
md8n: It simpliifies presentation in the sync page, and that's all - considering that the sync page in this form is going away sometime very soon this work is temporary, and only for while I get the publishing/syncing working

src/riot/Screens/Sync.riot.html
md8n: yep - will do - I forgot that we had made that change when I re-enabled this code

src/riot/Screens/Sync.riot.html
md8n: It's a presentation list for this page alone - purely related to figuring out what to publish/sync

src/ts/AppDataStatus.ts
md8n: I haven't touched that code - it is as it was.
As per the JSDocs `CacheKeys()` is a utility function that gets us a list of cache keys which helps identify precisely what is in the cache already, which can be used for reconciliation with pages.
Fundamentally this code is about producing a listing of cached items to show to the user in the sync page.  Whether we continue to need this is another story.

src/ts/AppDataStatus.ts
md8n: That will be part of a different :pr: - this is simply about re-enabling the sync page.

It is also likely that this specific functionality will go away.

src/riot/Screens/Sync.riot.html
md8n: This is also about identifying what is in the cache vs. what the manifest declares, which is necessary for doing sync properly

src/riot/Screens/Sync.riot.html
md8n: OK, so there's no such thing as `getRoute().manifest` because the manifest part of that is thrown away before it hits the `store`.
Furthermore, `getValidManifest()` only returns a `TManifestData` type from the store, because that's all that we put there.  Whereas this functionality needs all of the Manifest metadata

PeteCoward: take a look at  `getRoute().page.manifest` maybe

md8n: ruit lifecycle

PeteCoward (approved):
Not entirely convinced replacing all Promise based code with awaits is necessary, if for labelling then then adding `async ` alone is fine.

PR #747, opened 2021-06-11
Reactivate 'sync me' functionality

md8n:
# Description

Turns back on the 'sync me' functionality in the sync page


src/ts/AppDataStatus.ts
```diff
@@ -1,34 +1,31 @@
...
-//     setSubscriptions,
-//     unpublishItem,
-// } from "./Implementations/ItemActions";
-
+import {
+    getSubscriptions,
+    publishItem,
+    setSubscriptions,
+    unpublishItem,
+} from "./Implementations/ItemActions";
```
PeteCoward, line 16: is this another Appelflap only class?

PeteCoward (approved):

src/ts/AppDataStatus.ts
md8n: It would be more correct to call it an Appelflap mostly class

PR #749, opened 2021-06-15
solution for card heights

megganeturner:
Following from https://github.com/catalpainternational/hamahon/issues/258
Reference PRs from Alessandro: 
https://github.com/catalpainternational/refer/pull/100/files#
https://github.com/catalpainternational/refer/pull/118/files

Today I learned about [named slots](https://riot.js.org/api/#named-slots) which made this a lot easier than all previous attempts at a solution. Enabled me to shuffle the button elements around and use the power of grid to fill the screen up.

### DIFF
--- a/src/index.hbs
+++ b/src/index.hbs
@@ -54,6 +54,7 @@
         }
 
         body {
+            height: 100%;
             margin: 0;
         }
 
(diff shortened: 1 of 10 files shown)


PeteCoward (approved):
Looks fine to me, will get up to staging

PR #31, opened 2020-04-08
What browser to use

katevoss:
Closes https://github.com/catalpainternational/asteroid/issues/24

If you try to visit Asteroid from a Chrome browser on an iPhone/iOS device, you will see the following screen:
image

And you are unable to interact with any part of the app.

PR #750, opened 2021-06-15
Reactivate the CertChain

md8n:
# Description

Reactivates the CertChain code (gets a signed package publishing certificate)

Also 'reactivates' the CanoeHost code - an underlying change in Appelflap 'broke' it, but it didn't matter until now.

Adds appropriate logging for stuff that was otherwise hidden and hard to track

### DIFF
--- a/src/sw.js
+++ b/src/sw.js
@@ -3,6 +3,7 @@ import { registerRoute } from "workbox-routing/registerRoute.mjs";
 import { setCatchHandler } from "workbox-routing/setCatchHandler.mjs";
 import { CacheFirst } from "workbox-strategies/CacheFirst.mjs";
 import { CacheAnyOrFetchOnly } from "js/CacheAnyOrFetchOnly.mjs";
+import { NetworkOnly } from "workbox-strategies/NetworkOnly.mjs";
 
 import { RangeRequestsPlugin } from "workbox-range-requests";
 import { ExpirationPlugin } from "workbox-expiration";
@@ -78,6 +79,15 @@ registerRoute(
         plugins: [new RangeRequestsPlugin()],
     })
 );
+
+registerRoute(new RegExp(ROUTES_FOR_REGISTRATION.tokenAuth), new NetworkOnly());
+registerRoute(new RegExp(ROUTES_FOR_REGISTRATION.tokenAuth), new NetworkOnly(), "POST");
+
+registerRoute(new RegExp(ROUTES_FOR_REGISTRATION.appelflapPKIsign), new NetworkOnly(), "POST");
+
+// webpack-dev-server communicates over this endpoint. Without this clause, the
+// service worker caches these requests and breaks webpack-dev-server.
+registerRoute(new RegExp(ROUTES_FOR_REGISTRATION.socketInfo), new NetworkOnly());
 
 const getNotificationTitleMessageAndTag = (eventData) => {
     let title = null;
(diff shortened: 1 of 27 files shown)


src/sw.js
```diff
@@ -79,6 +80,15 @@ registerRoute(
...
 
+registerRoute(new RegExp(ROUTES_FOR_REGISTRATION.tokenAuth), new NetworkOnly());
+registerRoute(new RegExp(ROUTES_FOR_REGISTRATION.tokenAuth), new NetworkOnly(), "POST");
+
+registerRoute(new RegExp(ROUTES_FOR_REGISTRATION.appelflapPKIsign), new NetworkOnly(), "POST");
+
+// webpack-dev-server communicates over this endpoint. Without this clause, the
+// service worker caches these requests and breaks webpack-dev-server.
+registerRoute(new RegExp(ROUTES_FOR_REGISTRATION.socketInfo), new NetworkOnly());
+
```
PeteCoward, line 91: As i [mentioned](https://github.com/catalpainternational/canoe/pull/703#pullrequestreview-679127024) NetworkOnly registrations are not required any more

PeteCoward (changes requested):
no need for the sw registrations I think.

Also we have AppelFlapConnect, CanoeHost, CertChain and StartUp modules. 
Firstly these are all only relevant in appelflap and should be named or in a folder to show that, I figure this is a quick change to make so please do so.
Secondly do we really need all 4 of them? That might take a longer refactor so I am not requesting that change, just that you think about what are the responsibilities for each of those classes.

src/sw.js
md8n: > no need for the sw registrations I think.
I'll double check on the appelFlapPKIsign one, but the others I think you may be correct.

> Also we have AppelFlapConnect, CanoeHost, CertChain and StartUp modules.
> Firstly these are all only relevant in appelflap and should be named or in a folder to show that, I figure this is a quick change to make so please do so.
Will do

> Secondly do we really need all 4 of them? That might take a longer refactor so I am not requesting that change, just that you think about what are the responsibilities for each of those classes.
They came about because of how Wicher had originally defined the Appelflap APIs and their abstractions, and therefore me following through on that.
CanoeHost does very little and should probably be merged in with StartUp.  However, the others are reasonable divisions of logic.

PeteCoward (approved):

src/ts/Appelflap/BeroHost.ts
```diff
@@ -12,19 +15,23 @@ export class CanoeHost {
         return this.#afc;
     }
 
-    /** Tell Appelflap that Canoe is 'locked'
+    /** Tell Appelflap that Bero is 'locked'
      * and should not be rebooted */
```
boerwastaken, line 19: It's an insertion lock, not a reboot lock. And even then, in certain situations Appelflap performs requested injections anyway, such as on startup before loading any page.

src/ts/Appelflap/CacheUtilities.ts
```diff
@@ -13,7 +13,7 @@ export class CacheUtilities {
         return JSON.parse(statusDescription);
     };
 
-    /** Instruct Appelflap to reboot Canoe */
+    /** Instruct Appelflap to reboot Bero */
     reboot = async (): Promise<void> => {
```
boerwastaken, line 17: I see I didn't update the API documentation to reflect that we have /soft-reboot and /hard-reboot now. The former just stops the *session* (a bit like closing and reopening a tab), the latter stops and restarts the whole *engine* (the "Firefox process" so to speak).

See the #31 for the API status overview, from where it can also be gleaned that this endpoint is not part of the minimal set of functionality that you need to be Appelflap-contentexchanging. Of course it's not forbidden to address that endpoint, but it's just not envisioned to be necessary to do that in the normal course.

The endpoint referenced in the routing does not exist anymore.

Issue #751, opened 2021-06-15
Package publishing certificate doesn't renew after user logs in


md8n:
# Description

The package publishing certificate is not signed if the user is not currently logged in (and a few other scenarios).

However, once the user has signed in the package publishing certificate is not automatically resent to the canoe-backend in an effort to renew its signing.

Working around this is possible (assuming the user has logged in) by pulling down on the Appelflap menu bar and hitting the browser refresh button, but we really don't want users doing that.

PR #752, opened 2021-06-15
Use singletons and statics to avoid `globalThis` (and flinging an AppelflapConnect object around everywhere)

md8n:
# Description

# precursor to #751 

Prior to fixing up the handling of the package publishing certificate (certChain) when the user logs in and logs out, it seemed prudent to clean up some old and rather verbose code.

Converted to Singletons (and removed as globals accessed via `globalThis`):
* AppelflapConnect
* BeroHost
* CertChain

Converted to `static` classes (all methods are static):
* CachePublish
* CacheSubscribe
* CacheUtilties

Which meant that all of the following could be simplified, especially as there is no longer any need to pass around an `AppelflapConnect` object or set up globals:
* index.js
* Sync.riot.html
* AppDataStatus
* StartUp
* ItemActions

And also the AppelflapConnect.test file was fixed up - try `yarn test` it works.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -10,9 +10,6 @@ import { EMPTY_SLATE_BOOT_KEY } from "ts/Constants";
 
 let currentServiceWorkerState = getServiceWorkerState();
 
-// Create the global beroHost and certChain objects
-var beroHost = null;
-var certChain = null;
 // Initialise the beroHost object
 InitialiseBeroHost();
 
--- a/src/app.js
+++ b/src/app.js
@@ -34,7 +34,7 @@ initialiseIdentity()  // Am I logged in
         // perform independent actions that require login in parallel
         return Promise.all([
             initialiseUserActions(), // read actiion from api and idb
-            initialiseCertChain(),   // initialise the appelflap sharing cert
+            initialiseCertChain(),   // initialise the appelflap package publishing cert
         ]);
     }).then(() => {
         initialiseRouting();         // react to the navigation hash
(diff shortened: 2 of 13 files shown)


src/index.js
```diff
@@ -10,9 +10,6 @@ import { EMPTY_SLATE_BOOT_KEY } from "ts/Constants";
 
 let currentServiceWorkerState = getServiceWorkerState();
 
-// Create the global beroHost and certChain objects
-var beroHost = null;
-var certChain = null;
 // Initialise the beroHost object
 InitialiseBeroHost();
```
PeteCoward, line 14: I'd really prefer to do this in app.js

PeteCoward (changes requested):
failed to compile
```
ERROR in src/ts/Appelflap/AppelflapConnect.ts:37:1
prettier/prettier: Delete `····`
    35 |             const gt = globalThis as Record<string, any>;
    36 |             gt["AFC_MOCKMODE"] = gt["AFC_MOCKMODE"] || false;
  > 37 |
       | ^^^^
```

src/index.js
PeteCoward: no need to change that here though

md8n:
> failed to compile
> 
> ```
> ERROR in src/ts/Appelflap/AppelflapConnect.ts:37:1
> prettier/prettier: Delete `····`
>     35 |             const gt = globalThis as Record<string, any>;
>     36 |             gt["AFC_MOCKMODE"] = gt["AFC_MOCKMODE"] || false;
>   > 37 |
>        | ^^^^
> ```

Fixed - Thanks `prettier`

src/index.js
md8n: I was thinking of a little review of this stuff as a part of fixing #751 , because it is a reasonable to do so with fixing that issue

PeteCoward (approved):

PR #755, opened 2021-06-17
Correct false assumption that assets were unique by id

### LINKED ISSUE #754: Card not showing available video rendition
Continued from [here](https://github.com/catalpainternational/canoe-backend/issues/134)

On JID staging, I'm still seeing the infamous `Sorry: This video is missing` even if the original video file is available.  This is the card I used [on this card](https://jid-staging.temp.build/#/site/learn/introduction-to-emergency-care/the-value-of-emergency-care:content:1)

I found out that the original [is present](https://admin.jid-staging.temp.build/media/media_transcodes/12-original-video.kogz5ldh.mp4) in the manifest but not being loaded on the card:

```
<videocard>
  <div>
    <p class="error-message">Sorry: This video is missing.</p>
    <h3 class="media-title">Udan Boot</h3>
    <p class="media-description">jgcvjt</p>
  </div>
</videocard>
```

And here's the media object in the manifest

```
{
    "id": "12",
    "type": "video",
    "renditions": {
        "original-video": {
            "path": "media_transcodes/12-original-video.kogz5ldh.mp4",
            "size": 5698246
        }
    }
}
```

@boerwastaken is already trying to figure out why the transcoding didn't work for this file but the original file should cover for this, no?

PeteCoward:
fixes #754 

We somehow ended up newing up Assets with `Page` and `id` params, which causes a problem when you have two assets on a page with the same id, which is very possible because media ids can clash with image ids, which is what happened with Udan Boot in the linked issue.

Resolved by creating Asset objects with the actual entry instead of an id in the constructor

### DIFF
--- a/src/ts/Implementations/Asset.ts
+++ b/src/ts/Implementations/Asset.ts
@@ -49,22 +49,20 @@ const SIZE_DIFF_PERCENT = 0.02;
 
 /** An asset ( binary resource ) than can be cached */
 export class Asset extends PublishableItem {
-    /** id of this asset in the manifest page */
-    #id: string;
     /** reference to the parent page (used to point to the correct cache) */
     #page: Page;
     /** stored reference to the Asset details */
-    #entry: TAssetEntry | undefined;
+    #entry: TAssetEntry;
 
     /**
      * Instantiate a new asset from its page and id
      * @param page the page this asset is associated with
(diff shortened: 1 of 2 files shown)


PeteCoward:
This has been reviewed on jid-staging, merging to cut `v.3.0`

PR #759, opened 2021-06-18
New appelflap access

md8n:
# Description

This :pr: makes the changes required to support Appelflap's new way for the app to get the details it needs to communicate with Appelflap.

There's more work to be done to re-enable several tests, but that requires further investigation.  Probably a good thing to do when we actually add some more tests for other parts of the app.


src/ts/Appelflap/AppelflapConnect.ts
```diff
@@ -143,6 +151,17 @@ export class AppelflapConnect {
         /* eslint-enable no-case-declarations */
     };
 
+    private getEndpointProperties = async (): Promise<void> => {
+        if (!this.#endpointProperties) {
+            const { commandPath } = APPELFLAPCOMMANDS.getEndpointProperties;
+            logger.info(`Getting endpoint properties`);
```
PeteCoward, line 157: my logger class is really going to need some disabling and filtering functionality!

md8n: It is - I've tried to be careful in my use of it.  For example being specific about `log` vs. `info` vs, `warn` etc.

md8n: But I do prefer having the telemetry rather than not having it.

It does look good when running the tests.

PeteCoward: yes, I remember wanting to be able to configure an environment with `logClasses: ["AppelflapConnect", "PublishableItem"]` and/or `logLevel: "warn"` but I got hung up writing that code and realised i was yakshaving so stopped 👼

PeteCoward (approved):

PR #765, opened 2021-06-21
621 Bero homepage and Course Categories

megganeturner:
Needs backend: https://github.com/catalpainternational/canoe-backend/pull/223

Initial implementation for the Bero Homepage


src/riot/WagtailPages/BeroHome.riot.html
```diff
@@ -0,0 +1,145 @@
...
+                    courseCategories: () => gettext("Courses by Category"),
+                    resources: () => gettext("Resources"),
+                },
+
+                get page() {
+                    return getRoute().page;
+                },
+
+                get learningActivities () {
+                    return this.page.learningActivities.topics;
```
PeteCoward, line 102: This line throws an exception if the page does not exist ( like in JID )

PeteCoward (commented):
Will need to support the home page children being possibly absent

PeteCoward:
I've hacked a frontend fix for avoiding issues on JID when there is no learning activities page, and I noticed the new buttons causing problems
Screenshot 2021-07-05 at 16 40 42

PeteCoward (approved):

PR #766, opened 2021-06-21
751 package publishing cert

### LINKED ISSUE #751: Package publishing certificate doesn't renew after user logs in

md8n:
Closes #751 

# Description

The package publishing certificate (the certificate chain) needs renewing / refreshing if a user logs in, and also needs to be deleted if a user logs out (not specified in the above issue, but still necessary).

This :pr: does the above.

`AuthenticationUtilities.js` has been turned into a quasi-state machine, and the call to `initialiseCertChain` included for the log in and log out related actions.  `initialiseCertChain` has also been removed from the more generic start up in `app.js`, because that was the wrong place for it.


src/js/AuthenticationUtilities.js
```diff
@@ -3,78 +3,141 @@ import { unsubscribeFromNotifications } from "js/Notifications";
...
-            "X-CSRFToken": Cookies.get(CSRF_COOKIE_NAME),
-        },
-    }).catch((err) => {
+        credentials: "include",
+        headers: { "X-CSRFToken": Cookies.get(CSRF_COOKIE_NAME) },
+    };
+
+    return fetch(authUrl, requestInit).catch((err) => {
         // delete the Canoe=Offline-Session cookie
         // Django is configured to log users out if no Canoe-Offline-Session is received
```
boerwastaken, line 98: Is this (supposed to be) true? @PeteCoward ?

boerwastaken:
Upon webapp login, with an unsigned (=freshly generated) cert in Appelflap, I'm not seeing any calls to the backend to sign the cert, nor do I see a signed cert ending up in Appelflap. Are you seeing one?
Of course then I don't see any bundles being produced either :crying_cat_face:

md8n:
> Upon webapp login, with an unsigned (=freshly generated) cert in Appelflap, I'm not seeing any calls to the backend to sign the cert, nor do I see a signed cert ending up in Appelflap. Are you seeing one?
> Of course then I don't see any bundles being produced either 😿

I suspect that may have been because I still had `io.catalpa.canoe.*` floating around as the test for the userAgent string (and therefore the presence of Appelflap) that's been fixed and merged into RC-3.1 - which has now been merged up into this :pr:

canoe-environment-default.js
```diff
@@ -3,13 +3,13 @@
...
     GA_TAG: false,
-    API_BASE_URL: "http://127.0.0.1:8000",
+    API_BASE_URL: "http://localhost:8005",
     SKIP_SW: false,
     DONT_SHOW_COMPLETIONS_AFTER: "2020-2-15",
     APPLICATION_SERVER_KEY: "",
     WEBPACK_CONFIG: {
         devServer: {
-            port: 8080,
+            port: 8085,
```
PeteCoward, line 12: why?

src/js/AuthenticationUtilities.js
PeteCoward: yes, it may not be in the canoe-engineering branch though. Are you saying it is untrue?

src/ts/Appelflap/CertChain.ts
```diff
@@ -50,6 +52,26 @@ export class CertChain {
...
+        }
+
+        logger.info(
+            `Initialisation completed, package publishing certificate is ${this.certState}`
+        );
+
+        return this.certState === "signed";
+    }
+
+    private async CreateCertChain(): Promise<void> {
```
PeteCoward, line 74: please name ClassesLikeThis and methodsLikeThis

src/ts/Appelflap/CertChain.ts
```diff
@@ -50,6 +52,26 @@ export class CertChain {
...
+            await this.DeletePackageCertificateFromAppelflap();
+            this.#packageCert = undefined;
+            logger.info(
+                `User not authenticated, package publishing certificate is ${this.certState}`
+            );
+        }
+
+        logger.info(
+            `Initialisation completed, package publishing certificate is ${this.certState}`
+        );
```
PeteCoward, line 69: double log here

PeteCoward (commented):
we already have redux state, why add separate faux state?

PeteCoward (commented):
please make singleton pattern consistent with https://github.com/catalpainternational/canoe/blob/5aeeb7028ca08e8d7bb6c784bd4b7584ac83f490/src/ts/Implementations/Manifest.ts#L44

src/ts/Appelflap/CertChain.ts
```diff
@@ -110,18 +135,31 @@ export class CertChain {
...
+                let isCertSigned = false;
+                if (!certHeader) {
+                    // This is a simplistic examination of the cert to establish that it is signed
+                    // Appelflap checks the certificate chain explicitly
+                    const certsFound =
+                        decodedCertificate.match(/-END CERTIFICATE-/g) || [];
+                    isCertSigned = certsFound.length === 3;
+                } else {
+                    isCertSigned = certHeader === "3";
+                }
```
PeteCoward, line 157: do we need this if? Does `appelflapPKIsign` sometimes return this header and sometimes not?

boerwastaken: it never returns such a header. It does give the webapp a cert in a 2xx response, which the webapp is not supposed to try to understand, instead shoving it into Appelflap verbatim. And Appelflap will verify.

src/js/AuthenticationUtilities.js
boerwastaken: What I've observed is that while I do need that `Canoe-Offline-Session` cookie for my session to be accepted, if I leave it out, it doesn't log me out, as I can add that cookie in at a later point and get that same session ID accepted. In other words, it doesn't log me out.
So either the middleware implementation is incorrect or the above codecomment is incorrect, because the codecomment states Django would log me out.

boerwastaken: What I've observed last week is that while I do need that `Canoe-Offline-Session` cookie for my session to be accepted, if I leave it out, it doesn't log me out, as I can add that cookie in at a later point and get that same session ID accepted. In other words, it doesn't log me out.
So either the middleware implementation is incorrect or the above codecomment is incorrect.

PeteCoward: yes, the middleware does not actually log you out just prevent access. For the Bero frontend purposes it is effective as a logout though.
I'll open an issue to address this, not sure how big a priority it is though. Maybe we can just update the comment ;)

canoe-environment-default.js
md8n: Because this is the default development environment for the `catalpa_canoe` project

src/ts/Appelflap/CertChain.ts
md8n: Done

src/ts/Appelflap/CertChain.ts
md8n: Done

md8n:
> we already have redux state, why add separate faux state?

And you are talking about what?

Remember that redux is holding state only, but we have need for several stateful objects

md8n:
> please make singleton pattern consistent with
> 
> https://github.com/catalpainternational/canoe/blob/5aeeb7028ca08e8d7bb6c784bd4b7584ac83f490/src/ts/Implementations/Manifest.ts#L44

done

PeteCoward:
> Remember that redux is holding state only, but we have need for several stateful objects

If you have app state anywhere, please put it in redux anything that at any point may have relevance app wide. UI state can go in riot state.
Here I guess it's not really state, just events and event reactions.

src/js/AuthenticationUtilities.js
boerwastaken: See also https://catalpa.slack.com/archives/C0122BDJ7FA/p1624033875047000

PR #768, opened 2021-06-22
Because someone changed the API

md8n:
and now there's a `bundleType` for publish and subscribe


boerwastaken (approved):
Whoopsydaisy! There might be more of this coming, the cache/sw distinction will pop up in a number of places. I'll ping you!

PR #771, opened 2021-06-25
683 formative assessment

megganeturner:
issue: #683 
design: https://projects.invisionapp.com/d/main/#/console/21052158/452766162/preview
backend: https://github.com/catalpainternational/canoe-backend/pull/220

Front end implementation for formative assessments

Still todo:
- store the data


megganeturner:
@PeteCoward I made an attempt at adding in the logic to store data, using the exam score logic as a guide, but could be totally off! All those changes are in [this commit](https://github.com/catalpainternational/canoe/pull/771/commits/c17d74a3ba000c0fb0598f1bfc6fe7f582024548) if it's easier to exclude this for now (:

PeteCoward (commented):
Is there any need to store this data in redux I wonder? There is no way to go back and see it in the interface yet

src/riot/Teaching/FormativeAssessment.riot.html
```diff
@@ -0,0 +1,251 @@
...
+                },
+                decreaseStudents() {
+                    if (this.state.totalStudents > 1) {
+                        this.update({ totalStudents: this.state.totalStudents - 1 })
+                    }
+                },
+
+                decrementQuestionTally() {
+                    var answerA = this.currentQuestion.answers[0];
+                    var answerB = this.currentQuestion.answers[1];
```
PeteCoward, line 218: I think we're going to have to support more that two answers here ( the backend does )
Tallies should be stored for correct and incorrect, not associated with the answers

PeteCoward (approved):

PR #773, opened 2021-06-25
Update bundle type defs

md8n:
# Description

Adds better typing for the 'bundle' data returned by Appelflap from a `GET` `subscriptions` call


src/ts/Implementations/ItemActions.ts
```diff
@@ -125,12 +126,14 @@ export async function setSubscriptions(
...
     items.forEach((item) => {
         subscriptions.origins[self.origin].caches[item.cacheKey] = {
-            injection_version_min: item.version,
-            injection_version_max: item.version,
-            p2p_version_min: item.version,
-            p2p_version_max: item.version,
+            injection_version_min: item.version - aDay,
+            injection_version_max: item.version + aDay,
+            p2p_version_min: item.version - aDay,
+            p2p_version_max: item.version + aDay,
```
boerwastaken, line 136: This looks like policy, not mechanism. Is it? Policy should be configurable.

PR #774, opened 2021-06-25
Partial fix, but usable-ish

md8n:
# Description

This matches the changes to the subscription API as described in the docs.

A separate :pr: will cover actually ensuring it works with Appelflap

### DIFF
--- a/src/ts/AppDataStatus.ts
+++ b/src/ts/AppDataStatus.ts
@@ -215,7 +215,7 @@ export class AppDataStatus {
     async GetSubscriptions(): Promise<TSubscriptions> {
         const subscriptions = await getSubscriptions();
         if (typeof subscriptions === "string") {
-            return { origins: {} };
+            return { types: { CACHE: { groups: {} } } };
         }
         return subscriptions;
     }
@@ -224,7 +224,7 @@ export class AppDataStatus {
     async SetSubscriptions(): Promise<TSubscriptions> {
         const subscriptions = await setSubscriptions(this.itemListings);
         if (typeof subscriptions === "string") {
-            return { origins: {} };
+            return { types: { CACHE: { groups: {} } } };
         }
         return subscriptions;
     }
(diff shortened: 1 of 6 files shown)


src/ts/AppDataStatus.ts
```diff
@@ -242,10 +242,12 @@ export class AppDataStatus {
         const published = await this.PublishAll();
         const unpublished = await this.UnpublishAll();
         let subscriptions = await this.GetSubscriptions();
-        let origins = Object.keys(subscriptions.origins);
+        let origins = Object.keys(subscriptions.types.CACHE.groups);
```
boerwastaken, line 245: There may not always be a `CACHE` key. Is that caught somewhere?
Note that any `map` type in [the relevant descriptor classes](https://github.com/catalpainternational/appelflap/blob/master/koekoek/src/main/java/org/nontrivialpursuit/ingeblikt/Descriptors.kt#L74:L92) may be empty.

src/ts/tests/AppelflapConnect.test.ts
```diff
@@ -367,33 +367,37 @@ test.skip("Cache: setSubscriptions", async (t: any) => {
...
-                        p2p_version_max: 888,
-                        injected_version: 12,
-                    },
-                    "another-cache-name": {
-                        injection_version_min: 42,
-                        injection_version_max: 42,
-                        p2p_version_min: 1,
-                        p2p_version_max: 9000,
-                        injected_version: null,
+        types: {
```
boerwastaken, line 370: I'd recommend also throwing a SWORK type in here. And to test the empty map too.

boerwastaken (changes requested):
Note that any `map` type in [the relevant descriptor classes](https://github.com/catalpainternational/appelflap/blob/master/koekoek/src/main/java/org/nontrivialpursuit/ingeblikt/Descriptors.kt#L74:L92) may be empty.

Are you sure that's reflected in your types and tests?

PeteCoward:
I may regret asking, but what is [SWORK](https://www.urbandictionary.com/define.php?term=swork)?

md8n:
> I may regret asking, but what is [SWORK](https://www.urbandictionary.com/define.php?term=swork)?

I'm guessing an abbreviation for Service Worker

src/ts/tests/AppelflapConnect.test.ts
md8n: done

src/ts/AppDataStatus.ts
md8n: done

md8n:
> Are you sure that's reflected in your types and tests?

It is now

boerwastaken:
> >  I may regret asking, but what is SWORK?

> I'm guessing an abbreviation for Service Worker

Youse _could_ grep the docs, ya know.

PR #782, opened 2021-07-01
Api changes sync monitoring

md8n:
# Description

Adds support for the `injectables`, `launchWiFiPicker` and `launchStorageManager` API endpoints.
`injectables` is also wired up to support checking if there's any relevance to 'unlocking' the cache to let Appelflap know it can proceed with synchronisation.


src/ts/Appelflap/AppelflapConnect.ts
```diff
@@ -318,4 +318,17 @@ export class AppelflapConnect {
...
+    //#region Appelflap Debug
+    /**
+     * Inject all relevant bundles into the browser's cache
+     * @remarks When this call returns it indicates conclusively that cache syncing has completed
+     */
+    public injectAll = async (): Promise<string> => {
+        const { commandPath, method } = APPELFLAPCOMMANDS.setInjectAll;
+
+        logger.info(`Injecting all relevant bundles into the browser's cache`);
+        return await this.performCommand(commandPath, { method }, "text");
```
boerwastaken, line 331: When this gets moved to the official API, wouldn't you prefer if it would spew JSON?

src/ts/Appelflap/CacheUtilities.ts
```diff
@@ -56,4 +56,14 @@ export class CacheUtilities {
             await AppelflapConnect.getInstance()!.unlock();
         }
     }
+
+    /**
+     * Inject all relevant bundles into the browser's cache
+     * @remarks When this call returns it indicates conclusively that cache syncing has completed
```
boerwastaken, line 62: Actually, the API will actually return a report of successes and failures on all the things it has tried to do...

It may be better to reflect the actual outcome in the UI.
Summary-style "successfully injected 6 caches and 1 serviceworker" or more detailed per-item status.

src/riot/Screens/Sync.riot.html
```diff
@@ -248,9 +248,9 @@ <h5><small>{statusListing.type}:</small> { statusListing.title }</h5>
              */
             performInboundSync() {
                 this.update({syncStatus: "syncOn"});
-                // CacheUtilities.injectAll().then(() => {
+                CacheUtilities.injectAll().then(() => {
```
boerwastaken, line 251: Depending on what you want from `injectAll()`, this could be considered racy.
As soon as you release the injection lock (supposing you're holding it), Appelflap will go ahead and inject stuff as it sees fit. So it might have completed one or more staged injections before you hit the injectAll endpoint. The injectAll endpoint might then return that nothing was injected at all, or fewer things than you expected were injected, because there was nothing left to do for it at that point. However, you will be able to see those injections reflected in the subscriptions, of course. You could diff before and after and show the frontend what went down. But that might be a bit unwieldy.

One way out of this is to keep the lock, but hit the injectAll endpoint while passing a `honor_injectionlock=false` query parameter. Then you'll get an accurate report of what was injected.

boerwastaken (changes requested):
Currently it makes everyone believe that things are great while they might not be.
It would be better to actually evaluate the outcomes delivered through the injectAll API endpoint.

src/ts/Appelflap/AppelflapConnect.ts
md8n: Here, yes.  But otherwise it's only really used as a flag to say "job done".

md8n: Ahh, good point, it's really a list of failures and successes.

src/ts/Appelflap/CacheUtilities.ts
md8n: True, to start with it's simply 'job done' as far as the UI is concerned. But later I think nuance will be required.

src/riot/Screens/Sync.riot.html
md8n: I noted that from your previous comments.

I had also thought about doing a before and after examination of the subscriptions, which would be the next step that I'd take - another :pr:, and probably another release I think.

boerwastaken: A complicating factor is that serviceworker injections do not take effect until after a full browser engine reboot. So there's the question of how to represent this outstanding work — it's not "injected" yet, rather its "queued for injection" until after a reboot.
Perhaps we should keep the serviceworkers out of the Sync screen injection seances then. There's nothing you can meaningfully *do* with them in the Sync screen. You could look at the queued SW injections though, and tell the user "we also have an app update from July 1st 2021 15:31 that will be applied in the future" or something along those lines (this stuff is uncrystallized as of yet).

boerwastaken: Would it be useful to have an endpoint that shows you what Appelflap is planning to do, should you call injectAll now? I mean, the webapp could figure it out by querying `/appelflap/ingeblikt/publications` + `/appelflap/ingeblikt/subscriptions` and matching those things, but that logic already exists in Appelflap anyway, so it could easily just tell the webapp what it thinks needs doing.

boerwastaken: See #88 for this endpoint in the flesh. You can query the planned injections now at `/appelflap/ingeblikt/subscriptions/injectables`.

src/ts/AppDataStatus.ts
```diff
@@ -200,15 +201,30 @@ export class AppDataStatus {
...
     }
 
     /** Unpublish everything currently not flagged as isPublishable */
     async UnpublishAll(): Promise<TPublishResult> {
-        return this.PerformAll(
-            (listing) => !listing.isPublishable,
-            unpublishItem
-        );
+        const certChain = CertChain.getInstance();
+        if (certChain && certChain.certState === "signed") {
```
boerwastaken, line 219: why is "unpublishing" tied to the cert status here?

Anyone can and should publish the bundles they've received so far (in fact, that's what happens by default), so that they can be trickle-spread by peers in turn. Your cert status doesn't matter one bit for that.

You might want to stop spreading a certain bundle at a certain point in time - if it's from 8 years ago and no peers will be reasonably looking for it. Then you'd update the bundle's `p2p_version_min` through the subscriptions, at which point Appelflap will step in and garbage-collect the bundle if it doesn't match anything in the subscriptions anymore. It's declarative configuration.

To be clear: You don't need a certificate to spread a bundle. Bundles are always signed — by the creator of the bundle. There is no TLS. P2P peers don't care at all whether you have a certificate. You can serve and keep serving any bundle (created by yourself in the past, or created by others) regardless of your current cert status.

Deleting a bundle by issueing a straight HTTP DELETE is more something of an emergency thing you can do if you've inadvertently published some cache.

In the normal flow of operations, your only point of contact for expressing which things should be spread and which shouldn't be, is the declarative Subscriptions configuration. Appelflap takes care of garbage collecting anything that's not referenced by any matching subscription. This keeps things simple on the webapp side. 
When it comes to bundle management, the webapp needs to express its intentions through the Subscriptions, and in addition to that it can create some bundles if it wants to (yet note that if it creates one that doesn't match any subscription, then it'll be GC'ed pretty soon). So, the subscriptions are leading.

boerwastaken: Also note that while you can DELETE a cache, if your subscriptions say you're still interested in it, you might find it'll re-appear soon after because Appelflap duly re-acquires it from some peer. Again: subscriptions are leading.

src/ts/Appelflap/AppelflapConnect.ts
```diff
@@ -295,6 +295,20 @@ export class AppelflapConnect {
             "text"
         )) as Promise<TSubscriptions>;
     };
+
+    /**
+     * Get a list of all bundles that are 'injectable' into the cache in response to Subscriptions
+     * @remarks this corresponds with @see getSubscriptions which are the bundles that have been subscribed to
+     */
+    public injectables = async (): Promise<TBundles> => {
```
boerwastaken, line 303: Probably further on you might want to filter out SWORK bundles, or handle/represent them differently. They can't be injected while the webapp is running, they're difficult to explain to the user, and will get auto-injected when the app boots anyway. But you can tell the user that an app update is waiting and will be reapplied when the app starts anew. `window.close()` is hooked into app exit, so you could even offer the user to quit the app this way (they'll then have to re-launch it by themselves though)

src/ts/Appelflap/AppelflapRouting.ts
```diff
@@ -84,6 +95,8 @@ export const APPELFLAPCOMMANDS = {
         commandPath: `${AF_CACHE_API}/${AF_PUBLICATIONS}`,
         method: "DELETE",
```
boerwastaken, line 96: Emergency handbrake! As explained above, use the Subscriptions and let Appelflap do the garbage collection.

src/riot/Screens/Sync.riot.html
```diff
@@ -136,15 +131,14 @@ <h5><small>{statusListing.type}:</small> { statusListing.title }</h5>
...
 
         import TopMenu from "riot/Components/TopMenu.riot.html";
 
         export default {
             state: {
                 statusListings: [],
-                publications: "",
                 appDataStatus: new AppDataStatus(),
                 // mocked data below
                 deviceId: '2198734892374',
```
boerwastaken, line 144: A device ID is in the range of 55^3 - 55^4.
2198734892374 is way outside that range.
Also it's a number, not a string.
`12345657` would be a realistic peer ID.

src/ts/AppDataStatus.ts
```diff
@@ -200,15 +201,30 @@ export class AppDataStatus {
...
+            return this.PerformAll(
+                (listing) => !listing.isPublishable,
+                unpublishItem
+            );
+        }
+        // This user does not have authority to publish anything,
```
boerwastaken, line 225: Any user, any app instance, can and will spread on its collection of bundles (the ones you can see the index of at `/appelflap/ingeblikt/publications`.

src/ts/AppDataStatus.ts
md8n: Thanks for that info on the signing vs. sharing of bundles.  I'll make the relevant changes

src/ts/Appelflap/AppelflapConnect.ts
md8n: I was wondering

src/ts/Appelflap/AppelflapRouting.ts
md8n: I added the definition here because it was in the documentation - it's not in use.

Taking your warning into account though I'll remove/block this definition

src/riot/Screens/Sync.riot.html
md8n: That's not my code - It came from Meggan, and as it says it is mocked data, but I'll make a change to it

src/ts/AppDataStatus.ts
md8n: As above, I'll make the change for that.

src/ts/Appelflap/AppelflapRouting.ts
md8n: Handled via a separate :pr:

src/ts/Appelflap/AppelflapConnect.ts
md8n: This will be handled in a separate :pr: later

src/ts/AppDataStatus.ts
md8n: Done

src/ts/Appelflap/CacheUtilities.ts
md8n: To be handled in a separate :pr:

src/ts/Appelflap/AppelflapConnect.ts
md8n: outdated

PR #783, opened 2021-07-02
Get peer Id from Appelflap

md8n:
# Description

Via the mozilla extension used by Appelflap we now have access to a Peer ID for a user.

This can be wired into the Sync page as that user's 'peer' ID

The format returned looks like this:
``` json
{
  "ID": 3657874,
  "friendly_ID": "qZdP",
  "palette": "23456789abcdefghjkmnpqrstuvwxyzABCDEFGHJKLMNPQRSTUVWXYZ"
}
```
For which this :pr: provides a TypeScript type of `TPeerProperties` and a get property accessor on the `AppelflapUtilities` module called `peerProperties`.

This :pr: also does some refactoring now that we have some additional Appelflap specific functions, that have nothing to do with the caching. So that we have a more representative structure of actual functionality.

### DIFF
--- a/src/ts/Appelflap/AppelflapConnect.ts
+++ b/src/ts/Appelflap/AppelflapConnect.ts
@@ -15,6 +15,7 @@ import { AF_CERTCHAIN_LENGTH_HEADER, AF_LOCALHOSTURI, APPELFLAPCOMMANDS } from "
 
 import Logger from "../Logger";
 import { inAppelflap } from "../PlatformDetection";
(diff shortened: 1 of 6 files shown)


megganeturner (approved):

boerwastaken:
Sidenote: the palette tells you how the ID number has been converted into the friendly ID. Had it been "0123456789" then the friendly ID would have come out exactly the same as the ID (decimal).

So it's informational.

PR #787, opened 2021-07-02
Restore the ability to env-set a default language.

sandroroux:


src/js/Translation.js
```diff
@@ -18,12 +22,19 @@ export const SUPPORTED_LANG_CODES = process.env.LANGUAGES;
 
 export function initialiseLanguage() {
     let language = localStorage.getItem(LANGUAGE_STORAGE_KEY);
+    const envLanguage = process.env.CANOE_DEFAULT_LANGUAGE;
+
+    // Allow an env-set default language.
+    if (SUPPORTED_LANG_CODES.includes(envLanguage)) {
+        localStorage.setItem(LANGUAGE_STORAGE_KEY, envLanguage);
+        language = envLanguage;
+    }
```
sandroroux, line 31: I added this back in.

PeteCoward: no need, check the latest and you can define a project LANGUAGES array, the default is the first in that list

PeteCoward:
I think we can close this @sandroroux will try to catch you online to explain

PR #794, opened 2021-07-06
Feature - sync reenabling

md8n:
# Description

This feature branch re-enables the original sync page and sync code, and then moves everything to its new location within Bero under the Settings

It also includes many changes to the Appelflap API which support a variety of new UI elements within the Sync page, such as showing storage and wifi info, and listing peers (who are on the same wifi network)


src/ts/Appelflap/SyncStatus.ts
```diff
@@ -0,0 +1,984 @@
...
+ *     - starts polling WiFi connectivity
+ *     - starts polling Storage space
+ *     - checks Injectable bundles count
+ * - Exiting (On -> Off) - exiting the Sync State, stops all polling activity
+ * - WiFi Off->On - starts polling Peers, in the background Appelflap will try sharing files between Peers
+ * - WiFi On->Off - stops polling Peers, clears list of Peers, stops slow polling Injectable bundles count
+ * - Injectable bundles 0 -> +ve - calls for Cache Injection to commence, starts a quick polling loop:
+ *     - comparing the original Items to Sync snapshot with its updated state
+ *     - re-checks Injectables bundles count
+ * - Injectable bundles +ve -> 0 - stops quick polling Injectable bundles count, updates Items to Sync
```
PeteCoward, line 90: not sure what `bundles 0 -> +ve` and `bundles +ve -> 0 ` means here

src/ts/Appelflap/SyncStatus.ts
```diff
@@ -0,0 +1,984 @@
...
+            const pageId = pageIds[ix];
+            const manifestPage = Manifest.getInstance().pages[pageId];
+            const inCache = cacheKeys.includes(manifestPage.storage_container);
+            const pageListing = await this.PageListing.bind(this)(
+                pageId,
+                manifestPage,
+                inCache
+            );
+            pageListings.push(pageListing);
+        }
```
PeteCoward, line 833: the only code that reads from the cache to determine if a page is available offline should be in the Page.ts methods. Why not just use those methods, editing or adding to them as required?

PeteCoward (changes requested):
please rename SyncStatus.eTag to `subscriptionsEtag` if as I think is the case it is an etag just for the subscriptions endpoint

src/ts/Appelflap/SyncStatus.ts
```diff
@@ -0,0 +1,984 @@
...
+
+        const pageIds = Object.keys(Manifest.getInstance().pages);
+        const cacheKeys = await CacheKeys();
+
+        const pageListings: TItemListing[] = [];
+        for (let ix = 0; ix < pageIds.length; ix++) {
+            const pageId = pageIds[ix];
+            const manifestPage = Manifest.getInstance().pages[pageId];
+            const inCache = cacheKeys.includes(manifestPage.storage_container);
+            const pageListing = await this.PageListing.bind(this)(
```
PeteCoward, line 827: might be slighly more performant to do these in parallel with Promise.all(), not important for now though

src/ts/Appelflap/SyncStatus.ts
```diff
@@ -0,0 +1,984 @@
...
+            )
+        );
+        itemListings.push(
+            ...pageListings.filter(
+                (pageListing) =>
+                    !pageListing.isPublishable &&
+                    !pageListing.isAvailableOffline &&
+                    !pageListing.isValid
+            )
+        );
```
PeteCoward, line 860: kind of inefficient way to sort an array, not important for now though

src/ts/Appelflap/SyncStatus.ts
```diff
@@ -0,0 +1,984 @@
...
+        manifestPage: TWagtailPage,
+        inCache: boolean
+    ): Promise<TItemListing> {
+        const statusId = manifestPage.storage_container;
+        let isValid = false;
+        let isAvailableOffline = false;
+        let isPublishable = false;
+
+        if (inCache) {
+            const page = new Page(Manifest.getInstance(), pageId);
```
PeteCoward, line 796: Why not use Manifests getSpecificPage here?

src/ts/Appelflap/SyncStatus.ts
```diff
@@ -0,0 +1,984 @@
...
+            // await InitialiseFromCache(page);
+            isValid = page.isValid;
+            isAvailableOffline = await page.isAvailableOffline();
+            isPublishable = await page.isPublishable();
```
PeteCoward, line 800: All we care about it isPublishable. No need to do two awaits in series here.

PeteCoward (approved):

Issue #604, opened 2021-04-19
Devices are able to sync the latest content


pedro16v:
## Overview
Due to the low connectivity context, we want to enable devices to share their latest content with other nearby devices.

This will leverage the back-end work already prepared by @boerwastaken. 

## Conditions of satisfaction
- [ ] Devices use WiFi to sync content.

pedro16v:
Latest design from @andrewwapling https://invis.io/MD113620ZJS6

andrewwapling:
Hi @pedro16v thanks for setting this up. Look forward to reviewing with the team at a suitable time so that we may gather team feedback and refine and iterate the design direction if required.

pedro16v:
Hi @andrewwapling @PeteCoward @Juliana-t @KaraChesal , this is a summary of today's chat with @boerwastaken . I found the feedback to be a bit too dense to add as individual comments on InVision. I'm happy to follow up in there as well if required.

We can break this out further on a call. Let me know.

## Notes:
* There is no need of having a "sync" button.
  * Peer syncing is the ultimate goal and never depends on internet connection. There is no reason to turn it off that we can think of.
  * As for the existing content update mechanism, we have no control over it. It always happens as we use Bero, caching courses and lessons we visit. The only thing we have control over is explicit Lesson download from the Course listing page.
* The Wi-Fi network is useful
* The "People you can sync with" is a bit of a stretch at this point. We have no way of knowing exactly who's next to us. We can only tell that some device id is present on the network.
* At this point, it's probably best to just display a confirmation message saying "There are X available peers you can sync with". This and the Wi-Fi component should be enough to reinsure that everything is ok to sync.
* The "Sync status" component is ok. There's some complexity to how things sync exactly but users don't need to know that. We can use the states seen on your prototype
* Managing storage can get tricky:
  * We can know how much space is available on the device
  * We can't clear anything on the device that wasn't created by Bero which limits the options.
* I can imagine "The recent activity" component being a long list of "Sync success" since we have no way of knowing who we synced from.

Regarding the two tabs "General + Status", it might be useful to be able to see Wi-Fi status, Peer status, and Sync status all in the same place so that it's easier to "debug" for users.

boerwastaken:
> The "People you can sync with" is a bit of a stretch at this point. We have no way of knowing exactly who's next to us. We can only tell that some device id is present on the network

[...]

PR #806, opened 2021-07-07
Syncing FE tweaks

megganeturner:
Towards #604 

Some small tweaks:
- remove the list of 'statusListings'
- show the last updated date in the sync status (defined by the manifest version)
- update to the status icon according to `syncStatus` - loading icon if loading, tick if done, cross if error
- strip back the peer avatars for now - just display `friendly_id` and map to colours (some discussion about that [here](https://github.com/catalpainternational/canoe/issues/604) but awaiting further direction)


md8n (changes requested):
Immediately under `appelflapStatus() {` get rid of the first three lines as they will be surplus to requirements.

md8n (approved):

PR #812, opened 2021-07-08
Adds Tetun Translations and IGNORE_BROWSER_LANGUAGE env Setting

sandroroux:


canoe-project-default.js
```diff
@@ -23,5 +23,4 @@ module.exports = {
     REQUIRE_LOGIN: false,
     PROFILE_LINK: "profile",
     LANGUAGES: ["en", "fr", "tet"],
```
PeteCoward, line 25: surely you should add `IGNORE_BROWSER_LANGUAGE: false` here?

PeteCoward: and adding `LANGUAGES: ["en"]` which I should have done would avoid the need for the check in Translation.js maybe

sandroroux: I have `IGNORE_BROWSER_LANGUAGE` in canoe-environment. How should I decide if a setting belongs in -project or -environment?

PeteCoward (approved):
oh, they were there already. Then the code to guard against it in Translations is mostly unnecessary, approved anyway

PR #833, opened 2021-07-14
Sync (as a) state (machine)

md8n:
# Description

For the Sync page to be able to function in a reasonable way it needs to be able to respond to lots of 'environmental' changes.
* WiFi off/on
* List of Peers changing
* List of 'Injectable bundles' changing (the items that Appelflap has in storage ready for 'injecting' into Bero's cache)
* Storage space changing

And from these changes it needs to be able to respond appropriately, transitioning from state to state, and performing different actions required by those state transitions.

Appelflap can't (currently) push notification of changes into Bero, so Bero's SyncState has to poll for them at a reasonable frequency.
But Bero's SyncState will also need to respond to the above changes, which means that it will need to know when to turn polling off and on for different changes to its environment.  For example, polling for changes to peers is pointless if there's no Wi-Fi connectivity.

SyncState removes all of the code from the Sync page that was actually involved in the sync behaviour, and has turned it into a state machine (with the code that was in `AppDataStatus`) that stores any changes to its state in the redux store.  The Sync page simply turns SyncState `On` when mounting and `Off` when unmounting, and listens for changes to `syncState` in the redux store.


megganeturner (approved):

boerwastaken:
> Appelflap can't (currently) push notification of changes into Bero, so Bero's SyncState has to poll for them at a reasonable frequency.

Options:

1. long poll on HTTP API, it'll return a body at the moment something changes (potentially specifying in what area the change occured), at which point the webapp knows it may be worth polling some or all endpoints of interest. Could be a websocket too.
2. WebExtension bidirectional channel

The downside of option 2 is that it'd be a new interface requiring setup to debug. The HTTP API one can just `curl` which is quite luxurious.

Both option 1 and 2 would require quite a bit of rewiring in Appelflap to make the events that are happening in various subsystems tick over into the completion of the long-poll response, or into sending an event over the webextension channel.

PR #837, opened 2021-07-14
Better poll handling

md8n:
# Description

Ensures that all previous pollerIds are wiped on `On` - they "should be", but "should be" and "are" are two different terms.

Provides better feedback when getting the `peers` info fails.


PeteCoward (approved):
It's kind of hard for me to dive in and review having lost a bit of context in this.
Having said that I will approve while grinding my teath over the capitalized method attibutes and raising an eyebrow at the 1000 line file ;)

md8n:
> raising an eyebrow at the 1000 line file ;)

Well if it was python, you would say that it was too short

PR #838, opened 2021-07-14
Sizes superman a bit better.

sandroroux:


PeteCoward:
A bit concerned this might edit they way things look on other deployments in an unpredictable way. I might have to check before and after on PeP, and ask @megganeturner to do the same on JID

megganeturner:
This doesn't quite work for JID!

Before:


After:

sandroroux:
Oops! Thanks, @megganeturner for checking that (that's an excellent graphic too). I'll try again. @PeteCoward, can I add extra scss in canoe-overrides? The styling changes can be project specific

PeteCoward:
> can I add extra scss in canoe-overrides? The styling changes can be project specific

You can as long as that file is imported somewhere from core Bero.
Ideally such overrides are as small as possible, so that updates to Core Bero filter through to platforms reliably. If you override the whole of one big scss file any further changes to that file in core Bero are lost to you without care and attention.

The easiest solution here may be to just resize the image asset to be consistent with other projects ( it may be the other projects which are inconsistent ). I know designers never deliver us consistent sizes and ratios, but we can try to get them to live within the restrictions it really helps.

sandroroux:
@PeteCoward Noted. I'll work with Andy to get it tweaked

sandroroux:
@PeteCoward @megganeturner I got a dimension-matching replacement for the original. No need for new Sass. I'll close this. Thank you both for the review!

PR #841, opened 2021-07-15
Reset all of the pollers on reentry

md8n:
# Description

On exit (unMounted) from the Sync page it should call the `Off` method on `SyncStatus`, which in turn should shut down any pollers still in operation.

However, if it doesn't, we need to ensure that they're all off whenever we re-enter the Sync page so that when it calls the `On` method for `SyncStatus`, it will ensure that they're all reset (in effect ensuring that its state is actually `Off` first) and then turn itself `On`, transitioning from `Off` to `Initialising` correctly.

This :pr: does that ensuring that we have a clean exit from `SyncStatus` whenever possible, and also a clean start every time.


PeteCoward (approved):
`On` is a confusing method name in the context of javascript

PR #843, opened 2021-07-15
Ensure that the sync page gets updated first time

md8n:
# Description

The `Sync` page is 'refreshed' by listening to changes to the redux store for `syncState` and then comparing the previous and next versions of things.  It tends to not matter if the `Sync` page misses one or two of these change events, but it does when the sync process is first commencing.

The `SyncState` module can quickly push out several updates when it first starts, so that by the time the `Sync` page gets notified it can't clearly identify differences between the previous and next versions of some pieces of data.

The `Sync` page needs to know when it needs to just update everything regardless of whether previous and next states are the same.  This :pr: does that adding a `firstUpdate` state to the `Sync` page so that it knows to set up everything at least once.


src/riot/Screens/Sync.riot.html
```diff
@@ -224,7 +226,23 @@ <h5 style="color:{statusListing.isPublishable ? 'green' : 'red'}; margin: 0.2em
             storeListener(previousStoreState, storeState) {
                 const prevSyncState = previousStoreState.syncState || {};
                 const nextSyncState = storeState.syncState || {};
-                if (JSON.stringify(prevSyncState) !== JSON.stringify(nextSyncState)) {
+                if (this.state.firstUpdate) {
+                    // Set everything
+                    const updates = {
+                        syncStatus: nextSyncState.syncState,
+                        ssid: nextSyncState.ssid,
+                        peerId: nextSyncState.peerId || { id: 0, friendly_id: "unknown" },
```
boerwastaken, line 234: I'd give `friendly_id` a value of the correct length (4). `"0000"` would be fine, it's actually a value that would not occur with the base55 palette currently in use. It'd look suspicious but would not mess up the UI.

src/riot/Screens/Sync.riot.html
```diff
@@ -210,6 +211,7 @@ <h5 style="color:{statusListing.isPublishable ? 'green' : 'red'}; margin: 0.2em
 
             async onMounted() {
                 // Fire up the SyncStatus
+                this.state.firstUpdate = true;
```
boerwastaken, line 214: isn't that already set just so b.c.o [here](https://github.com/catalpainternational/canoe/pull/843/files#diff-2237848bb2b6ec5a01a618966987095a88aa66e5e0f90c71bf18cb0ba6e5f9aaR128)?

src/riot/Screens/Sync.riot.html
```diff
@@ -224,7 +226,23 @@ <h5 style="color:{statusListing.isPublishable ? 'green' : 'red'}; margin: 0.2em
...
+                        usedSpacePercentage: nextSyncState.usedStoragePercentage,
+                        bundles: nextSyncState.bundles,
+                        firstUpdate: false,
+                    };
+                    if (this.state.isNotProduction) {
+                        updates["statusListings"] = nextSyncState.itemListing || [];
+                    }
+                    
+                    this.update(updates);
+                } else if (JSON.stringify(prevSyncState) !== JSON.stringify(nextSyncState)) {
```
boerwastaken, line 245: is this comparison robust wrt ordering? is the ordering of fields within the compared state objects stable? Because
```js
JSON.stringify({"a": 1, "b":2}) == JSON.stringify({"b": 2, "a":1})
>> false
```

Perhaps you can forgo this approach for a simple event counter in the thing that generates these state snapshot frames.

boerwastaken:
I must admit I'm not that familiar with what is being built/extended here, so I'm unfamiliar with the problems of this architecture, thus I can't judge any solutions to those problems right now :-/

src/riot/Screens/Sync.riot.html
md8n: In general the ordering will be consistent, or more correctly, it is reliable enough for the purposes required.

If the fields do ever end up with different ordering then it doesn't matter for this process, as that would simply be a false positive.  The only thing we want to avoid is false negatives where we are not presenting a valid change of state to the user.

src/riot/Screens/Sync.riot.html
md8n: I prefer to properly default values at their creation even if I'm going to explicitly set them to the same value later..  It ensures that the value is of the same type consistently, rather than `undefined` | `type`.

megganeturner (approved):

PR #844, opened 2021-07-15
prettier sync peer avatars

boerwastaken:
I thought a linebreak would allow for increased font size and thus increased legibility without claiming more space.

Before:

cramped_peervatar

After:

prettier_peervatar

The colors differ since I threw out the use of that hash function — as the peer ID is already a random number from a uniform distribution, grabbing its friendly string representation and hashing that to get a number in a hopefully-also-uniform distribution (depends on the quality of that hash function...) looks like doing too much work to me ;-)

Sidenote: it turns out that modern browsers let you access quality cryptographic hash functions, but they're a bit cumbersome to use, mainly because you're driven into async-land. This'd be a way to reduce a string to an integer using those APIs (but it's not a drop-in for `hashString()`, because async):
```js
crypto.subtle.digest("SHA-1",new TextEncoder().encode("heh")).then(buf => new DataView(buf).getUint32()).then(console.log)
>> 1118692072
```


megganeturner (approved):

megganeturner:
> Sidenote: it turns out that modern browsers let you access quality cryptographic hash functions, but they're a bit cumbersome to use, mainly because you're driven into async-land. This'd be a way to reduce a string to an integer using those APIs (but it's not a drop-in for `hashString()`, because async):
> 
> ```js
> crypto.subtle.digest("SHA-1",new TextEncoder().encode("heh")).then(buf => new DataView(buf).getUint32()).then(console.log)
> >> 1118692072
> ```

I have never encountered this before, nifty!

PR #845, opened 2021-07-16
Better subs info, constrain polling

md8n:
# Description

Based on a conversation with Wicher, and observed behaviour, it seems important to constrain how hard Jetty (Appelflap's web server) is hit, so that it has sufficient time to respond to each request.

In particular, when we tell Appelflap to do cache injection (because `injectables` count > 0), we want to be able to fast poll the count of injectables as it goes back down to 0, so we know when to transition out of the `Syncing` state, and also to provide some better feedback to the user about the progress of actual Syncing (hopefully).

The above means that we also should turn off all other polling temporarily.

And just to ensure that we know we've subscribed to something, which is the trigger that drives Appelflap to share bundles between phones and to have them ready to inject into the cache, there's also an update to the `SetSubscriptions` function to log more accurate info of the actual number of items subscribed to.


src/ts/Appelflap/SyncStatus.ts
```diff
@@ -187,9 +187,9 @@ export class SyncStatus {
     }
 
     set slowPoll(value: number) {
-        if (this._fastPoll > 1200) {
-            this._fastPoll = 1200;
-        }
+        this.HardConstrainSlowPoll();
```
PeteCoward, line 190: I'm having difficulty working out where the `slowPoll` and `fastPoll` setters might be called from.
I see it is possible they are called through `UpdateSyncStatusData` but to actually see where and when that might happen is challenging me

md8n: slowPoll is used everywhere for everything, fastPoll is only used when cache injection has been called (Syncing).

I set them up this way so that it is possible for their values to be set by the Sync page itself.

PeteCoward: > slowPoll is used

I can see where the values are used. I can;t see where the setters are used.

>  it is possible for their values to be set

We should not commit code that is not actually exercised in any way

md8n: > We should not commit code that is not actually exercised in any way
Normally yes.  However, in this case, because there's almost certainly a need to tweak and adjust these values, and determining where and how best to do that is still to be decided, then I strongly believe in taking an engineers approach and facilitating options to change these via a UI, settings values, etc. is prudent.

i.e. YAGRI - You ain't gonna regret it.

boerwastaken:
> Based on a conversation with Wicher, and observed behaviour, it seems important to constrain how hard Jetty (Appelflap's web server) is hit, so that it has sufficient time to respond to each request.

I don't think I said that, it's not quite so simple.

The Appelflap webserver is quite performant, it's just that operations surrounding injection and subscriptions are serialized (it's shared state, we don't want multiple edits at the same time!), so if you fire off many requests, Appelflap will serve those with many threads, but just one of those threads will be doing something or perhaps none of them will be doing anything while Appelflap itself is busy with the subscriptions.

So if you spawn and stack requests in rapid succession willy nilly you might up with a lot of idle threads on the Appelflap side and a lot of stalled requests on the webapp side.

Even if there are no serverside limits, with fetch & promises, it's a good idea to put a limit on the amount of requests in flight at any one time, because there are also browserside limits on how many you can run - it's an exhaustible resource. But maybe you'll hit Jetty's thread limit first.

Anyway, it's a good idea to guarantee you're not creating a thundering herd of `Fetch()`-es.

src/ts/Appelflap/AppelflapConnect.ts
```diff
@@ -327,27 +329,36 @@ export class AppelflapConnect {
...
+                "response"
+            )) as Response;
+
+            getSubscriptions = {
+                eTag: subResponse.headers.get(AF_ETAG) || "",
+                subscriptions: (await subResponse.json()) as TSubscriptions,
+            };
+
+            // Strip all double quotes from the eTag
+            getSubscriptions.eTag = getSubscriptions.eTag.replace(/"/g, "");
```
boerwastaken, line 346: But don't you need to tack them again later? The etag value in the `If-Match` request header is double-quoted.

src/ts/Appelflap/SyncStatus.ts
```diff
@@ -507,24 +519,39 @@ export class SyncStatus {
...
+                `Finished injection of bundles into the cache ${
+                    result ? "successfully" : "unsuccessfully"
+                }`
+            );
+            // If we're still here, we're going to call it a success
+            result = true;
         } catch (e) {
             // The injection failed
             result = false;
+            logger.warn("Failed to inject bundles into the cache");
```
boerwastaken, line 544: at least dump the actual error to the log?

src/ts/Appelflap/SyncStatus.ts
```diff
@@ -968,17 +998,32 @@ export class SyncStatus {
...
-    async SetSubscriptions(): Promise<TTaggedSubscriptions> {
-        const itemListings = await this.ItemListings.bind(this)();
-        logger.info(`Setting Subscriptions with ETag:${this.eTag}`);
-        const taggedSubs = await setSubscriptions(this.eTag, itemListings);
+    private async SetSubscriptions(): Promise<TTaggedSubscriptions> {
+        if (!this.eTag) {
+            logger.warn(
+                "Subscriptions ETag has not yet been retrieved, cannot set the subscriptions without it"
+            );
+            return AF_EMPTY_TAGGED_SUBSCRIPTIONS;
```
boerwastaken, line 1006: Am I reading here that if you don't have an etag because you didn't fetch it yet, you assume the serverside subscription state is empty?

I left some questions on your #811 regarding this too.

https://github.com/catalpainternational/canoe/pull/811#pullrequestreview-702947736

PeteCoward:
> Anyway, it's a good idea to guarantee you're not creating a thundering herd of `Fetch()`-es.

further to a conversation with @md8n today...
@boerwastaken is it possible to just have one enpoint that would return for us all of
- peers
- wifi status
- pending injectables
-? storage(although seperating this out to a completedly standalone component was my first recomendation)

that would be super simple to control concurrency from our side as it's just one call?

An alternative discussed was to unify it Bero side using a Promise.all() and make sure that is single flight at any one time. Giving jetty 3/4 concurrent requests at one time maximum.

Both solutions would remove this mulitple polling at different rates at the same time situation. I'm not sure a super fast responsive refresh rate is required here. I'd almost rather slow it down and control it for ease of communicating it.

boerwastaken:
> @boerwastaken is it possible to just have one enpoint that would return for us all of

Not really. The things you want to bundle there come from very different places and have very different performance & concurrency characteristics. I could bundle the peers & wifi status & storage as those are quick & read-only, but then again, since they're indeed quick, having them on separate endpoints is not a problem.

I realize that when one has the reduxy thing in place the temptation is to keep an accurate world state in there just in case some component is interested in a certain state transition, but changing the HTTP API to deliver the whole world on one's doorstep in one bunch is putting the cart before the horse ;-)

I'd rather we understand the problem better first.

For a start, a simple test:

Benchmarking the three info-endpoints in parallel (`ab -v 1 -H "Authorization: Basic YXBwZWw6YXBwZWw=" -t 30 http://127.0.0.1:9090/appelflap/info/storage-info` etc) yields all three running with over 200 requests/second, so 600 requests/second in total. Upping the concurrency to 30 connections one time overall (`-c 10`, but we run three in parallel) actually increases throughput to 800 requests/second.
Clearly not a problem there. You could refresh the info from those endpoints at higher-than-screen-refresh-rates and Appelflap still wouldn't break a sweat.

I'd like to see evidence of the actual problem but I have trouble getting Bero even to call the injectables endpoint...

src/ts/Appelflap/AppelflapConnect.ts
md8n: That's done when the If-Match header gets rebuilt

src/ts/Appelflap/SyncStatus.ts
md8n: This specific error I have never seen, but good point ...

src/ts/Appelflap/SyncStatus.ts
md8n: I tried fetching it, but the fetch failed (no response)

src/ts/Appelflap/SyncStatus.ts
```diff
@@ -178,44 +173,66 @@ export class SyncStatus {
...
     }
 
+    /** Set hard absolute limits for the Slow Polling delay value */
+    private HardConstrainSlowPoll(): void {
+        if (this._slowPoll < 6000) {
+            this._slowPoll = 6000;
+        } else if (this._slowPoll > 30000) {
+            this._slowPoll = 30000;
+        }
+    }
```
boerwastaken, line 220: These magic numbers would look better as constants. Are these bounds on polling intervals? Is that what you mean by "delay"? Then how are you preventing requests (delayed or not) from stacking up?

Dynamic feedback systems can work, if your feedback has high enough frequency and state changes are not too abrupt. Locking on a semaphore with only 1 token is a very abrupt change in timing characteristics, basically you go from a couple of millis of latency to infinite latency for the subsequent request that's serialized behind the former.

Really we should try to understand what the exact problem is that you're trying to solve. Which endpoints are slow for you, and do you even know how many requests you have in-flight when you run into trouble?

boerwastaken:
In the meantime, while I haven't been able to reproduce your issue so far (the webapp never ever polls /subscriptions/injectables for me :crying_cat_face:  ) I'll go and walk through everything to try to find potential deadlocks.

In the same meantime, please take `/subscriptions/injectables` out of your polling, and wire it up to a simple "check" button in the UI that polls that endpoint exactly once, and only when you click it. Just for the moment!

md8n:
All the useful bits have been extracted and used.  So what's left is no longer necessary

PR #847, opened 2021-07-16
Teaching activity scroll

megganeturner:
~Automatically scrolls users to the end of the page when returning from a formative assessment~

Latest:

Closes https://github.com/catalpainternational/pep/issues/32 & https://github.com/catalpainternational/pep/issues/34

- when a user returns to the teaching activity page from formative assessment, they are scrolled back to the formative assessment card
- when a user returns to the teaching activity page from viewing tips, they are scrolled back to the tip CTA card
- when a user clicks on 'Ready to Teach' or 'Extend Your Knowledge' at the bottom of the plan & teach tabs, they are scrolled to the tab section, rather than the top of the page
- if the user is on the tab that they click (for instance, click 'plan' while plan is the active tab), they are scrolled to the tab section

This behaviour was kind of fiddly to execute for a few reasons:
- Our routing pattern prohibits the use of linking to element ids (e.g. clicking on an `<a href="#tips">` appending `#tips` to the url and scrolling to that element).
   - Additionally, when the url/riot hash changes (when switching between tabs), the page scrolls to the top automatically (which mightn't happen when updating queryParams), so we have to listen for a `hashchange` event to finish to then execute any scrolling logic
- In the case of tips, there's no easy way to determine the context of the previous route to determine which element to scroll to (hence appending the `:scrollToTips` to the url when returning from tips)

Nothing we can't work around, but worth noting. These issues also help explain why there's a noticeable 'flash' when we scroll after a url change (this doesn't occur when returning from formative assessment as the url doesn't change for that, all that logic is just in state).

https://user-images.githubusercontent.com/12974326/127967739-3fd8b429-a28d-4378-a488-01ce6044aec9.mov


megganeturner:
Actually - just spotted this issue with some additional scroll-y requests. Will roll these into this PR
https://github.com/catalpainternational/pep/issues/32

PeteCoward:
Hmm, well we could look into changing the routing to not use hash, but then we'd have to
1) make sure nginx and devserver serve all non-existant files as index.html ( I think we might have done this already )
2) update the Routing.js file to parse the url slightly differently ( and all the riot tags that have internal routing )
3) make sure we know how to navigate internally inside Bero without a full navigation request ( using hash makes this easy, we'd have to edit the behaviour of most anchor tags to achieve it )
4) update the backend view live / preview urls to use the new routing scheme

It's not impossible, just not simple

megganeturner:
@PeteCoward definitely no need to change anything! Just felt a need to explain why this code seems a little convoluted 😅

PeteCoward:
@megganeturner this should go into ( and have merged into it ) RC-3.4

PeteCoward (approved):

PR #853, opened 2021-07-20
Improve organisation & comprehension of SyncStatus

md8n:
# Description

Based on feedback from Peter this :pr: reorganises the SyncStatus module to put several methods that are more important in functional terms higher up in the code, and also add additional comments to explain several aspects that were fairly opaque.

Apart from the renaming of several methods that were private (internal) to other methods (to make it clearer what they do) there is absolutely no code changes to this :pr:


PeteCoward (approved):
Ok, just cosmetic ( I hope )
My main feedback remains that what is important is...
1. reading the current state and telling appelflap about it ( RefreshPubSub )
2. poling for info from appelflap and displaying it
3. if appelflap tells us we have injectibles then goto 1. ( maybe after a delay to communicate and let appelflap do its thing )

md8n:
> Ok, just cosmetic ( I hope )
> My main feedback remains that what is important is...
> 
> 1. reading the current state and telling appelflap about it ( RefreshPubSub )
> 2. poling for info from appelflap and displaying it
> 3. if appelflap tells us we have injectibles then goto 1. ( maybe after a delay to communicate and let appelflap do its thing )

OK - I'll take all those comments on board.  But I'll commit this change first.

PR #859, opened 2021-07-20
Refresh subscriptions correctly

md8n:
# Description

This :pr: changes the behaviour when setting subscriptions.  So that only the manifest and bundles for outstanding / outdated pages are subscribed to, and then only if that's different from any current subscription.  Also this is only done after the list of items (from the manifest) and the current subscriptions (from Appelflap) have been retrieved.

It also includes a new package `fast-equals` to do the comparison between the existing subscriptions and the existing list of items (as the proposed subscriptions), and replaces any use of the JSON.stringify approach to doing deep equality comparisons.


src/ts/Implementations/ItemActions.ts
```diff
@@ -85,40 +85,53 @@ export async function getSubscriptions(): Promise<
...
     const taggedSubs = AF_EMPTY_TAGGED_SUBSCRIPTIONS;
     taggedSubs.eTag = eTag;
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     taggedSubs.subscriptions.types.CACHE!.groups[self.origin] = { names: {} };
 
     const aDay = 24 * 60 * 60 * 1000;
+    const aYear = aDay * 365;
 
     items.forEach((item) => {
+        const versionMin = item.version - aDay;
```
PeteCoward, line 101: Why are we subscribing to versions a day old?

src/ts/Implementations/ItemActions.ts
```diff
@@ -85,40 +85,53 @@ export async function getSubscriptions(): Promise<
...
+        const versionMax =
+            item.version + (item.type === "manifest" ? aYear : aDay);
```
PeteCoward, line 103: Why are we setting maximum version interested in to the version I have plus a day?

src/ts/Implementations/ItemActions.ts
```diff
@@ -85,40 +85,53 @@ export async function getSubscriptions(): Promise<
...
+): TTaggedSubscriptions {
     const taggedSubs = AF_EMPTY_TAGGED_SUBSCRIPTIONS;
     taggedSubs.eTag = eTag;
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     taggedSubs.subscriptions.types.CACHE!.groups[self.origin] = { names: {} };
 
     const aDay = 24 * 60 * 60 * 1000;
+    const aYear = aDay * 365;
 
     items.forEach((item) => {
```
PeteCoward, line 100: We should actually not be subscribing to updates to items that are up to date. Any page update will come with a manifest update, we do not want the page update before the manifest update.

PeteCoward: > We should actually not be subscribing to updates to items that are up to date

except for the manifest :)

md8n: you saw that ...

src/ts/Implementations/ItemActions.ts
md8n: According to https://github.com/catalpainternational/appelflap/blob/master/docs/API/update-subscription.md
The versionMin and Max, as well as the p2pVersionMin and Max accept values that define an 'inclusive' range.  However, when I tried this it never worked, it seemed to be that they were actually 'exclusive' values, hence the 1 day bracketing.

Furthermore, we don't necessarily have any actual version of this page we're subscribing to.  The version number, used in these calculations, is coming from the manifest, not the page.

src/ts/Implementations/ItemActions.ts
md8n: According to https://github.com/catalpainternational/appelflap/blob/master/docs/API/update-subscription.md
The versionMin and Max, as well as the p2pVersionMin and Max accept values that define an 'inclusive' range.  However, when I tried this it never worked, it seemed to be that they were actually 'exclusive' values, hence the 1 day bracketing.

src/ts/Implementations/ItemActions.ts
boerwastaken: > We should actually not be subscribing to updates to items that are up to date

Actually it'd be wise to p2p-subscribe (≠ injection-subscribe!) to "unknown updates from the future". That is, set p2p_version_max to, I dunno, MAXINT or, with the webapp's unix-time versioning scheme, current_version +1 year or something.
They won't be injected if they don't fall into the injection_version_min…injection_version_max range. Nothing bad will happen. Good things will happen though: as the p2p network relies on peers (surprise!), what's not immediately useful for *this peer* may well be immediately useful for *some other peer*; the p2p network works much better (in terms of throughput, resilience, and uptodateness) if bundles are spread altruistically-proactively.

Gating everything behind a manifest bump needlessly serializes work (of acquiring bundles) that could have been done in parallel and ahead-of-time. By the time the peer has realized (based on its updated manifest) "gosh yes now I'm absolutely convinced that I really finally want those updated pages, too" it might be too late to grab them (peers come and go) or a dogpile may form on the ONE peer that actually has the bundles that every other peer now suddenly wants. Doesn't that make sense?

src/ts/Implementations/ItemActions.ts
boerwastaken: I'll double-check this

boerwastaken: [This is the beef](https://github.com/catalpainternational/appelflap/blob/13ab30d716de5e5f11bb93b1ca5cbc2b9f1d63b4/koekoek/src/main/java/org/nontrivialpursuit/ingeblikt/BundleMatching.kt#L5:L16) of the bundle classification.

Note that Kotlin ranges are inclusive on both ends, so `1 in 1..2 && 2 in 1..2` is `true`.

From that snippet it follows that a bundle is injectable when:
- if the `injected_version` is `null` (nothing's been injected yet, or someone wiped the state): the candidate bundle version falls in the injection range
- if there is a non-null `injected_version`: the candidate bundle version is `gte (max(injected_version, injection_version_min))` AND greater than the `injected_version`.

A bundle is p2p-acquirable/redistributable when:
- it is injectable (duh)
- in addition, if its version fals within the p2p_version range.

This should work and if not it's a bug ;-)
Try to strip magic corrections from your puppeteering of this logic. Perhaps when you put this in there were confounding problems that led you to think the ranges exclude the ends, or perhaps you weren't aware of the logic around `injected_version`.

src/ts/Implementations/ItemActions.ts
```diff
@@ -85,40 +85,53 @@ export async function getSubscriptions(): Promise<
...
         ] = {
-            injection_version_min: item.version - aDay,
-            injection_version_max: item.version + aDay,
-            p2p_version_min: item.version - aDay,
-            p2p_version_max: item.version + aDay,
+            injection_version_min: versionMin,
+            injection_version_max: versionMax,
+            p2p_version_min: versionMin,
+            p2p_version_max: versionMax,
             injected_version: item.isPublishable ? item.version : null,
```
boerwastaken, line 112: Just to confirm: You're setting the `injected_version` to the version of which you *know* that **the webapp itself** has just installed into a cache? IOW, you're reflecting a webapp-side-performed cache update in the registry so that Appelflap won't go and inject it again (or a lower version, deity forbid) if it stumbles upon something? That would be the correct behaviour. Anything else would probably not be, except perhaps corner cases such as "I hosed my cache and would like to have it injected again" etc.

boerwastaken: note that this is shared bookkeeping. You should only overwrite values set by Appelflap (as it does, when it injects a cache) after much consideration ;-)

PR #861, opened 2021-07-21
Ahh riot, well anyway

md8n:
# Description

This is not really Riot's fault (although it's always fun to blame it).

The `Transition` method in SyncStatus is calling and handling various state transition methods that return a promise.  However, it itself is not meant to return anything.

So when Riot calls `Off` when its unmounting the page it gets the return from the `Off` method very quickly, and promptly exits the page, leaving the `TransitionToOff` still about to do its thing, which is to stop all of the polling.  As a consequence stopping the polling doesn't actually happen.

~This :pr: just calls the `StopAllPolling` method directly within the `Off` method, which ensures that the job is done, even though `TransitionToOff` would still do this later if it had the chance.~

This :pr: gets BeroHost to call `getInstance` on SyncStatus, which should keep it alive (and in the correct scope) all the way through.


PeteCoward:
It might be better to keep the SyncStatus ref around on something not related to riot tags ( eventually we may want to be publishing stuff after makeAvailableOffline succeeds )
Is one of the BeroHost componenets always in memory?

md8n:
> It might be better to keep the SyncStatus ref around on something not related to riot tags ( eventually we may want to be publishing stuff after makeAvailableOffline succeeds )
> Is one of the BeroHost componenets always in memory?

SyncStatus is already a singleton, and nominally independent of the Sync Page.  However, I did notice this behaviour in my testing yesterday.  And this strategy does work to alleviate the problem.

PeteCoward:
I'm assuming that it does not stop polling when unmounted because the only reference held to it is in the Riot Sync tag, when that object is unmounted it is garbage collected, so the SyncStatus reference is lost, so the SyncStatus object has no reference so it is garbage collected. Storing a reference in an object at app level would prevent this happening, and be preferable to this solution.

md8n:
> I'm assuming that it does not stop polling when unmounted because the only reference held to it is in the Riot Sync tag

This "shouldn't" be true, once an instance of SyncStatus has been created it should remain until the App is actually closed.  However, the behaviour that we're seeing implies that it is scoped to Riot.

PeteCoward (approved):

PR #865, opened 2021-07-21
Sync state always poll injectables

md8n:
# Description

This is a simpler-ish approach to Bero getting information related to its Sync State.

In this :pr: it always polls the Injectables endpoint as this is the key piece of information to establish is some new/updated item(s) has(have) been received from a peer and is(are) ready for injection.

Other polling is still performed, but that is now largely informational.

No WiFi and no peers could still be used to turn off polling injectables (assuming the injectables have been polled at least once), but there is still an issue about getting peer information on some versions of Android.  So until that's resolved, always polling injecgables would be preferable.


boerwastaken:
> No WiFi and no peers could still be used to turn off polling injectables [...] So until that's resolved, always polling injecgables would be preferable.

Always polling injectables is always preferable ;-)

Bundles can appear out of thin air, or at least, the webapp shouldn't concern itself deeply with how they ended up on the device. Currently there are two other ways besides the p2p-net that this can happen, more might follow.

Also, as I think you state implicitly, indeed the presence or absence of peers/wifi NOW does not have any bearing on bundles collected earlier.

Though I understand the temptation of using the (informal) state machine to model the workings of the whole world, it's hardly necessary for the webapp to do so and things become less fragile if it would abstain from doing so. Iff Appelflap says there are injectables, then there are, iff it says there aren't, then there aren't — the webapp shouldn't surmise.

md8n:
> Iff Appelflap says there are injectables, then there are, iff it says there aren't, then there aren't — the webapp shouldn't surmise.

Actually Appelflap says that there aren't any injectables, and then magically a whole heap of items get injected into the cache anyway without any action from Bero (writing this up as a new issue now)

boerwastaken:
> Actually Appelflap says that there aren't any injectables, and then magically a whole heap of items get injected into the cache anyway without any action from Bero 

It might have already injected those, and then by the time you queried for injectables there weren't any left because they'd all been injected :clap: :partying_face: 

Appelflap determines what it should inject and what you see as injectable by the exact same logic, the first thing I'd look for is thus whether the injection lock is honoured. It looks like the webapp is setting it correctly immediately when it boots (and on many an occasion after that), never relinquishing it (which is fine). So then the problem would be that Appelflap doesn't honor it. The only places where it's not honoured are:
- injection through the bundle opener - deliberately ignores injection lock
- injection via http call to `/appelflap/do/inject-caches` - deliberately ignores it (it's the webapp's will, after all)

Then there's the `FRONTEND_USES_INJECTIONLOCK` build variable that you might have had set to `false`.

src/ts/Appelflap/SyncStatus.ts
```diff
@@ -78,28 +78,25 @@ const logger = new Logger("SyncStatus");
  * The App's Sync state, as a state machine
  *
  * Sync State depends on (in order of importance):
- * 1. WiFi connectivity - this can go on or off at any time
- * 2. Injectable bundles - whether Appelflap has anything new or updated in its local store that can be injected into the cache
+ * 1. Injectable bundles - whether Appelflap has anything new or updated in its local store that can be injected into the cache
+ * 2. WiFi connectivity - this can go on or off at any time
  * 3. Peers - whether Appelflap has anyone to share things with
```
boerwastaken, line 83: or "share from"

boerwastaken:
It still `PUT`s the things it has already `PUT` before. I don't think it should - it should query `/appelflap/ingeblikt/publications` to establish whether anything needs packing up, so it can skip packing up things that are already packed up.
Of course we all know PUT is supposed to be idempotent, and here it is too, but at the same time it's a bunch of needless churn and does not scale well with the projected dozens and dozens of caches as used by the Bero webapp.

md8n:
> Then there's the `FRONTEND_USES_INJECTIONLOCK` build variable that you might have had set to `false`.

I shall check this out - thanks for the reminder

src/ts/Appelflap/SyncStatus.ts
md8n: Changed (hopefully correctly)

PR #867, opened 2021-07-22
Fetch with timeout

md8n:
# Description

Please note that this is an experimental branch, and should not be merged arbitrarily.


md8n:
Going to resolve the problem this :pr: was targeting in a different way

PR #869, opened 2021-07-22
866 sync sharing versions

### LINKED ISSUE #866: Sync - Set Manifest sharing versions appropriately
# Description

The manifest is the key document for Bero's operation.

For its synchronisation it needs the minimum version numbers set just right.  This is assuming that Bero already has a manifest.

- `p2p_version_min` (part of identifying the version being shared with peers) should be set to the manifest's current version.
- `injection_version_min` (part of identifying the range of version values being subscribed to) should be set to the manifest's current version + 1.

md8n:
Closes #866 

# Description

This sets the 'min' values used in defining subscriptions to what should hopefully be correct values.


src/ts/Implementations/ItemActions.ts
```diff
@@ -98,17 +98,25 @@ export function buildSubscriptions(
...
-        const versionMax =
+        // If our version is publishable then we want any higher version,
+        // otherwise we'll take something up to a day prior to the version the manifest told us about
+        const injVersionMin = item.isPublishable
+            ? item.version + 1
+            : item.version - aDay;
+        // For manifests we're interested in anything more than one year greater than the version we have
+        // otherwise up to tomorrow's version is fine as well
+        const injVersionMax =
             item.version + (item.type === "manifest" ? aYear : aDay);
```
boerwastaken, line 109: I'm not sure - would you really want to have a version of a page injected that your current manifest might not be aware of (since from its perspective, it's "from the future")? **Gathering** and **redistributing** it (through `p2p_version_*` seems like a good proactive behaviour, but having it injected willy nilly?

I don't think we want behaviour that the webapp sans Appelflap would not exhibit. In webapp sans Appelflap land, last time I looked at it, the following invariant holds: the version of the manifest is never lower than the versions of any of the things it refers to.

boerwastaken:
I think I might need to overhaul this subscriptions API, to make it more fine-grained and to make it easier to handle "co-authoring" it.

src/ts/Implementations/ItemActions.ts
PeteCoward: We definitely do not want a version of anything newer than the latest manifest tell us it should be.

PeteCoward: Ideally the manifest tells Bero we need Page Y version X, Bero tells Appelflap it needs Page Y Version X, Appelflap only delivers Page Y version X.
Bero when it has manifest version Z should be telling Appelflap that it needs Manifest version >Z though.

boerwastaken: For the injections, yes.
For the p2p bundle acquisition/redistribution, it'd be very wise to enter Page Y Version X+N, with N some reasonable number, for reasons outlined elsewhere.

md8n: Thanks for restating the comments here.  I originally used the version (date) ranges for everything simply to get something shared, because they weren't being shared otherwise.

md8n: Now that we can see sharing happening consistently (and other bad things that were occasionally happening have stopped happening) I think it's a really good time to go back and revisit these numbers exactly as per your comments above.

PeteCoward (commented):
I really do not understand the use of a day and a year here.

What makes sense in my head is...

For the manifest item:
Subscribe for my version + 1 , no maximum
For a page cache:
Subscribe to the version my current manifest tells me to and only that

PeteCoward (changes requested):
Please change the use of `aYear` to either a very large number, Or if the api supports it null for no maximum

src/ts/Implementations/ItemActions.ts
```diff
@@ -95,20 +95,27 @@ export function buildSubscriptions(
...
+        // This will mean, with the exception of the manifest, that we won't actually get anything (by design).
+        // Otherwise we'll take what the manifest has told us is the version we want.
+        const injVersionMin = item.version + (item.isPublishable ? 1000 : 0);
+        // For manifests we're interested in anything up to today's version (adding a day ensures no UTC or fingerprinting issues)
+        // otherwise we want what the manifest has told us is the version we want
+        const injVersionMax =
+            item.type === "manifest" ? Date.now() + aDay : item.version;
+        // For sharing with others we're saying that this item is good for a day more than its version stamp
+        // This is something we should review in the future once we have better metrics
+        const p2pVersionMax = item.version + aDay;
```
PeteCoward, line 110: I think this line ( but not the comment ) says that no versions of this item's cache greater than a day newer than the known item version will be looked for or shared. I think this is wrong.
It means that appelflap will not go looking for any version of the manifest newer than a day after the last version it has.

PeteCoward (changes requested):

PeteCoward (approved):

PR #872, opened 2021-07-23
Poll everything

### LINKED ISSUE #604: Devices are able to sync the latest content

md8n:
Closes #604 (_functionally only, UI changes should go into a separate issue_)

# Description

Frankly the Bero App has got no control over its 'environment' including its own 'list of items' (Manifest and Pages).
So the answer is to poll   E V E R Y T H I N G , which is what this :pr: does.

SyncStatus polls the:
* item listings (in effect determining what the manifest says we should have vs. what's actually in the cache) and the 
* item subscriptions (what we've told Appelflap that we want/need)

together.

If either of these change, then the entire list of items published and subscribed to gets updated with Appelflap.

This was key because Appelflap can effectively update the item listings without any instruction from Bero (or the user).

To facilitate this the `UpdateSyncStatusData` was upgraded to return a boolean indicating whether anything had changed or not in the data being polled.  That returned value was also used in the other polling methods to reduce the amount of times they would do additional (but unnecessary) processing when nothing had changed in what they were polling. 

So that the user can really see changes to these values some simple (but maybe not simple enough) metrics were added to the sync page, along with a change to the presentation of the 'Peers' so that they take up less vertical space.

_The UI changes are temporary only.  Any changes to them require input from design and then a proper issue raised._


boerwastaken:
I haven't looked at this yet but would you like a deterministically derived ETag for those endpoints to simplify the "but has it changed" part of this equation for you?

PeteCoward (approved):

PR #876, opened 2021-07-23
Gratuitous Grabbelton

boerwastaken:
Just three things I bumped into in `AppelflapRouting.ts`

Explanation/motivation in the inline comments.


src/ts/Appelflap/AppelflapRouting.ts
```diff
@@ -1,13 +1,12 @@
-/** If Appelflap is running, it is always on localhost
- * @remarks this MUST be 'localhost' not '127.0.0.1',
- * because geckoview translates '127.0.0.1' to 'localhost' in the background
- * which results in the route registration failing
+/** If Appelflap is running, it will be available on the loopback interface
+ * on a dynamically chosen port for which we have to follow a discovery
+ * procedure.
  */
-export const AF_LOCALHOSTURI = "http://localhost";
+export const AF_LOCALHOSTURI = "http://127.0.0.1";
```
boerwastaken, line 5: For me it works fine and as expected with `127.0.0.1`. I see no evidence of Geckoview "translating" "in the background" (I'm not sure what that means, I'd need to see a tcpdump for verification).

src/ts/Appelflap/AppelflapRouting.ts
```diff
@@ -1,13 +1,12 @@
...
 
 export const AF_API_PREFIX = "appelflap";
 
-export const AF_ENDPOINT = `moz-extension://${AF_API_PREFIX}`;
+export const AF_ENDPOINT = `moz-extension://appelflap`;
```
boerwastaken, line 9: The `AF_API_PREFIX` only coincidentally holds the same value as the extension name. They're independent and can change independently. We wouldn't want the extension URL to break just because we change the API URL.

src/ts/Appelflap/AppelflapRouting.ts
```diff
@@ -19,7 +18,7 @@ export const AF_INFO_API = `${AF_API_PREFIX}/info`;
 
 export const AF_DEBUG_API = `${AF_CACHE_API}/jeffreystube`;
 
-export const AF_INS_LOCK = "insertion-lock";
+export const AF_INS_LOCK = "injection-lock";
```
boerwastaken, line 21: This one is actually important! Changed in catalpainternational/appelflap#114

src/ts/Appelflap/AppelflapRouting.ts
md8n: I'll try it out (via a commit of this :pr:) If I still have problems I'll dig out the diagnostic info

md8n (approved):

PR #881, opened 2021-07-27
Refactor out sync itemListing and itemPublishing

md8n:
# Description

As part of making `SyncStatus` more digestible, this :pr: splits out the functions associated with getting item listing information (from the Manifest), and also those related to publishing items (again with reference to the Manifest), and puts each of these into separate modules with the functions inside available as `static` members.

A beneficial aspect of this is that we're not flinging `this` around (via `bind(this)`) for these functions.  Also references to the `Manifest` and `Page` classes have been removed from the `SyncStatus` module where they didn't really belong (at least not directly).

This _seems_ to have made the Syncing behaviour a little more stable, although further testing is required (as usual).


src/ts/Appelflap/SyncItemListings.ts
```diff
@@ -0,0 +1,107 @@
...
+import { Manifest } from "../Implementations/Manifest";
+import { Page } from "../Implementations/Page";
+
+/**
+ * The Item Listings, derived from the manifest, prepared for use by SyncStatus
+ */
+export class SyncItemListings {
+    private static async ManifestListing(): Promise<TItemListing> {
+        const manifest = Manifest.getInstance();
+        const isAvailableOffline = await manifest.isAvailableOffline();
```
PeteCoward, line 14: as mentioned before isAvailableOffline is irrelevant for Syncing

src/ts/Appelflap/SyncItemPublishing.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    private static async ManifestToPublishableItem(
+        manifest: Manifest
+    ): Promise<TPublishableItem> {
+        return {
+            cacheKey: manifest.cacheKey,
+            version: manifest.version,
+            isValid: manifest.isValid,
+            isAvailableOffline: await manifest.isAvailableOffline(),
+            isPublishable: manifest.isPublishable,
```
PeteCoward, line 45: manifest.isPublishable at the moment seems to be just a syncronous return of isValid. It should check the cache has a response for the manifest url like page.IsPublishable

src/ts/Appelflap/SyncItemListings.ts
md8n: Incorrect.  It is explicitly checked for (and needed) in `SyncStatus.SetSubscriptions`

src/ts/Appelflap/SyncItemPublishing.ts
md8n: It has been this way for a long time.  Making such a change is out of scope for this :pr:

Issue #622, opened 2021-04-19
On the course listing, courses are organized by category


pedro16v:
## Overview 
When creating courses, content managers can set their category. This should enable the course listing to be organized taking that information in consideration. 

## Conditions of Satisfaction
- Courses can be associated with a category
  - We can use the “tag” field as the category (already on Wagtail)
  - If a course is not associated with a category/tag, will be displayed on the app under “Other courses” category (for non-defined courses)
- Courses are grouped by category, on the course listing
  - For this release, ignore the circle avatars of categories on the top banner
  - Add one section per category with
    - Title
    - List of courses under that title
    - For now ignore the carousel navigation, courses will appear listed as they currently are on the course list
- Categories are alphabetically ordered on the all courses page

## Resources
* Reference design: [InVision](https://invis.io/MV10QV3Y9S2N)
* Possible solutions: [Google Slide](https://docs.google.com/presentation/d/1353MdkSb2xpF-NF10J59R-RkpC3-nNJH7MFUgs4MBlk/edit)

pedro16v:
Added `possible solutions` document. Current decision is to follow "Option 2" from [that document](https://docs.google.com/presentation/d/1353MdkSb2xpF-NF10J59R-RkpC3-nNJH7MFUgs4MBlk/edit#).

pedro16v:
@PeteCoward @megganeturner here's a few categories for you https://docs.google.com/spreadsheets/d/1HUfMfBq0SFN_7lZ1sKyTLaeqgppy4Y9TRAUEHLPnMD4/edit

megganeturner:
Hey @pedro16v just to clarify what we need to do here:

Front end wise, we want to get rid of the tag filters at the top of the page
<img src="https://user-images.githubusercontent.com/12974326/126250805-c4a73d50-955c-4730-ac77-8adf85daed7b.png" width="250">

and instead display each course under its category
<img src="https://user-images.githubusercontent.com/12974326/126250938-d95274fd-7ef2-448a-8d13-15495a083eca.png" width="350">

In terms of carousel navigation this is easy enough to do if we want it in to start with. I've done this on the homepage. Let me know.

In terms of backend functionality
- if courses can only belong to one category, we can modify the `tag` field
- however the `tag` field is in use for resources too - do we want resources to fit under one category too? And should resources & courses have the same categories available, or different? 

[...]

PR #885, opened 2021-07-28
622 course categories

megganeturner:
#622 

Merging into the homepage branch because the backend support for homepage & categories is also together in 1 PR https://github.com/catalpainternational/canoe-backend/pull/223#pullrequestreview-716610116

Implements course category support in the front end

Course page now lists courses organised by category:

You can click into a category to view all courses:

Categories are also listed on the homepage:

A fun bug this has surfaced is with the 'back' buttons on a course or category landing page. 
Our back arrows use the `parent.loc_hash` attribute on the `page` object. This worked previously because you could only get into a course from the AllCourses listing page. Now you can enter a course from: the homepage, the categories listing, and the category overview. There is no longer one parent to rule them all. (The same is true for categories).
I worked on an implementation of this back-button-inside-the-webpage functionality in my last job and it was tricky, and always buggy. So just flagging that as something that could take some time to do properly. I will work on that in another branch as this one is getting big!


src/riot/WagtailPages/AllCoursesList.riot.html
```diff
@@ -78,12 +117,30 @@ <h5 class="section-title">{ TRANSLATIONS.allCourses() }</h5>
...
+
+                get otherCoursesPageIsOpen() {
+                    const hashAtZero = getRoute().riotHash[0];
+                    if (hashAtZero && hashAtZero === 'otherCourses') {
+                        return true;
+                    }
+                    return false;
+                },
+
+                openOtherCourses() {
```
PeteCoward, line 139: I think we may want to display other courses by default IF there are no categories defined yet

PeteCoward (approved):
Happy to merge, but we should
- discuss Back/Up with team
- default to other courses view if no categories exist

we can do that in the `621-bero-homepage` PR

Issue #682, opened 2021-05-18
Content projection: teachers can cast media


pedro16v:
## Overview
Some important notes while designing a solution:
* Teachers need to easily know how and when they need to project lesson activity content in my classroom
* Teacher have an easy way to jump into projection mode and back into the “teach” flow
* Teachers using the app for the first time feel supported and confident to setup the projection of the activity content

## Conditions of satisfaction
* Media in a lesson can be casted using an external network device
* Media in a lesson shows a casting symbol, informing users that they can cast that content

## Resources
- Reference design: [InVision](https://invis.io/6V102V1VURS9#/443710126_Learning_Activities_-_Detail_-_Teach)

pedro16v:
Latest design from @andrewwapling https://invis.io/6V102V1VURS9#/443710126_Learning_Activities_-_Detail_-_Teach

andrewwapling:
Thanks for setting this up @pedro16v – let me know if you require any additional design support on this one.

pedro16v:
Some notes to help with future direction [here](https://docs.google.com/document/d/1YulOVeTxdOW8rfphGiBTqi56SsUo0koqa_5kJ3_bXJI/edit)

pedro16v:
In light of recent conversations, there are a few questions that **need answers.**

The [most updated design](https://projects.invisionapp.com/share/6V102V1VURS9#/screens/443710126_Learning_Activities_-_Detail_-_Teach) is a simple dismissible help text box informing the user about the benefits of casting media. You can also see a casting icon over the video.

The rationale is straightforward: show help next to the castable element.

<img width="320" alt="Screenshot 2021-07-16 at 10 59 06" src="https://user-images.githubusercontent.com/2813677/125930667-89121aec-6816-4e18-919e-c366e1308e24.png">

---

Now, I seem to recall @PeteCoward say that it's not easy to just make that help text box appear in different locations. It's easier to define a static position on the card. Am I correct @PeteCoward? 

**Thoughts**:
- Having to manually add casting help to cards that contain media (in the same way we add tips) seems odd. Unlike tips, the content here will always be the same. You'll see the same help text for every media element.
- If we can't position the help text box next to the media element, the alternative would be to find a place for it somewhere else in the interface. 
  - The very top? The very bottom? — we loose context
  - Some sort of roll up or banner? — would it appear every time we visit a card that contains media? Super annoying.
[...]

PR #890, opened 2021-07-29
682 + 683 tweaks

megganeturner:
Addresses Pedro's feedback in #682 + #683 

- Tweaks the "pass" scores for formative assessment ever so slightly
- Reorders video content in activity + limits the height of video previews


src/scss/_activity.scss
```diff
@@ -370,6 +370,21 @@ ActivityFeedback {
     }
 }
 
+VideoCard {
```
PeteCoward, line 373: @megganeturner does this affect ALL video cards even when not in activities?

megganeturner: Ahh I thought this selector was nested within `ActivityPage`! Will fix this now

PeteCoward (approved):

PR #899, opened 2021-07-30
Wait for the Injectables to stabilise before calling inject-caches

md8n:
# Description

This started life in a previous :pr:, but that wasn't possible to readily test thanks to other stability issues that impacted the testing.

This :pr: waits for two conditions to be true:
* The `quiescent` header returned by a GET to `injectables` to be 10 (seconds) or greater
* The count of injectable bundles to stabilise (at least 2 polls without change, or around ~6-12 seconds) before it calls to inject the bundled caches, which will normally be followed by a reboot of the webApp.

By doing this the webapp is highly unlikely to be rebooted multiple times when there are large numbers of bundles being shared with the device from other peers.

This is obviously less distressing for users (one reboot vs. several).  But it also reduces the risk of the app ending up in some indeterminate state because it reboots while some other bundles are coming in.


PeteCoward:
recommend lettgin @boerwastaken see this one

boerwastaken:
I think it's indeed a good idea to wait for stabilization in order to indeed "batch" the updates. :+1: 
Appelflap knows quite intimately how much time has passed since the subscriptions and the bundleset have been updated.
Wouldn't it be more convenient if Appelflap would pass this info along in a HTTP header at the `/appelflap/ingeblikt/ingeblikt/subscriptions/injectables` endpoint? Say

```
X-Appelflap-Settled: 12
```

which would indicate that the state hasn't changed in the last 12 seconds.
And then the frontend can decide to inject once this has reached some tunable value, no counting/diffing required. Counting is a bit course btw, state may be wildly in flux while the injectables count stays the same.

boerwastaken:
See catalpainternational/appelflap#117 for an implementation of the idea of letting Appelflap itself tell whether it's living in fast times or not.

boerwastaken (approved):

boerwastaken:
> The count of injectable bundles to stabilise (at least 2 polls without change, or around ~6-12 seconds) before it calls to inject the bundled caches, which will normally be followed by a reboot of the webApp.

The injectables cannot change without the quiescence being reset to 0. So it doesn't add anything.
Taking that out makes things simpler - just one thing to reason about; whether the quiescence is over $THRESHOLD or not. It makes it also easier to reason about for an external observer (who can witness the header, but can't witness the webapps internalized abstract of the appelflap state).

PR #904, opened 2021-08-02
Improve error handling when no backend

md8n:
# Description

When the app can't hit a backend endpoint a lot of errors are thrown and visible in the developer console.

In particular the frequent references to the `manifest` whenever routing results in lots of unhandled exceptions.  Although these do not break the app they do make debugging significantly harder.

Part of the problem is caused by not wrapping `await`ed calls to `PublishableItem.getResponse` at appropriate places in the call hierarchy.  Matters are also confused by throwing errors from methods that return a `Promise` when a `Promise.reject` should be used instead.

There was also a low impact bug, using `await` with a `.then(` handler on the same call

The changes in this :pr: significantly reduce the amount of uncaught/unhandled errors thrown whenever the app can't reach the backend, although it doesn't eliminate all of them.
These changes also seem to improve performance slightly, although there are still delays (long delays in the Android App) when first starting the app, or when trying to access a page that has never been downloaded due to having to wait for fetch timeouts.

This has been tested on both desktop and the mobile app.


PeteCoward:
> There was also a low impact bug, using await with a .then( handler on the same call

Can you point this out for me please?

PeteCoward:
> Matters are also confused by throwing errors from methods that return a Promise when a Promise.reject should be used instead

I can see that Promise.reject would avoid a logged console error, which is fine. What other differences does it make?

md8n:
> > There was also a low impact bug, using await with a .then( handler on the same call
> 
> Can you point this out for me please?

In `PublishableItem.ts` there is this line: `response = await this.getResponseFromCache().then(`

md8n:
> > Matters are also confused by throwing errors from methods that return a Promise when a Promise.reject should be used instead
> 
> I can see that Promise.reject would avoid a logged console error, which is fine. What other differences does it make?

Here's the answer to that question:
https://stackoverflow.com/questions/33445415/javascript-promises-reject-vs-throw

An extract from the above:
* `reject is a callback function that gets carried out after the promise is rejected, whereas throw cannot be used asynchronously`

That can be reworded as `throw` smashes its way out of the method. Whereas `reject` is more `async` friendly, and can be set up and then `return`ed when required.

PeteCoward:
> In `PublishableItem.ts` there is this line: `response = await this.getResponseFromCache().then(`

you mean [here](https://github.com/catalpainternational/canoe/pull/904/files#diff-c8deca1273ce1363ff0b8b72a94dc8f8cc4a483d194829cf4e824de9076f6461L127) ? Can you explain why that is a bug? What happens that should not happen?

PeteCoward (changes requested):
We can't introduce a json test in PublishableItem.getResponse. Only Page's and the Manifest are json, assets are not. Future PublishableItem's may not be either.

PeteCoward:
One day I will rename PublishableItem!!!!

md8n:
> > In `PublishableItem.ts` there is this line: `response = await this.getResponseFromCache().then(`
> 
> you mean [here](https://github.com/catalpainternational/canoe/pull/904/files#diff-c8deca1273ce1363ff0b8b72a94dc8f8cc4a483d194829cf4e824de9076f6461L127) ? Can you explain why that is a bug? What happens that should not happen?

Yep.

- What's going into the `then` is `cacheResponse` which is an actual `Response`.
- Within the `then`, the call to `getResponseFromNetwork` is not awaited so `networkResponse` is a `Promise<Response>`
- The return for that `then` is `return cacheResponse || networkResponse;` so it's either a `Response` or a `Promise<Response>` being returned, obviously they're not the same type.
- That outer `await` will handle that heterogenous type being returned, but it's not ideal.

md8n:
> We can't introduce a json test in PublishableItem.getResponse. Only Page's and the Manifest are json, assets are not. Future PublishableItem's may not be either.

Only `Manifest` and `Page` invoke `getResponse`, `Asset` does not invoke `getResponse`.
I've changed the test to only check for a status of `ok` (is true), which 'should' be OK with respect to the `Response.body` being of the correct type.

md8n:
> One day I will rename PublishableItem!!!!

Now in https://github.com/catalpainternational/canoe/pull/905 !!!!

PeteCoward (changes requested):
fails to compile now #905 is merged in

md8n:
> fails to compile now #905 is merged in

- It ain't 'compiling', it's a `prettier is ugly` 'error' that should be a 'warning', and someday I'll change that config setting.
- It was all tested before that typo of mine
- Now that you merged it by accident, I've retested everything (on and offline, because that's what this is about), it all still works.

PeteCoward:
> It ain't 'compiling', it's a prettier is ugly 'error' that should be a 'warning', and someday I'll change that config setting.

I have asked you to change that config a few times. 

If not a matter of formatting, it can be an error.
If it's just formatting and prettier/eslint tools can auto-fix it it can be an error.
If it's just formatting and requires manual fixing, it should not even be a warning I think.

PeteCoward (approved):

PR #910, opened 2021-08-04
remove AllCourses courses properly

PeteCoward:
One last ref to courses updated, and the original course property removed


PeteCoward:
oh dear, still no good

PeteCoward:
try now @megganeturner

megganeturner (approved):

PR #913, opened 2021-08-05
Add ky handle offline better

md8n:
# Description

The `ky` library is a wrapper around `fetch` that deals with a lot of fetch's undesirable behaviors, especially it's lack of a controllable timeout.

This is critical for the Android App as GeckoView seems to have a default timeout of 25 seconds when running on an Android device.  This timeout renders the app virtually unusable when used offline / no internet connectivity for several key tasks.

 `ky` will also give us better options to do things like test the accessibility of URLs with less overhead than `fetch`, this is likely to be necessary to get the Android Apps performance up when it's being used offline.


PeteCoward:
An alternative maybe a simple fetch wrapper like described [here](https://dmitripavlutin.com/timeout-fetch-request/)

md8n:
> An alternative maybe a simple fetch wrapper like described [here](https://dmitripavlutin.com/timeout-fetch-request/)

That's pretty much how `ky` does it.
I've seen a couple of other articles like this before, one of them was even a blatant cut-n-paste of the code within `ky`.
`ky` does give us a few other advantages, not the least of which is no dependencies and about 3KB minimised size.

One of `ky`'s specific advantages is that it has a default timeout of 10 seconds. Rather than 'leave it to the browser'.

Issue #912, opened 2021-08-04
Show progress for download content


pedro16v:
## Overview

In Bero, there's currently 3 different moments when users can make content offline:
* Course Lessons
* Learning Activities
* Resources

On a Course, users can select which Lessons to make offline and will see a loading spinner. Learning Activities and Resources have a similar control but no loading information is being displayed which can lead to confusion if the contents are large enough for the process to take more than just a few milliseconds.

## Conditions of satisfaction

- [ ] Animate "make offline" for Learning Activities and Resources.

## Resources

| Courses | Resources | Learning Activities |
| - | - | - |
![Screen Recording 2021-08-04 at 07 34 55 mov](https://user-images.githubusercontent.com/2813677/128133638-b397d918-d8c0-48b5-bdb2-96fbdc1d67f8.gif) | <img width="375" alt="Screenshot 2021-08-04 at 07 31 00" src="https://user-images.githubusercontent.com/2813677/128133643-fa5a3986-9a3e-4526-90e6-97290dd6c7a3.png"> | <img width="375" alt="Screenshot 2021-08-04 at 07 40 45" src="https://user-images.githubusercontent.com/2813677/128134673-e6a94864-537b-4d45-980c-53876c049fc2.png">

megganeturner:
Duplicate of https://github.com/catalpainternational/canoe/issues/908 😊

pedro16v:
Mine is nicer, has pictures in it! 😆 

I'll let Andy solve the design issue based on this information and feed it back to Peter's issue once it's done. Thanks Meggan!

andrewwapling:
Thanks for spinning this one up @pedro16v – please see proposed design solutions for each scenario:

**1. Teaching Activities** 
https://invis.io/52Y02L9YTRF#/454490471_Learning_Activities_-_Category_Listings
Interaction recommendations:
• Make offline label fades off
• CTA element transitions to loader element and smoothly moves down
• Light blue background tint fades in and download icon flashes smoothly in and out from 50% opacity to 100% opacity to reinforce activity
• Circle stroke fills with blue as item loads from 0% to 100%
• Loaders fades off and available offline label fades in

**2. Offline Courses** 
https://invis.io/52Y02L9YTRF#/446572478_Course_Detail_-_Offline_Modal
Interaction recommendations:
• Make offline label fades off
• CTA element transitions to loader element and smoothly moves down
• File size count transition from full file size in grey to file size counter in cool black
[...]

PR #934, opened 2021-08-09
908 dowload loading state

### LINKED ISSUE #908: Make Available Offline: Feedback
The make available offline buttons provide no feedback for users that they have been clicked, I'm not sure we have design for this, but typically we would "disable" the button (if it is a button) or hide and show some kind of message "retrieving..."

megganeturner:
Closes #908 & #912 (which is a duplicate of 908)

Provides visual feedback to users when resources/teaching materials are being downloaded:

https://user-images.githubusercontent.com/12974326/128655413-fcbc1862-e5af-43f7-a296-5c2f4242a459.mov


PeteCoward (commented):
Should the loading spinner scss belong in  `_course.scss` it seems like it might be better in a file that can be re-used in other places.

megganeturner:
> Should the loading spinner scss belong in `_course.scss` it seems like it might be better in a file that can be re-used in other places.

@PeteCoward have moved all of this out into its own component now :)

PR #937, opened 2021-08-09
Fix broken page types

md8n:
# Description

This fixes up the 'hidden' bugs related to unknown page types breaking the app, e.g. by causing the home pages not to show up, because we haven't allowed for the changes to page type names.  This does include some deprecated names for different page types, but they should all be flagged as deprecated to help identify this (see `CanoeEnums.ts` first).  This will give us some leeway in transitioning the front end separately from the back end (although they'll both be 'back together' soon anyway).

This leverages the Typescript definition for TPageTypes that cleans this all up, and ensures that we don't have random page types turning up that we know nothing about.

Also `routing.js` has been moved to `routing.ts` (with the associated types) to help keep this cleaner.

This is currently installed on PeP-staging


PeteCoward:
Please chat to me first before taking on work like this. The only need for it is because you are running incompitable frontend and backend, which we won't ever be doing in production. The simpler fix is to use compatible backends and frontends.

PeteCoward:
The way I want to deal with this kind of issue is...
1. consolidate backend and frontend into one repo
2. If a frontend is using a new manifest or content from a non-compatible older/newer version, throw an error and demand either a frontend code update or a backend content update

Doing this in the mean time just adds code we have to delete. I'm working on 1., would welcome thoughts on 2.

md8n:
> The way I want to deal with this kind of issue is...
> 
> 1. consolidate backend and frontend into one repo
> 2. If a frontend is using a new manifest or content from a non-compatible older/newer version, throw an error and demand either a frontend code update or a backend content update
> 
> Doing this in the mean time just adds code we have to delete. I'm working on 1., would welcome thoughts on 2.

This code effectively does 2 - if a manifest is received that has incompatible page types then it will break in an obvious way.

PeteCoward:
> If a frontend is using a new manifest or content from a non-compatible older/newer version, throw an error and demand either a frontend code update or a backend content update
> This code effectively does 2

No it doesn't and the way it goes about it is specific to this particular combination of circumstances which will never be repeated.

I'm thinking of a way we can bump API version, and specify that a frontend version A has backend content not A, then it shows a friendly error, and cleans it's caches. We really need to think through how to do this though. The frontend code update should maybe be offered to the user with a warning instead of mandatory as it is now.

md8n:
The PageTypes as such are still broken, and closing this issue doesn't make that go away.

PeteCoward:
> The PageTypes as such are still broken, and closing this issue doesn't make that go away.

I would fix them by removing that type https://github.com/catalpainternational/canoe/pull/949

src/riot/Components/BottomMenu.riot.html
```diff
@@ -2,11 +2,11 @@
     <footer if="{showBottomMenu()}">
         <div class="bottom-menu">
             <a href="#" if="{shouldShowLearningHome()}">
-                <span class="all-courses-icon { page.type === 'homepage' ? 'active' : ''}">
```
PeteCoward, line 5: I'd rather revert this whole file in this PR ( it will cause conflicts with the home page branch )

src/ts/Implementations/Manifest.ts
```diff
@@ -265,6 +266,7 @@ export class Manifest extends ContentItem implements StorableItem {
             case "teachingactivitypage":
                 pageInstance = new TeachingActivity(this, pageId, parent);
                 break;
+            case "tipspage": // Fall through to default
```
PeteCoward, line 269: as tips are essentially lesson pages, it may make more sense to return a lesson page. I assume that from the Sync page some component is calling this method for the tips page? It should never be routed to in normal fashion

src/ts/Types/CanoeEnums.ts
```diff
@@ -17,12 +17,21 @@ export type TAppelflapResult = "succeeded" | "not relevant" | "failed";
...
     | "homepage"
     | "resourcesroot"
     | "resourcearticle"
     | "coursepage"
-    | "lessonpage";
+    | "lessonpage"
+    | "teachingactivitieshomepage"
+    | "teachingactivitytopicpage"
+    | "teachingactivitypage"
+    | "tipspage";
```
PeteCoward, line 34: I'm really not happy about having to remember to make a change here everytime we add new pages or rename existing ones. I'm not convinced the value outweighs the drudge

PR #945, opened 2021-08-10
Init manifest properly

md8n:
# Description

Ensures that the manifest is always 'prepared' on entry into the App.

This is important when using a specific URL to enter the App, rather than the usual home page.

Needed for improved reliability of the Sync behaviour when offline.


PeteCoward (commented):
This should be done in routing. We don't want two places the manifest prepare is called from.

md8n:
> This should be done in routing. We don't want two places the manifest prepare is called from.

routing does not get invoked if the webapp is started with a URL that points to a Bero system page, and it is currently offline (relying on the cache).

That in turn breaks the start up process.

Routing in its current role does work well for the repeated attempts at refreshing the manifest.

PeteCoward:
> routing does not get invoked if the webapp is started with a URL that points to a Bero system page

what do you mean by a Bero system page? Give me an example please.

PeteCoward (commented):
As discussed. We need to know exactly what happens in the error case this is designed to avoid before committing code that fixes it.

PR #949, opened 2021-08-10
Remove TPageType

PeteCoward:
TPageType has not been kept up to date, and does not provide clear value.

It's been out of date for a good while with no ill effects, which makes me think it does nothing worthwhile. If it did we would have noticed it.


md8n (approved):

md8n:
I would prefer it if you add a comment to the switch block that uses this field to tell developers where to look in the backend code for what they match up to.

PR #956, opened 2021-08-12
Make resources available offline

PeteCoward:
This is a hotfix in a dead branch of Canoe to support offline resource capability in haroman
Published and previewable on https://beta.haroman.tl/

On navigating to a resource page, this will request the resource ( making it available offline ) then show a download link, which should be instant ( data is immediately available ).
If you visit a resource page, see the downlod link, then go offline and navigate away and back, you should be able to access the resource still.

- registers download path with the service worker
- on mount of a PDFCard
    - fetch the document url
    - report progress
    - build a data url
    - show the download link
    - apologise for any errors


pedro16v:
While testing this on `beta.haroman.tl` I got an error when loading.

Screenshot 2021-08-12 at 07 39 56

Screenshot 2021-08-12 at 07 40 08

PR #958, opened 2021-08-12
subscription range fixes

boerwastaken:
- fix some erroneous math, which resulted in, among other things, "tomorrow" being placed in the year 53586 .
- where versions were dynamically determined ("a day from now"), change for a value that only changes once a day (to cut down on spurious bumps; we really don't need to adjust the `p2p_version_max` every time the clock advances a millisecond)
- where a version was effectively "thousands of years into the future" but dynamically determined, swapped this out for a static value, again to cut down on bump noise. Though note that the comment states that the intention was to be 1 day ahead, not thousands of years. A day ahead for participating in the p2p network is not great for redistribution throughput. So in this case one bug's effect (1 day being too small) was alleviated by another bug (calculating a day as thousands of years); the effect of that double bug has been preserved (precaching aggresively). We can tone that down.

An outstanding problem is that these versions are calculated using the local clock, which is almost certainly asking for problems on wildtype devices.


boerwastaken:
So, the `injection_version_max` is determined using a local clocksource. If it's behind, it can mean that the manifest never gets updated as by the time the peer finally "wants" that version, everyone else has moved on and no one is sharing it anymore.

As noted before, part of the problem here is that the bero webapp plays fast and loose with app versions and the data format versions that that app version can handle. There is no formalized relation between the two. Load a new manifest into a webapp version that's "too different", and you'll have problems.
If it weren't for the absence of a rigorous formalization of that relation, we could safely set the manifest injection version to some very high number.

One easy thing that the webapp could do to make this better, is to use different caches (manifest-v1, manifest-v2, ....) for different manifest formats, and let the webapp read the manifest version that suits it. For the webapp-Appelflap interaction, such would mean that we can indeed safely set some very large value for the `injection_version_max` without running the risk of poisoning an old webapp with a too-new manifest. And then we can get rid of using the local clock.

md8n (approved):

PR #960, opened 2021-08-13
954 better sync page data

### LINKED ISSUE #954: Sync Page - prepare data for expected presentation to the user
# Description

We now have a better idea of what kinds of information we may need to show to the user so they can understand their 'sync status'.

Not all of this information may be shown to the user.  Some of it may be aggregated in different ways.

But we still want it available to the Sync page.

md8n:
Closes #954 

# Description

We needed different points of information on the sync page to help the user understand what's going on.  While at the same time trying to keep things simpler (in terms of presentation).

* To share content, the user needs to be authenticated (logged in) and their 'publishing certificate' needs to be signed.
When the user uses the Sync page it will refresh the list of items (bundles) that it can share, and the list of items needed.
* To receive content, the user needs Peers who have shared content already. In other words those Peers have also accessed their Sync page while logged in, with a valid publishing certificate, and while on the same WiFi hotspot.

The new information provided from `SyncStatus.ts` is the user's authentication status, certificate status, total number of pages (i.e. does NOT include the manifest in this count), and number of 'publishable' pages.

The Sync page itself has only been updated to show the 'publishable' pages as a percentage.

For the rest we're waiting on the design


PeteCoward (commented):
we should set up a time to get @megganeturner involved in implementing [the latest design](https://projects.invisionapp.com/share/MD113620ZJS6#/screens)

src/riot/Screens/Sync.riot.html
```diff
@@ -261,33 +256,42 @@ <h5 style="color:{statusListing.isPublishable ? 'green' : 'red'}; margin: 0.2em
             storeListener(previousStoreState, storeState) {
```
PeteCoward, line 256: I forgot to mention it before, but ALL bero riot tags update on ALL store commits because https://github.com/catalpainternational/canoe/blob/RC-3.5/src/riot/App.riot.html#L101

This means you can just allow the app to update this tag, and read direct from the state ( without having to duplicate redux state as tag state )

md8n: Let's do that as a separate :pr:

Issue #902, opened 2021-07-30
I can delete my own comments


pedro16v:
Follow up from #770

## Conditions of satisfaction

- [ ] I have the option to delete my own comments
- [ ] Deleting my comment updates the interface
  - [ ] A placeholder will inform other users about the deleted comment
  - [ ] A toast message will inform me about the success of the operation
- [ ] Content managers have access deleted comments (on Wagtail)

## Resources

* Explore the Invision [prototype](https://projects.invisionapp.com/share/4N10MWTLADZM#/screens/449199854)

A — Deleting a first level reply (note the confirmation toast)
B — Deleting a second level reply
C — Confirmation modal

| A | B | C |
| - | - | - |
<img width="320" alt="Screenshot 2021-08-13 at 12 56 47" src="https://user-images.githubusercontent.com/2813677/129354525-dcfd0409-31a0-473d-94d4-356c57ab1df5.png"> | <img width="320" alt="Screenshot 2021-08-13 at 13 00 30" src="https://user-images.githubusercontent.com/2813677/129354480-e8ec047f-9365-4465-9a78-0aee7129d59c.png"> | <img width="320" alt="Screenshot 2021-08-13 at 12 59 07" src="https://user-images.githubusercontent.com/2813677/129354495-7b345485-0c83-45e3-9143-b92640493001.png">

pedro16v:
@andrewwapling i believe the "A placeholder will inform other users about the deleted comment" requirement is a new thing. What do you think about this? The rationale here is to let users know that content was previously there which could explain oddness on the chain of comments.

Note for developers: if there's no design for a placeholder, we could start by simply hiding the deleted comment.

andrewwapling:
Hi @pedro16v thanks for setting this up. Please see proposed design for the deleted comment alert:

https://invis.io/JF11HW55SZRX

Kindly review and let me know if you're happy with this approach.

pedro16v:
I'm pulling this one to the backlog while we resolve [this ongoing discussion](https://projects.invisionapp.com/share/JF11HW55SZRX#/screens/454656278/comments/144634574) on InVision.

andrewwapling:
Hi @pedro16v pulling this one back in from Invision as promised please see below x3 scenarios for consideration:

1. Alert (show general alert in thread to all users letting them know a comment has been deleted) – https://invis.io/AZ11IPABU43C
[...]

PR #964, opened 2021-08-16
902 delete comments

megganeturner:
Building on Lee's work in #961
Requires backend: https://github.com/catalpainternational/canoe-backend/pull/250

Works towards #902 

Adds in the ability to delete a user's comments, and provide a visual cue for where comments have been deleted.

How it looks 

This is now working with the latest from the associated back end branch!

| ~Note for this to work~ |
| ----------------- |
| ~In order to see the 'deleted comment' component, we need to update the `get_discussion` definition in [`discussion/library.py`](https://github.com/catalpainternational/canoe-backend/blob/RC-3.5/discussion/library.py) so that it isn't filtering out comments where the `is_deleted` flag is `true`. Lee is aware of this and is going to rejig so that we're not sending through user & comment body information in the API for deleted comments (we don't display it and we don't need it)~ |

Behaviour demo:

https://user-images.githubusercontent.com/12974326/129524161-b80a60be-4916-4c0a-a52f-2d1e9506a8ee.mov

Introduction to tropical dairy has discussion questions to test - you will just need to log in and add some comments.


PeteCoward:
Does the "this comment has been deleted" show for all users? not just the deleter?

megganeturner:
> Does the "this comment has been deleted" show for all users? not just the deleter?

Yes, it shows for all users 😄

PeteCoward (approved):

PR #968, opened 2021-08-18
display courses in wagtail order

megganeturner:
From https://github.com/catalpainternational/canoe/discussions/957

Reinstates the regular wagtail ordering to course listings, so that in-progress courses aren't doubled up at the top of the list and in the carousel (keeps the completed courses at the end though)


pedro16v:
This will be easier to test once I see it with some JID content 😁

PeteCoward (approved):

PR #982, opened 2021-08-25
Identify browser as appelflap (when in appelflap)

md8n:
# Description

The `getBrowser` method in `PlatformDetection` was returning the browser user agent string for GeckoView hosted by Appelflap as `FireFox`, which is what it actually is.  However, we need it to return `Appelflap` so that server side behaviour for push notifications can work in a completely different way, letting Appelflap poll an endpoint every 15 minutes (or multiple thereof) for notifications.

This :pr: requires https://github.com/catalpainternational/django-push-notifications/pull/1 for it to work properly (mitigates downstream exceptions when other methods use the info from `getBrowser` when communicating with the server).

Also at the same time we needed to:
* properly identify `Edge` browsers (like what's done for `Opera`)
* get the browser version numbers properly
* eliminate support for IE, which the Bero build chain does not support anyway.


src/ts/PlatformDetection.ts
```diff
@@ -8,33 +8,39 @@ export function getPlatform(): TPlatform {
...
+ * and neither does Microsoft.
+ */
 export function getBrowser(): TBrowser {
     const ua = navigator.userAgent;
     let tem: any;
     let M: any[] =
-        ua.match(
-            /(opera|chrome|safari|firefox|msie|trident(?=\/))\/?\s*(\d+)/i
-        ) || [];
+        ua.match(/(chrome|edg|safari|firefox|opera)\/?\s*([\d.]+)/i) || [];
```
PeteCoward, line 20: edg?

PeteCoward (commented):
I'll believe you if you just confirm the `edg` usage, what `edg` starting strigns is that designed to match?

src/ts/PlatformDetection.ts
md8n: Edge browser

md8n:
> I'll believe you if you just confirm the `edg` usage, what `edg` starting strigns is that designed to match?

`Edg` is how the Edge browser ultimately identifies itself. It does start with `Chrome` first, then `Safari` (yep) and then finally `Edg`

md8n:
For example: `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/92.0.4515.159 Safari/537.36 Edg/92.0.902.78`

PR #986, opened 2021-08-25
Guest user details

md8n:
# Description

This ensures that, in general, there is always a `user` data structure, even when the user is not logged in.  This primarily benefits the `Profile` page.  It could benefit the Discussion pages as well, but first we'd have to allow for anonymous comments and replies in the model.

In all our usages of the `user` data structure we are never concerned with whether it is empty or not. `iaAuthenticated` and `isAuthenticationDetected` are much more useful for that purpose.

This clean up fixes a bug on the Profile page that occurred whenever the user structure did not have a `userId` (which is all the time now).

Also cleaned up:
* Google Analytics being fed the `userId` as `dimension1`, they know who you are anyway ...
* `getUsername`, `getCapitalizedUsername` and `getUserId` in `AuthenticationUtilities` have all been removed as they added no value


src/js/GoogleAnalytics.js
```diff
@@ -1,13 +1,11 @@
-import { getUserId } from "js/AuthenticationUtilities";
-
 const isGoogleAnalyticsAvailable = () => {
     return !!process.env.GA_TAG;
 };
 
 const getDimensions = () => {
```
PeteCoward, line 5: I think we can remove the whole method and it's calls, but would need to check I'm right

md8n: Oops - I missed that one (merge legacy I suspect)

md8n: It is trivial, but I left it in there as a kind of placeholder for where the other dimensions could be added (not that Google needs that because they already know everything about everyone)

Issue #965, opened 2021-08-16
Restructure Profile page


pedro16v:
## Overview
To complement the work described in #922, here's the ideal Profile page structure so that it scales across the different implementations. Pay attention to how the information is organised in these new screens.

**A** — Start here
**B** — Clicking on the cog icon
**C** — Clicking on "General settings"

| A | B | C |
| - | - | - |
![Profile@2x](https://user-images.githubusercontent.com/2813677/129565150-aa76bc2b-8091-41f4-8e1a-336815d9fcc8.png) | ![Settings@2x](https://user-images.githubusercontent.com/2813677/129565161-3348e331-c4d1-4feb-be70-eaa3db35f1bf.png) | ![Settings - General@2x](https://user-images.githubusercontent.com/2813677/129565170-14151147-fbbd-4118-b331-b075f744edec.png)

## Notes

* We're not implementing notifications yet
* On the `Learning Statistics` section, stats will show up depending on the type of content of each installation. In other words, if a Bero installation only has Courses and no Teaching topics, users will only see stats for Courses and Lessons.

## Resources

* Check the InVision prototype — https://projects.invisionapp.com/d/main?origin=v7#/console/21435394/454166189/preview?scrollOffset=1075

pedro16v:
Also @megganeturner please mind @andrewwapling 's [remarks on the original Profile issue](https://github.com/catalpainternational/canoe/issues/921#issuecomment-896576127)

pedro16v:
Accessing the Profile on JID staging is showing an error and causing the classic "stacked screens" bug

<img width="579" alt="Screenshot 2021-09-06 at 12 47 26" src="https://user-images.githubusercontent.com/2813677/132158062-0c4e18ae-5083-4374-af7e-e486638b2a5a.png">
<img width="375" alt="Screenshot 2021-09-06 at 12 47 37" src="https://user-images.githubusercontent.com/2813677/132158069-7108218e-46fd-40d6-99ae-0d8bb8518ad2.png">

Issue #921, opened 2021-08-06
I see my personal details on the Profile page


pedro16v:
## Overview
After logging in, users don't have a way of knowing which account they're logged into. Showing some basic user information would personalise their experience.

## Conditions of satisfaction
* On the profile screen, I can see:
  - [x] My name
  - [x] A placeholder avatar (the same generated for discussions)

## Resources
* Check on [InVision](https://projects.invisionapp.com/d/main?origin=v7#/console/20251865/429853514/preview?scrollOffset=2296)

<img width="320" alt="Screenshot 2021-08-06 at 13 44 47" src="https://user-images.githubusercontent.com/2813677/128512352-afe8cacd-ad78-4f2f-800a-e681ae00740a.png">

pedro16v:
Profile and settings wireframes from @Juliana-t for @andrewwapling 's reference: https://docs.google.com/presentation/d/1LetV2rQ1Xwr8hfDBVt2m-A3yeHoldp6CGqWUQG_wF1U/edit

andrewwapling:
@pedro16v @megganeturner  great to see the avatar element making it in! Really helps weight this section. A few small visual issues I can see immediately that I recommend refining include:

• The page title in the top nav bar is too big, please reduce and match design
• The Avatar circle is too small
• The Avatar initials are too large
• The name font weight looks a bit heavy
• The padding seems slightly off

Let's check these elements and try and match balance achieved in the design: https://invis.io/UQ11IZOWD8M2

PR #996, opened 2021-08-31
965 profile page restructure

megganeturner:
Addresses #965 & a comment here: https://github.com/catalpainternational/canoe/issues/921#issuecomment-896576127

Most of the work for the restructure was already done in the sync work, so this just adds a couple of extra things:
- Adds name/username into the profile settings (not editable, but might be in the future?)
- Since we're starting to see inputs more I've pulled the input out into its own component so that we're not repeating styles/logic - updated the login form to use these custom inputs
- Add the completion stats for teaching topics & activities to the profile page
- Fixes the logic to show an error on login when user/password combo is incorrect


src/riot/Screens/Profile.riot.html
```diff
@@ -42,26 +42,50 @@ <h5 class="section-title">{ TRANSLATIONS.completed() }</h5>
...
-                    </h2>
-                    <h5>{ TRANSLATIONS.lessonsCompleted() }</h5>
+                <div class="card stat">
+                    <div class="profile-card">
+                        <span class="stat-lessons-complete"></span>
+                        <h2>
+                            { allCourses.countCompleteLessons }
+                            <span class="stat-subheading">/ { allCourses.countLessons }</span>
+                        </h2>
+                        <h5>{ TRANSLATIONS.topicsCompleted() }</h5>
```
PeteCoward, line 63:
```suggestion
                        <h5>{ TRANSLATIONS.lessonsCompleted() }</h5>
```
??

PeteCoward (approved):

PR #1003, opened 2021-09-06
Get the APPLICATION_SERVER_KEY from file

md8n:
# Description

Every time that the VAPID keys get regenerated on the server (which includes the initial install) the corresponding `APPLICATION_SERVER_KEY` has to get copied into all of the relevant environment files and then a front-end redeployment performed.

Getting the correct form of the `APPLICATION_SERVER_KEY` has been greatly improved with the vapid_push management commands, but forgetting to copy the key to the correct place in the front-end environment files is still very likely to happen (i.e. all.the.time)

With this code change there's now an API endpoint in notifications, which calls through to vapid_push to get the application server key.

So reissuing the VAPID keys on the back-end is now a no brainer to handle on the front-end.

This :pr: is dependent on the back-end :pr: https://github.com/catalpainternational/canoe-backend/pull/268


PeteCoward (changes requested):
As discussed, try requesting it from the backend rather than configuring it

md8n:
> As discussed, try requesting it from the backend rather than configuring it

I'll go with this. In combination with suitable changes to catalpainternational/canoe#1007 this represents the best solution.

This :pr: will simply be more about removing all references to APPLICATION_SERVER_KEY coming via process.env

PeteCoward (approved):

PR #1006, opened 2021-09-07
972 discussion tooltips

### LINKED ISSUE #972: Onboarding: dismissible discussions helper tooltip
## Overview

Following up on the design work done in #849

User testing found that users not always realise they have a Discussion tab next to Lessons. Similarly, it wasn't obvious to them that they could save content for offline use. We decided to make these part of the onboarding process for new users.

We'll create onboarding mechanism for the following features:
* Make content offline
* Discussions

### Make content offline
- [x] When I visit a Course for the first time
  - [x] The helper would fade in and highlight the download UI control
  - [x] The user can then dismiss it permanently.
- [x] For Bero installations that allow guest users, this behaviour should apply in every new session (after clearing caches)

Screenshot 2021-08-19 at 11 50 47

### Discussions
- [x] When I visit a Course for the first time **AND** I scroll past a certain point, the onboarding mechanism should show.
  - [x] Certain point: when there's less than X pixels left to the `tab-titles` container (LESSONS | DISCUSSION)
  - [x] Reaching this point, the helper would fade in and highlight the Discussions UI control. 
  - [x] The user can then dismiss it permanently.
- [x] For Bero installations that allow guest users, this behaviour should apply in every new session (after clearing caches)

Screenshot 2021-08-19 at 11 20 11

## Resources

- Check the InVision prototype:
  - Make content offline https://projects.invisionapp.com/share/7Y11HPE4KQ63#/screens/446572479
  - Discussions https://projects.invisionapp.com/share/52Y02L9YTRF#/screens/455859748

megganeturner:
Closes #972 

Builds a reusable tooltip component and implements it on the course page for:
- course download icon
- discussion tab (if the course has a discussion tab)

For logged in users, utilises `localStorage` to store onboarding completions:
```js
{
    // a hashed username, so personal details aren't exposed to any particularly savvy users
    "679359813": {
        "courseDownloadTipCompletion": true,
        "courseDiscussionTipCompletion": true
    }
}
```

For logged out users, utilises `sessionStorage` to store completions:
```js
{
    "courseDiscussionTipCompletion": "true",
    "courseDownloadTipCompletion": "true"
}
```

Demo (note that the completions persist after page refresh):
https://user-images.githubusercontent.com/12974326/132299791-039e3972-e70b-4022-b243-6bfb598e4522.mov


PeteCoward (commented):
what the hell is an IntersectionObserver? What fun!
I suspect the saving may be overcomplicated though, and we should just treat all users the same whether logged in or not, just a first impression

src/riot/Components/ToolTip.riot.html
```diff
@@ -0,0 +1,137 @@
...
+            },
+
+            renderTooltip() {
+                if (this.state.currentElement.complete) return this.dismissTooltip();
+
+                const targetElement = document.getElementById(this.state.currentElement.targetId);
+
+                if (!targetElement) {
+                    this.update({ showToolTip: false });
+                    return;
```
PeteCoward, line 43: might be good to log something here so it can be spotted easily if we break stuff in future

PeteCoward:
let's chat tomorrow about it. I suspect if we just use localStorage and clear it on logout we'd be fine, would remove some code

megganeturner:
Fixed now @PeteCoward - simplified to just use localStorage for all users 😊

PeteCoward (approved):

PR #1007, opened 2021-09-08
Refresh subscription automagically if necessary

### LINKED ISSUE #999: Push Notifications: Correct subscription behaviour
# Description

How push subscriptions are handled currently is somewhat disjointed from what is supposed to happen:

- When a user subscribes for push notifications a `registrationID` is extracted from the subscription response and saved locally.  However, this is meaningless because the definition of whether a browser (not the user) is subscribed or not can only be determined with a call to the service worker's `pushManager.getSubscription()` method.
- A push notification subscription identifies a communication channel between the browser and the server (via the endpoint defined in the subscription). Even if the browser has a subscription, as defined by `pushManager.getSubscription()`, that subscription could be out-of-sync with the server, e.g. if the VAPID keys have been reissued on the server.  So the browser needs to verify that its subscription is up to date, and if it isn't it needs to go through the process of unsubscribing and resubscribing. These needs to happen whenever the app gets a connection to the server after previously losing its connection (e.g. the device lost internet connectivity). This needs to be done without user involvement as much as possible.

The first point above has already been fixed in the vapid_push feature branch.  The second point will greatly improve the 'reliability' of users receiving push notifications they have subscribed to.

md8n:
Closes #999 

# Description

The browser holds the authoritative information as to whether the user is subscribed or not.  However, this information may still be out of date with respect to the subscription info held on the server, or the APPLICATION_SERVER_KEY as provided by the server (e.g. if the VAPID certs on the server get updated this setting will change).

Also the user may have changed their permissions to revoke the permissions related to push notifications, so there's a need to revoke (unsubscribe) if the user has done this.

This :pr: checks `isSubscribed` on startup.  `isSubscribed` has been updated to verify the permission revocation, and if the user is subscribed it will also check for a matching subscription on the server and refresh the subscription if that's not found.


PeteCoward (approved):

src/app.js
```diff
@@ -33,6 +35,9 @@ initialiseIdentity()  // Am I logged in
     .then(() => {
         // read action from api and idb
         return initialiseUserActions();
-    }).then(() => {
+    }).then(async () => {
+        // If the user is subscribed this will also refresh their subscription
+        await Notifications.IsSubscribed();
```
PeteCoward, line 40: I'm not sure this should hold up routing, why not let it run in the background?

src/ts/Notifications.ts
```diff
@@ -0,0 +1,265 @@
+import Cookies from "js-cookie";
+
+import { TVapidPushSubscription } from "./Types/VapidPushTypes";
+
+import Logger from "./Logger";
+import { getBrowser } from "./PlatformDetection";
+
+import { BACKEND_BASE_URL, CSRF_COOKIE_NAME } from "js/urls";
+import { alertIfRequestWasMadeOffline } from "js/Errors";
+import { logUnsubscribedFromNotifications } from "js/GoogleAnalytics";
```
PeteCoward, line 10: lets discuss this. I'd rather not be sending this to GA. If someone unsubscribes maybe we can store it in our new vapid_push models?

src/app.js
md8n: I was wondering the same thing myself ...

src/ts/Notifications.ts
md8n: Happy to discuss this. However, it is not part of this :pr: - it's pre-existing code.  We need to raise this as a separate issue.

src/app.js
md8n: Changed to not `await`

PeteCoward (approved):

PR #1012, opened 2021-09-10
pep 42: teaching activity typography clean up

megganeturner:
Closes https://github.com/catalpainternational/pep/issues/42

fixes for typography - making it more consistent & correcting the visual hierarchy across lessons & teaching activities

Before:

After:


PeteCoward (approved):
I'll get it in, looks harmless and best test is to get it on staging

Issue #1021, opened 2021-09-16
Push Notifications: Extend basic support in Service Worker


md8n:
# Description

We've never really had push notifications properly supported for anything in Bero.  Which means that we only have a very basic set of functionality in the service worker, and what we have isn't really adequate to do anything with.

There's a need to add some basic hooks for the service worker to be able to invoke the kinds of behaviour we want.

PR #1026, opened 2021-09-20
Push Notification: service worker improvements

md8n:
Partial support for #1021 

# Description

The service worker code was never properly built out to process push notifications when received.

This adds improved type support for the client side notification handling, as well as identifying where many hooks should go for future development (e.g. adding support for discussion response notifications).  It also removes concepts that originally came through from suggestions for django_push_notifications that compromise how push notifications from the server should be processed and how OS/Browser notifications should be generated client side.

This :pr: depends upon the backend :pr: https://github.com/catalpainternational/canoe-backend/pull/284

**IMPORTANT** - this is combined in :pr: https://github.com/catalpainternational/canoe-backend/pull/288 - review and merge that one instead


PeteCoward:
closing as now combined in canoe-backend

